PR #2, opened 2020-03-24
SOC-6322 : Add template params when posting an activity in UserRestResourcesV1

ahamdi:
Actuallay when creating an activity with the Rest service UserRestResourcesV1, all template params are not loaded and saved with the activity.
The actual fix will add template params when posting an activity in UserRestResourcesV1


ahmadmido (approved):
LGTM

PR #3, opened 2020-03-24
Create Getting Started Portlet

SaraBoutej:
Create the Getting started Portlet.
This portlet will be displayed in the stream page

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/gettingStarted/service/GettingStartedService.java
+++ b/component/service/src/main/java/org/exoplatform/social/gettingStarted/service/GettingStartedService.java
@@ -29,31 +29,28 @@ public class GettingStartedService {
 
   private List<GettingStartedStep> gettingStartedSteps;
 
-  public List<GettingStartedStep> getUserSteps(String lang) {
+  public List<GettingStartedStep> getUserSteps() {
 
     String userId = ConversationState.getCurrent().getIdentity().getUserId();
     List<GettingStartedStep> gettingStartedSteps = new ArrayList<>();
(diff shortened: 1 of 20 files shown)


component/service/src/main/java/org/exoplatform/social/gettingStarted/service/GettingStartedService.java
```diff
@@ -0,0 +1,144 @@
...
+  public List<GettingStartedStep> getUserSteps(String lang) {
+
+    String userId = ConversationState.getCurrent().getIdentity().getUserId();
+    List<GettingStartedStep> gettingStartedSteps = new ArrayList<>();
+    ResourceBundleService resourceBundleService = CommonsUtils.getService(ResourceBundleService.class);
+    ResourceBundle resourceBundle = resourceBundleService.getResourceBundle("locale.portlet.social.GettingStartedPortlet",
+                                                                            new Locale(lang));
+
+    GettingStartedStep stepAvatar = new GettingStartedStep();
+    stepAvatar.setName(resourceBundle.getString("social.portlet.gettingStarted.avatar"));
```
boubaker: I think that it's better to let I18N in portlet only. Here, we can return a keyword ('avatar' by example)
And in portlet, you use:
``
$t(`social.portlet.gettingStarted.${step.name}`)
``
In addition, in I18N file, please use 'locale.portlet.GettingStartedPortlet' (without 'social'). Same thing for keywords (`gettingStarted.step.avatar` instead of `social.portlet.gettingStarted.avatar`)

SaraBoutej: Done!

boubaker (approved):

PR #12, opened 2020-03-25
Fix Notification link computing

boubaker:
When the default site is turned into 'dw', the notifications on 'intranet' site (with new notification portlet) doesn't work anymore.
Thus, this fixes the problem by making the link computing more generic.


azayati (approved):
LGTM

PR #13, opened 2020-03-26
Fix searchbar position in tablet and mobile

azayati:


azayati (commented):
@boubaker To be reviewed

search/portlet/src/main/webapp/js/common/quicksearch.js
```diff
@@ -459,18 +459,17 @@ window.initQuickSearch = function initQuickSearch(portletId,seeAllMsg, noResultM
       $(this).toggleClass('uiIconCloseSearchBox')
              .parents('#ToolBarSearch').find('input[type="text"]').toggleClass("showInputSearch").removeClass('loadding').focus()
              .parents('#ToolBarSearch').find('.uiQuickSearchResult').hide()
-             .parents('#PlatformAdminToolbarContainer').toggleClass('activeInputSearch')
+             .parents('.SearchPortletTDContainer').toggleClass('activeInputSearch')
```
boubaker: IMO, the "active" style of search box shouldn't be applied switch parent container class name, it should be managed in portlet itself

azayati: In fact, there is a class applied in tablet on .SearchPortletTDContainer when the .activeInputSearch is present in the DOM **.activeInputSearch .SearchPortletTDContainer**
If the class activeInputSearch will be under the parent container on the portlet div, how to apply this class in that case ?

boubaker: No JS should be used IMO and no Parent has to be modified or receive a class neither. The application (quicksearch here) should set position of its element, independently from parent layout containers. So the CSS of Quicksearch application shouldn't reference topbar of DW and Toolbar of intranet, and TopBard of meeds and toolbar of customers sites ... to get well positioned and styled.

So let's see again what's expected:
* In desktop & mobile : we want to have a text box that is displayed when clicking on search icon
* When the searchbox is opened, the 'search' icon is modified to 'Close' icon
* On Mobile & Tablet only, the searchbox get 100vw (full width) of the screen

Is there something else to implement ?

azayati: yes that is it normally

azayati:
@boubaker to be validated

boubaker (approved):

PR #22, opened 2020-03-31
feature/composer: Add loader over attachments icon when uploading and prevent from posting

aelbenney:


webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
```diff
@@ -136,6 +155,13 @@ export default {
     },
     executeAction(action) {
       action.onExecute();
+    },
+    setUploadingCount: function(action) {
+      if (action === 'add') {
+        this.uploadingCount++;
+      } else if (action === 'remove') {
+        this.uploadingCount--;
```
thomasdelhomenie, line 163: We should check if the count is already at 0 or not before decreasing it IMO, it would avoid having negative count even if something went wrong.

thomasdelhomenie (changes requested):

webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
aelbenney: resolved here https://github.com/Meeds-io/social/pull/23/commits/a5996bfd2a130cd0e4c2ad9f0c19c623e35fb15f

PR #28, opened 2020-04-02
Feature/activity-composer: Refactor extension

aelbenney:


webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
```diff
@@ -61,16 +91,27 @@ export default {
...
+      actionsData: [],
+      actionsEvents: [],
+      actionComp: [],
     };
   },
   computed: {
     postDisabled: function() {
       const pureText = this.message ? this.message.replace(/<[^>]*>/g, '').replace(/&nbsp;/g, '').trim() : '';
-      return pureText.length === 0 || pureText.length > this.MESSAGE_MAX_LENGTH;
+      return pureText.length === 0 && this.actionsData['file'].value.length === 0 || pureText.length > this.MESSAGE_MAX_LENGTH || this.uploading;
```
thomasdelhomenie: This links the "file" plugin to the activity composer. It should not. The activity composer should be agnostic of the plugins.
We should find a way to set the `attachments` variable to `this.actionsData['file'].value` when files are selected. Then we can only work with `this.attachments`. WDYT ?

thomasdelhomenie (changes requested):

webapp/portlet/src/main/webapp/activity-composer-app/extension.js
```diff
@@ -9,9 +9,9 @@ export function getActivityComposerExtensions() {
   return activityComposerActions;
 }
 
-export function executeExtensionAction(extension) {
+export function executeExtensionAction(extension, component) {
   if(extension.hasOwnProperty('onExecute') && isFunction(extension.onExecute)) {
-    extension.onExecute.call();
+    extension.onExecute(component[0]);
```
thomasdelhomenie, line 14: Why `component[0]` and not `component` ?

webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
```diff
@@ -82,6 +123,15 @@ export default {
   },
   created() {
     this.activityComposerActions = getActivityComposerExtensions();
+    this.activityComposerActions.forEach(action => {
+      if (action.component) {
+        this.actionsData[action.key] = action.component.model;
+        this.actionsEvents[action.key] = action.component.events;
+      }
+    });
+    this.$watch('actionsData.file.value', function () {
```
thomasdelhomenie: `file` is still "hardcoded" here, can we get rid of it ?

thomasdelhomenie (changes requested):

webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
aelbenney: So what we will put? what we will be watching to put in the attachments array?

thomasdelhomenie: We could add a new event in the ECMS plugin to call a method addAttachment (to be defined in the ExoActivityComposer component):
```
        {
          'event': 'uploadingFileFinished',
          'listener': 'addAttachment',
          'listenerParam': 'attachment'
        }
```
so we can remove this watch. WDYT ?

thomasdelhomenie (changes requested):

webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
aelbenney: I suggest making a watch to the attachments array in the attachments component because we need the exact number to display and a loader to be displayed. Otherwise, we will be creating three events(finishUpload, remove, addExisting) for handling the attachments from the new composer.

PR #36, opened 2020-04-06
Add suggestions peoples and spaces portlet

SaraBoutej:


webapp/vue-portlet/src/main/webapp/suggestions-people-space/components/ExoSuggestionsPeopleAndSpace.vue
```diff
@@ -0,0 +1,147 @@
...
+        this.currentUser = data.username;
+      });
+    },
+    initSpaceSuggestionsList() {
+      suggestionsServices.getSpaceSuggestions().then(data => {
+        this.spacesSuggestionsList = data.items;
+      });
+    },
+    connectionRequest(item) {
+      suggestionsServices.sendConnectionRequest(item.suggestionId);
```
boubaker: This is a `Promise` so next line should be in `then`, isn't it ? (You may want to delete suggestion line independently from response status code)
(Same for other methods)

SaraBoutej: Okey done!

boubaker (approved):

PR #37, opened 2020-04-07
Moved IdentitySearchRESTService from ECMS

VladMartyshevskyi:
Moved IdentitySearchRESTService from ECMS.
This PR should be merged along with https://github.com/exodev/ecms/pull/568 after merging https://github.com/exodev/ecms/pull/551

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/service/rest/IdentitySearchRESTService.java
+++ b/component/service/src/main/java/org/exoplatform/social/service/rest/IdentitySearchRESTService.java
@@ -1,0 +1,236 @@
+/*
+ * Copyright (C) 2003-2020 eXo Platform SAS.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Affero General Public License
+ * as published by the Free Software Foundation; either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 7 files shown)


component/service/src/main/java/org/exoplatform/social/service/rest/IdentitySearchRESTService.java
```diff
@@ -0,0 +1,236 @@
...
+import org.exoplatform.social.service.rest.editors.ErrorMessage;
+import org.exoplatform.social.service.rest.editors.HypermediaLink;
+import org.exoplatform.social.service.rest.editors.HypermediaSupport;
+import org.exoplatform.social.service.rest.editors.IdentityData;
+
+/**
+ * The Class IdentitySearchRESTService.
+ */
+@Path("/identity/search")
+public class IdentitySearchRESTService implements ResourceContainer {
```
boubaker, line 59: Why adding a new endpoint for users ?
Isn't sufficient the following endpoints:
* https://github.com/Meeds-io/social/blob/develop/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
* https://github.com/Meeds-io/social/blob/develop/component/service/src/main/java/org/exoplatform/social/rest/impl/userrelationship/UsersRelationshipsRestResourcesV1.java
* https://github.com/Meeds-io/social/blob/develop/component/service/src/main/java/org/exoplatform/social/rest/impl/relationship/RelationshipsRestResourcesV1.java

If it's not, I suggest to enrich those existing.

What is the main idea behing the following endpoints ? is it for suggestion ? we have other endpoints for this too. I invite you to know more about it : 
* https://github.com/Meeds-io/social/blob/develop/component/service/src/main/java/org/exoplatform/social/rest/invitation/PeopleAndSpacesRestService.java
* https://github.com/Meeds-io/social/blob/develop/component/service/src/main/java/org/exoplatform/social/rest/suggest/PeopleRestServices.java

PR #55, opened 2020-04-15
New composer : Hide the old composer by default

thomasdelhomenie:
This improvement hides the old composer by default.
An action link allows to show it.

The style and script are all done directly in the gtmpl in order to make it easier to remove them soon.
This improvement is a temporary step since the old composer will very probably be completely removed soon.


aelbenney (dismissed):

thomasdelhomenie:
Replaced by https://github.com/Meeds-io/social/pull/64

PR #56, opened 2020-04-16
Feature people page merge

boubaker:


component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/ConnectionDAOImpl.java
```diff
@@ -242,6 +242,66 @@ public int getConnectionsByFilterCount(Identity identity, ProfileFilter profileF
     return query.getResultList();
   }
 
+  @Override
+  public int getConnectionsInCommonCount(String id1, String id2) {
```
ahamdi, line 246: Could we add a check if user with id1 or user with id2 have no connection, then we return 0 directly,  WDYT ?

boubaker: In fact, the query is long but it's performant. In addition, this information will be cached, see :
https://github.com/Meeds-io/social/blob/108361965b9f50593493e7baa76d62762e7f776c/component/core/src/main/java/org/exoplatform/social/core/storage/cache/CachedRelationshipStorage.java#L839-L863

IMO, in most cases, users will have connections, and to avoid making two request, I would prefer make this one once for all and then cache the information.

( BTW, we will see its impact, in UI, it's used when the user displays his invitations or pending users only )

SaraBoutej (dismissed):

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/ConnectionDAOImpl.java
ahamdi: ok

ahamdi (dismissed):

ahamdi (approved):

PR #65, opened 2020-04-22
Feature/activity-composer-vue: Move exo modal to common components in social

aelbenney:
We moved the exo modal component here in social to make it reusable in other modules.

### DIFF
--- a/webapp/vue-portlet/src/main/webapp/common/components/ExoModal.vue
+++ b/webapp/vue-portlet/src/main/webapp/common/components/ExoModal.vue
@@ -1,0 +1,58 @@
+<template>
+  <div v-show="show" ref="modal" class="modal-mask uiPopupWrapper">
+    <div class="uiPopup modal-content">
+      <div class="popupHeader">
+        <a class="uiIconClose pull-right" @click="$emit('close')"></a>
+        <span class="popupTitle">{{ title }}</span>
+      </div>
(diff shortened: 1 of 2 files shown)


webapp/vue-portlet/src/main/webapp/common/components/ExoModal.vue
```diff
@@ -0,0 +1,58 @@
...
+          evt = evt || window.event;
+          if (evt.keyCode === escapeKeyCode) {
+            this.$emit('close');
+          }
+        };
+
+        //this is workaround to fix chrome bug
+        //popup is blurring when using translate() and the size is odd
+        Vue.nextTick(function() {
+          const width = this.$el.offsetWidth;
```
boubaker: IMO, it's better to use Vuetify that manages perfectly all of those events, you can take a look at : 
https://github.com/Meeds-io/social/blob/ee43b9d23779803bfcc2900c69dbc70432a1fe01/webapp/vue-portlet/src/main/webapp/common/components/ExoConfirmDialog.vue

that is a modal for 'confirmation'

aelbenney: it's done!

thomasdelhomenie (dismissed):

boubaker (dismissed):

thomasdelhomenie (approved):

PR #68, opened 2020-04-24
US-ShortMessage-02 : Display 'Switch to an article' when the number of caracters exceeds 1300

Rdidmohamedelhadi:


webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
```diff
@@ -87,7 +93,7 @@ export default {
   },
   data() {
     return {
-      MESSAGE_MAX_LENGTH: 2000,
+      MESSAGE_MAX_LENGTH: 1300,
```
thomasdelhomenie, line 96: Max length is still 2000. 1300 is for displaying the message "Switch to an article" and change the characters count in red.

webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
```diff
@@ -97,7 +103,10 @@ export default {
       attachments: [],
       percent: 100,
       actionsData: [],
-      actionsEvents: []
+      actionsEvents: [],
+      enableSwitchAnArticle: false,
+      showSwitchAnArticle: false,
```
thomasdelhomenie: There should not be any reference to News addon in social IMO. We should allow to define this behavior by using a plugin from News.

thomasdelhomenie (changes requested):

webapp/portlet/src/main/webapp/activity-composer-app/extension.js
```diff
@@ -9,6 +10,16 @@ export function getActivityComposerExtensions() {
   return activityComposerActions;
 }
 
+export function getActivityComposerSwitchToAnArticle(messageLength) {
```
thomasdelhomenie: The term 'switchToAnArticle' should be removed everywhere and replaced by something related to hint action. For example here it could be 'getActivityComposerHintActionExtensions'.
And we should probably rename the method `getActivityComposerExtensions` to `getActivityComposerActionExtensions`.

thomasdelhomenie (changes requested):

webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
```diff
@@ -16,7 +16,13 @@
         <div class="content">
           <exo-activity-rich-editor ref="richEditor" v-model="message" :max-length="MESSAGE_MAX_LENGTH" :placeholder="$t('activity.composer.placeholder').replace('{0}', MESSAGE_MAX_LENGTH)"></exo-activity-rich-editor>
           <div class="composerButtons">
-            <button :disabled="postDisabled" type="button" class="btn btn-primary ignore-vuetify-classes" @click="postMessage()">{{ $t('activity.composer.post') }}</button>
+            <div v-for="action in activityComposerHintAction" :key="action.key" class="action">
+              <i class="fas fa-pencil-alt fa-sm	colorIcon" @click="activityComposerHintAction[0].onExecute()"></i>
+              <a class="message" href="javascript:void(0)" @click="activityComposerHintAction[0].onExecute()" >{{ getLabel(action.labelKey) }} </a>
+            </div>
+            <div v-if="activityComposerHintAction === null || messageLength < 1300" class="emptyMessage">
```
thomasdelhomenie:
```suggestion
            <div v-if="activityComposerHintAction === null || messageLength < MESSAGE_MAX_LENGTH" class="emptyMessage">
```

webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
```diff
@@ -121,6 +132,10 @@ export default {
     }
   },
   watch: {
+    message(){
+      const pureText = this.message ? this.message.replace(/<[^>]*>/g, '').replace(/&nbsp;/g, '').trim() : '';
+      this.activityComposerHintAction = getActivityComposerHintActionExtensions(pureText.length);
```
thomasdelhomenie: Why do we need to call this method each time the message is updated ? We could instead retrieve the list of extensions and use it only when the message is too long. WDYT ?

webapp/portlet/src/main/webapp/activity-composer-app/extension.js
```diff
@@ -9,6 +10,19 @@ export function getActivityComposerExtensions() {
...
+  const MESSAGE_MAX_LENGTH = 1300;
+  if (messageLength > MESSAGE_MAX_LENGTH) {
+    activityComposerHintAction = getExtensionsByType('activity-composer-hint-action');
+    if (activityComposerHintAction.length === 0) {
+      activityComposerHintAction = null;
+    }
+  } else {
+    activityComposerHintAction = null;
+  }
+  return activityComposerHintAction;
```
thomasdelhomenie, line 23: Since we can only use 1 plugin at a time for this type of plugin (since there is room only for 1 link in the UI), we should sort it by rank, so the plugin with the 1st rank will be used.

thomasdelhomenie (changes requested):

PR #69, opened 2020-04-28
PLF-8674: Disabled users' display issues

lbengayed:


component/service/pom.xml
```diff
@@ -295,6 +295,10 @@
         <artifactId>jersey-client</artifactId>
         <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>org.exoplatform.social</groupId>
+          <artifactId>social-component-webui</artifactId>
```
boubaker: The Rest Services shouldn't depend on Webui artifact

component/service/src/main/java/org/exoplatform/social/service/rest/PeopleRestService.java
```diff
@@ -499,7 +499,10 @@ public Response suggestUsernames(@Context UriInfo uriInfo,
   private LinkedHashSet<UserInfo> addSpaceMembers (String spaceURL, ProfileFilter identityFilter, LinkedHashSet<UserInfo> userInfos, String currentUser) {
     String[] spaceMembers = getSpaceService().getSpaceByUrl(spaceURL).getMembers();
     for (String spaceMember : spaceMembers) {
-      userInfos = addUsernameToInfosList(spaceMember, identityFilter,userInfos, currentUser, true);
+      Identity identity = LinkProvider.getIdentityManager().getOrCreateIdentity(OrganizationIdentityProvider.NAME, spaceMember, false);
```
ahamdi: Why not 

```suggestion
      Identity identity = getIdentityManager().getOrCreateIdentity(OrganizationIdentityProvider.NAME, spaceMember, false);

```

ahamdi (changes requested):
Please update Unit tests in PeopleRestServiceTest

ahamdi (approved):
LGTM

component/service/pom.xml
ahamdi: @boubaker  This was fixed, could you re-validate please :)

PR #75, opened 2020-04-29
US-ShortMessage-02 : Display "Switch to an article" when the number of caracters exceeds 1300

Rdidmohamedelhadi:


webapp/portlet/src/main/webapp/activity-composer-app/extension.js
```diff
@@ -1,6 +1,7 @@
 let activityComposerActions = null;
-
-export function getActivityComposerExtensions() {
+let activityComposerHintAction = [];
```
thomasdelhomenie: Why do we need an array for 1 action ?

webapp/portlet/src/main/webapp/activity-composer-app/extension.js
```diff
@@ -9,6 +10,19 @@ export function getActivityComposerExtensions() {
...
+export function getActivityComposerHintActionExtensions() {
+  activityComposerHintListActions = getExtensionsByType('activity-composer-hint-action');
+  if (activityComposerHintListActions.length > 0 && activityComposerHintAction.length === 0) {
+    activityComposerHintAction.push(activityComposerHintListActions[0]);
+  }
+  for (let i = 1; i < activityComposerHintListActions.length; i++) {
+    if (activityComposerHintAction.rank > activityComposerHintListActions[i].rank) {
+      activityComposerHintAction = activityComposerHintListActions[i];
+    }
+  }
```
thomasdelhomenie, line 21: Can you explain the logic please ?

thomasdelhomenie (changes requested):

webapp/portlet/src/main/webapp/activity-composer-app/extension.js
Rdidmohamedelhadi: To ensure that the action take the least rankType a message

Rdidmohamedelhadi: To ensure that the action take the least rank

thomasdelhomenie: I see you simplified the logic, but it does not take the rank in account anymore, right ?

thomasdelhomenie (dismissed):

thomasdelhomenie (approved):

PR #88, opened 2020-05-07
US-InsertVideo-02 : Video links preview after posting

Rdidmohamedelhadi:


component/service/src/main/java/org/exoplatform/social/processor/EmbedProcessor.java
```diff
@@ -27,13 +29,15 @@
   private static final String TITLE = "title";
   private static final String COMMENT = "comment";
   private static final String LINK_ACTIVITY = "LINK_ACTIVITY";
+  private static final String HTML_PARAM = "html";
```
thomasdelhomenie, line 32:
```suggestion
  private static final String MEDIA_HTML = "media_html";
```
I would rename it like that to make it more meaningful. WDYT ?

thomasdelhomenie (changes requested):

thomasdelhomenie (dismissed):

rdenarie (approved):
LGTM

PR #113, opened 2020-05-29
Feature/activity-composer : Update embed processor to display link preview with files

Rdidmohamedelhadi:
In order to display link preview when files exits we should not update the activity type. However, we need to add the existing file params to the link template params to pass them to the activity builder.


rdenarie (changes requested):
Please remove ITOP commit to merge

rdenarie (approved):

PR #115, opened 2020-06-03
EE-6.0: Enable/Disable main stream composer

Rdidmohamedelhadi:


plamarque (changes requested):
property name does not clearly reflect that it is the composer on stream page that is configured.

Also, it should be false by default on eXo and true by default on Meeds.  This PR only deals with the case of eXo.

Finally, I think the property should be present and documented with a comment inside the build-in exo.properties.

In current state this PR essentially creates an invisible property. It should be exposed.

Rdidmohamedelhadi:
> property name does not clearly reflect that it is the composer on stream page that is configured.
> 
> Also, it should be false by default on eXo and true by default on Meeds. This PR only deals with the case of eXo.
> 
> Finally, I think the property should be present and documented with a comment inside the build-in exo.properties.
> 
> In current state this PR essentially creates an invisible property. It should be exposed.

ok !! well noted

extension/war/src/main/webapp/groovy/social/webui/composer/UIComposer.gtmpl
```diff
@@ -25,9 +25,14 @@
   import org.exoplatform.container.ExoContainerContext;
   import org.exoplatform.commons.api.settings.ExoFeatureService;
   import org.exoplatform.social.webui.composer.UIActivityComposerContainer;
+  import org.exoplatform.commons.utils.PropertyManager;
 
   ExoFeatureService featureService = ExoContainerContext.getCurrentContainer().getComponentInstanceOfType(ExoFeatureService.class);
-  if(featureService.isActiveFeature("new-composer")) {
+  def isEnable = PropertyManager.getProperty("exo.composer.enable");
```
azayati:
```suggestion
  def isMainStreamComposerEnabled= PropertyManager.getProperty("exo.config.mainStreamComposer.enabled");
```

mzorai:
> property name does not clearly reflect that it is the composer on stream page that is configured.
> 

I proposed this property name  exo.social.composer.stream.hidden
true: to disable displaying the composer in the stream
false: to enable displaying the composer in the stream

extension/war/src/main/webapp/groovy/social/webui/composer/UIComposer.gtmpl
```diff
@@ -25,9 +25,14 @@
...
+  if( isEnable == null ){
```
azayati: To be deleted

extension/war/src/main/webapp/groovy/social/webui/composer/UIComposer.gtmpl
```diff
@@ -25,9 +25,14 @@
...
+      isEnable = false;
+  }
+  if(featureService.isActiveFeature("new-composer") && isEnable) {
```
azayati:
```suggestion
  if(featureService.isActiveFeature("new-composer")) {
  if (isMainStreamComposerEnabled) {
```

azayati (changes requested):
See review

azayati:
> > property name does not clearly reflect that it is the composer on stream page that is configured.
> 
> I proposed this property name exo.social.composer.stream.hidden
> true: to disable displaying the composer in the stream
> false: to enable displaying the composer in the stream

@mzorai See with archi, we will use this name for the prooperty exo.config.mainStreamComposer.enabled

mzorai:
> > > property name does not clearly reflect that it is the composer on stream page that is configured.
> > 
> > 
> > I proposed this property name exo.social.composer.stream.hidden
> > true: to disable displaying the composer in the stream
> > false: to enable displaying the composer in the stream
> 
> @mzorai See with archi, we will use this name for the prooperty exo.config.mainStreamComposer.enabled

OK so it should be true to enable and false to disable

azayati (approved):
LGTM

PR #156, opened 2020-06-30
MAINT-24540 : Update regex to fix Notification parsing for Edge

ahamdi:
With Edge browser, Images included in notifications are not replaced with the Inline Image placeholder.
The regex was not valid for HTML produced from Edge, this fix takes care of that difference between browsers.


ghost (approved):

ahmadmido (approved):
LGTM

PR #164, opened 2020-07-02
MAINT-26328 : Feature/news: Update activity composer extension

aelbenney:
In order to make it possible for plugins to make rest calls, modifications,.. before self init, we added an event listener `activity-composer-extension-updated` to update the composer extension whenever needed.


webapp/vue-portlet/pom.xml
```diff
@@ -0,0 +1,125 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
boubaker: Please pay attention, this file is not needed anymore

ahamdi (approved):
LGTM

ahamdi (changes requested):
Please open a task on maintenance.

ahamdi (approved):
LGTM

PR #165, opened 2020-07-02
Fix Search Connectors API to allow retrieving data from external connectors

boubaker:
When attempting to add a wikipedia Search connector, I've found that the Origin information is needed. This change will avoid to hide this information when requesting Search resources from external sites


boubaker:
cc @hghorbel I let you include this in scope of 6.0.x (to see with others)

@ahamdi , @mzorai  WDYT ? (This will enable adding Wikipedia Search Connector for example)

azayati (approved):
LGTM

PR #170, opened 2020-07-03
MAINT-26392 : Fix avatar file type validation

boubaker:
The file type of avatar and banner are already restricted in 'HTML file input' by 'accept' attribute, but in some OS, they give possibility to change type of files that could be attached. This PR will avoid uploading files different from type 'image/*'


ahamdi (approved):
LGTM

PR #178, opened 2020-07-07
Maint-26430 : Feature New Composer :  When exceeding 1300 char in the new composer a wrong label is displayed

Rdidmohamedelhadi:
Problem : when exceeding 1300 char in the new composer in the activity stream the action switch to an article is displayed so the fix is adding a check for existing space


ahamdi (approved):
LGTM

PR #192, opened 2020-07-16
MAINT-26790: Add error msg for space name length

AzmiTouil:


component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
```diff
@@ -232,10 +232,13 @@ public Response createSpace(@Context UriInfo uriInfo,
                                                 "<br />\"visibility\": \"private\"," +
                                                 "<br />\"subscription\": \"validation\"<br />}" 
                                                 , required = true) SpaceEntity model) throws Exception {
-    if (model == null || model.getDisplayName() == null || model.getDisplayName().length() == 0 || model.getDisplayName().length() > 200 || !SpaceUtils.isValidSpaceName(model.getDisplayName())) {
+    if (model == null || model.getDisplayName() == null || model.getDisplayName().length() == 0 || !SpaceUtils.isValidSpaceName(model.getDisplayName())) {
       throw new WebApplicationException(Response.Status.BAD_REQUEST);
```
ahamdi: Why don't you change the exception here instead of adding another if block. IMO this is the block to validate the space display name.

ahamdi (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
AzmiTouil: But this is a generic exception includes many cases, in fact i need a specific exception to inform the user that it is over 200 characters in the space name. WDYT ?

AzmiTouil: Done !

component/core/src/main/java/org/exoplatform/social/core/space/SpaceException.java
```diff
@@ -76,7 +76,10 @@
 
     /** The SPAC e_ alread y_ exist. */
     SPACE_ALREADY_EXIST,
-    
+
+    /** INVALID space name length */
+    INVALID_SPACE_NAME_LENGTH,
```
ahamdi: Let's make the error more generic
```suggestion
    INVALID_SPACE_NAME,
```

webapp/portlet/src/main/resources/locale/portlet/social/SpacesListApplication_en.properties
```diff
@@ -43,6 +43,7 @@ spacesList.warning.descriptionExceededLength=The length of the text in field "De
 spacesList.warning.requiredDisplayName=Display name is required.
 spacesList.warning.lastManager=You are the last manager of this space. You need to promote another member as manager of the space before you can leave it.
 spacesList.error.spaceWithSameNameExists=A space with the same name already exists
+spacesList.error.spaceNameLengthInvalid=The length of the text in field "name" must be between 3 and 200 characters.
```
ahamdi: Please add unwanted unauthorized characters in the error message.

ahamdi (changes requested):

ahamdi (approved):
LGTM

PR #193, opened 2020-07-16
Maint-26740 : Fix Edit Experiences information when End date or start date is greater than the present.

ahmadmido:
This PR will fix the start and end date in Edit Experiences information when selected date is greater than the present.


webapp/portlet/src/main/webapp/profile-work-experience/components/ProfileWorkExperienceDrawer.vue
```diff
@@ -83,6 +83,14 @@ export default {
           this.handleError(this.$t('profileWorkExperiences.invalidEndDate'));
           return;
         }
+        if (experience.startDate && experience.endDate && new Date(experience.endDate) > new Date()) {
```
ahamdi: Why do you need to check the startDate here ?

ahmadmido: Done!

ahamdi (approved):
LGTM

PR #195, opened 2020-07-16
Add icons for activity drop-down menu items

SaraBoutej:
In this PR we add icons to all social activity drop-down menu items


ahamdi (approved):
LGTM

PR #198, opened 2020-07-17
Maint-26836:Feature Group binding : Leave button is not disabled

Rdidmohamedelhadi:


webapp/portlet/src/main/webapp/spaces-list/components/ExoSpaceCardFront.vue
```diff
@@ -114,7 +114,7 @@
       <v-btn
         v-if="space.isMember"
         :loading="sendingAction"
-        :disabled="sendingAction"
+        :disabled="sendingAction || space.hasBindings"
```
ahamdi: What if the user is a member and was not bound to that space ?

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
```diff
@@ -182,6 +183,15 @@ public Response getSpaces(@Context UriInfo uriInfo,
         space.setHasBindings(spaceBindingService.isBoundSpace(space.getId()));
         
         SpaceEntity spaceInfo = EntityBuilder.buildEntityFromSpace(space, authenticatedUser, uriInfo.getPath(), expand);
+        List<UserSpaceBinding> userSpaceBinding = spaceBindingService.findUserSpaceBindingsBySpace(space.getId(), authenticatedUser);
+        spaceInfo.setIsUserBound(false);
+        if (userSpaceBinding.size() != 0) {
+          for (int i = 0; i < userSpaceBinding.size(); i++) {
+            if (authenticatedUser.equals(userSpaceBinding.get(i).getUser()) && !userSpaceBinding.get(i).isMemberBefore()) {
+              spaceInfo.setIsUserBound(true);
+            }
```
ahamdi: This is unneeded I think, you have just to check the size of userSpaceBinding, and if it is bigger than zero, you can call the function org.exoplatform.social.core.binding.impl.GroupSpaceBindingServiceImpl#isUserBoundAndMemberBefore to check if the user is bound before, otherwise, you enable the leave button

ahamdi (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
Rdidmohamedelhadi: ok

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
```diff
@@ -182,6 +183,11 @@ public Response getSpaces(@Context UriInfo uriInfo,
         space.setHasBindings(spaceBindingService.isBoundSpace(space.getId()));
         
         SpaceEntity spaceInfo = EntityBuilder.buildEntityFromSpace(space, authenticatedUser, uriInfo.getPath(), expand);
+        List<UserSpaceBinding> userSpaceBinding = spaceBindingService.findUserSpaceBindingsBySpace(space.getId(), authenticatedUser);
+        spaceInfo.setIsUserBound(false);
```
boubaker: To know if user is bound from group or not, we seems to have only one method :
`spaceBindingService.countUserBindings(space.getId(), userId) > 0`
@rdenarie , do you confirm ?

rdenarie: Yes I confirm. In fact, we add it in webui utils file : 
https://github.com/Meeds-io/social/blob/develop/component/webui/src/main/java/org/exoplatform/social/webui/Utils.java#L889
It could be usefull to add it in GroupSpaceBindingService if needed in different location

boubaker: Thank you @rdenarie .

@Rdidmohamedelhadi I let you change this into : 

`spaceInfo.setIsUserBound(groupSpaceBindingService.countUserBindings(spaceId, authenticatedUser) > 0)`
(we should avoid adding WebUI artifact dependency)

Just another concern, I think this information is heavy when requesting the list of spaces, so we should retrieve this field only when `expand` contains `"all"`, @ahamdi WDYT ?

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
```diff
@@ -182,6 +183,7 @@ public Response getSpaces(@Context UriInfo uriInfo,
         space.setHasBindings(spaceBindingService.isBoundSpace(space.getId()));
         
         SpaceEntity spaceInfo = EntityBuilder.buildEntityFromSpace(space, authenticatedUser, uriInfo.getPath(), expand);
+        spaceInfo.setIsUserBound(spaceBindingService.countUserBindings(space.getId(), authenticatedUser) > 0);
```
boubaker:
```suggestion
        if (space.hasBindings()) {
          spaceInfo.setIsUserBound(spaceBindingService.countUserBindings(space.getId(), authenticatedUser) > 0);
        }
```

boubaker: If space doesn't has binding, no need to compute `IsUserBound` because it will be always `false`

Rdidmohamedelhadi: Aaah yes !! Thanks @boubaker

boubaker (approved):

ahamdi (approved):

PR #199, opened 2020-07-20
Fix changing user language setting with locale using country

boubaker:
The controller.xml defines a list of allowed languages that could be defined by user. This list of languages uses "-" instead of "_" which should both be correct. This change will convert locale code using `"_"` into `"-"` to let Gatein controllers parse the correct language code.


ahamdi (approved):
LGTM

PR #201, opened 2020-07-20
Maint-26893: show alert message when adding group

hassenelkamel-exo:
**Problem:** none popup shown when user enter invalid character for group name.
**Solution:** add regex to check group name


webapp/portlet/src/main/webapp/idm-groups-management/components/GroupsManagementFormDrawer.vue
```diff
@@ -161,6 +170,13 @@ export default {
       this.drawer = true;
     },
     saveGroup(event) {
+      const regex =  /^[a-zA-Z0-9-_]+$/;
```
ahamdi, line 164: It is better if you handle this at Rest service level as it is already done for validating if name or label are empty.

ahamdi (changes requested):

webapp/portlet/src/main/webapp/idm-groups-management/components/GroupsManagementFormDrawer.vue
hassenelkamel-exo: done

webapp/portlet/src/main/webapp/idm-groups-management/components/GroupsManagementFormDrawer.vue
```diff
@@ -161,6 +161,15 @@ export default {
...
+      const isValid = regex.test(this.group.groupName);
+      if (!isValid){
+        this.error = this.$t('GroupsManagement.popup.message');
```
boubaker: you can call `this.handleError(this.$t('GroupsManagement.popup.message'))` instead.
+
`GroupsManagement.popup.message` is not a significant key, you can use `GroupsManagement.error.invalidGroupName` for example.

webapp/portlet/src/main/webapp/idm-groups-management/components/GroupsManagementFormDrawer.vue
```diff
@@ -96,7 +96,7 @@ export default {
     saving: false,
     confirmNewPassword: null,
     group: {},
-    parentGroup: null,
+    parentGroup: null
```
boubaker, line 99: Just for curiosity, why this change ?

hassenelkamel-exo: in the first time, i add a popup and new variable popupBodyMessage and ali tell me that you don't need to add those so I revert code

webapp/portlet/src/main/webapp/idm-groups-management/components/GroupsManagementFormDrawer.vue
```diff
@@ -186,9 +195,8 @@ export default {
       }).then(resp => {
         if (!resp || !resp.ok) {
           if (resp.status === 400) {
-            return resp.text().then(error => {
-              this.fieldError = error;
-              throw new Error(error);
+            return resp.text().then(() => {
+              throw new Error(this.$t('GroupsManagement.popup.message'));
```
boubaker: This change will affect the other errors to not display properly

boubaker (changes requested):

webapp/portlet/src/main/webapp/idm-groups-management/components/GroupsManagementFormDrawer.vue
hassenelkamel-exo: ok done

webapp/portlet/src/main/webapp/idm-groups-management/components/GroupsManagementFormDrawer.vue
hassenelkamel-exo: done

boubaker (approved):

ahamdi (approved):
LGTM

PR #213, opened 2020-07-22
Maint-26952: check the validity of role name and role description

hassenelkamel-exo:
- add a check to insure that the name must be between "3" and "30" characters, The length of the text in field "Role Description" must be between "3" and "255" characters in front side.


ahamdi (changes requested):
Check the Build failure of the PR

ahamdi (approved):
LGTM

PR #219, opened 2020-07-23
MAINT-26916: Add custom checks to the add/edit group form.

mdiha:
* I added a check on text length of group Name field in order to display a warning message if the length is lesser than 3 characters or greater than 30 characters.
* I added a check on group Name field in order to display a warning message if it starts with a number.
* I added a check on text length of group Label field in order to display a warning message if the length is lesser than 3 characters or greater than 50 characters.
* I added a check on text length of group Description field in order to display a warning message if the length is greater than 255 characters.


ahamdi (changes requested):
Add description to the PR

mdiha:
> Add description to the PR

Okay

ahamdi:
Will this block the saving of the group ?

mdiha:
> Will this block the saving of the group ?

yes it does till the label length is between 3 and 50 characters

webapp/portlet/src/main/webapp/idm-groups-management/components/GroupsManagementFormDrawer.vue
```diff
@@ -161,20 +161,64 @@ export default {
       this.drawer = true;
     },
     saveGroup(event) {
+      const startWithRegex =  /^\d/;
       const regex =  /^[a-zA-Z0-9-_]+$/;
```
ahamdi: Could you combine those two regex in one ?

mdiha: First one is for: Name starts with number: "The "Group Name" field must start with a character and must not contain special characters."
Second one is for: special characters.
And each one is used apart.

mdiha: Done

ahamdi (approved):

PR #223, opened 2020-07-23
MAINT-26972 Fix saving about Me user information using correct encoding

boubaker:
When appending manually Query string of POST Form data, the characters encoding of the sent request will be ignored, thus the server will decode the form data using bad Character encoding. By using javascript FormData object to simulate sending a form, the data is correctly encoded and thus correctly saved in Server side using the correct Character encoding


ahamdi (approved):
LGTM

PR #227, opened 2020-07-27
Enable user profile activity generation for avatar and banner

boubaker:
When a user changes his avatar or banner, the activity isn't populated anymore. In fact, the IdentityManager should detect automatically this change, but this wasn't possible to specify in Service layer, thus the update Type for Avatar and Banner attachments modifications are managed in REST layer by adding manually the update type to correctly dispatch Profile Update operation listeners.


ahamdi (approved):
LGTM

PR #230, opened 2020-07-28
MAINT-27093: Make sure that bound users can not be removed from the space to which they are bound.

mdiha:
According to the group space binding rules, bound user can not be removed from a space to which they are bound, i added a check to verify if a user is already bound to the space in concern before enabling the "Remove member" extension.


ahamdi (approved):
LGTM

PR #231, opened 2020-07-28
MAINT-27093: Make sure that bound users can not be removed from the space to which they are bound.

mdiha:
Backport to stable/6.0.x


ahamdi (approved):
LGTM

PR #243, opened 2020-08-18
MAINT-25739: Session TIMEOUT Problem, No popup is appeared when the session expired

AzmiTouil:
Add a popup when session expires to tell user to backup your editor content


ahamdi (changes requested):
We can not rely on Javascript to count when a session is expired. Check PortalHttpRequest.js to see how to handle the session expiration for Ajax requests

PR #244, opened 2020-08-19
Maint-27530 : No warning message appears when length of space name is less than 3 characters

Rdidmohamedelhadi:
Problem : no warning message appears when length of space name is less than 3 characters so the fix is : add this condition `model.getDisplayName().length() < 3 `


azayati (approved):
LGTM

PR #259, opened 2020-09-07
MAINT-27504: Add redactor role to space members filters

azayati:
We need to add redactors filter to the space portlet filters in order to display the list of space redactors

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/space/impl/SpaceServiceImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/space/impl/SpaceServiceImpl.java
@@ -17,14 +17,24 @@
 package org.exoplatform.social.core.space.impl;
 
 import java.io.InputStream;
-import java.util.*;
-import java.util.stream.Collectors;
+import java.util.ArrayList;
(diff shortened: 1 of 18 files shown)


component/core/src/main/java/org/exoplatform/social/core/space/impl/SpaceServiceImpl.java
```diff
@@ -1475,6 +1509,29 @@ private void removeWebNotifications(String spaceId, String userId) {
       LOG.error("Cannot update web notfication. WebNotificationService is null");
     }
   }
+  
+  /**
+   * {@inheritDoc}
+   */
+  public void setRedactor(Space space, String userId, boolean isRedactor) {
```
ahamdi: IMO we need to split this and add functions addRedactor, removeRedactor instead of having one single one for both.

azayati: I have done same as setManager

ahamdi (dismissed):
LGTM

ahamdi (approved):

PR #274, opened 2020-09-10
MAINT-28235 : use userName as value for users suggested from a space

ahamdi:
When suggesting users who are members of a space, the list of items provides the fullname of the user as value and this does not allow using it with other APIs
The fix sets the username as the value for each suggested item.


mkrout (approved):
LGTM

azayati (approved):
LGTM

PR #280, opened 2020-09-14
MAINT-28293: Add limitation for organization and job title fields.

mdiha:
I added fields checks in order that Job title and Organization fields must not start or end with a white space.

### DIFF
--- a/webapp/portlet/src/main/webapp/profile-work-experience/components/ProfileWorkExperienceDrawer.vue
+++ b/webapp/portlet/src/main/webapp/profile-work-experience/components/ProfileWorkExperienceDrawer.vue
@@ -70,7 +70,6 @@ export default {
   methods: {
     save() {
       this.error = null;
-      const regex = /^[^\s]+(\s+[^\s]+)*$/;
 
       const experiences = this.experiences.filter(experience => experience && (experience.startDate || experience.endDate || experience.position || experience.company || experience.description || experience.skills));
       for (const experience of experiences) {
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/profile-work-experience/components/ProfileWorkExperienceDrawer.vue
```diff
@@ -70,6 +70,7 @@ export default {
   methods: {
     save() {
       this.error = null;
+      const regex = /^[^\s]+(\s+[^\s]+)*$/;
```
ahamdi: Use String.trim() and check that the text is bigger than minimum field length

ahamdi (changes requested):

webapp/portlet/src/main/webapp/profile-work-experience/components/ProfileWorkExperienceDrawer.vue
mdiha: So no need to display a warning message ?

ahamdi: For me yes, simply we get rid of all spaces and make the text get checked with the default rules.

ahamdi: @mzorai WDYT ?

mzorai: it's OK for me to trim the seized value at the begining and the end from spaces

mdiha: If field contains more then a white space at the beginning or the end we end up in the same situation. WDYT ?

ahamdi (approved):
LGTM

PR #284, opened 2020-09-15
MAINT-28140: Impossible to create a space which the registration is closed

Rdidmohamedelhadi:
**Problem :** Impossible to create a space which the registration is closed
**Fix :** I changed the value of check box


ahamdi (approved):
LGTM

PR #291, opened 2020-09-18
displaying correctly avatar and make square form for space avatar

hassenelkamel-exo:
-make the space avatar in the square form
- display correctly the avatar user instead of displaying the default avatar

### DIFF
--- a/webapp/portlet/src/main/webapp/common/js/SuggesterService.js
+++ b/webapp/portlet/src/main/webapp/common/js/SuggesterService.js
@@ -56,7 +56,7 @@ function searchUsers(filter, items, typeOfRelation, searchOptions) {
               providerId: 'organization',
               profile: {
                 fullName: item.name || item.text,
-                avatarUrl: item.avatar || item.avatarUrl || `/portal/rest/v1/social/users/${username}/avatar`,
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/common/js/SuggesterService.js
```diff
@@ -56,7 +56,7 @@ function searchUsers(filter, items, typeOfRelation, searchOptions) {
               providerId: 'organization',
               profile: {
                 fullName: item.name || item.text,
-                avatarUrl: item.avatar || item.avatarUrl || `/portal/rest/v1/social/users/${username}/avatar`,
+                avatarUrl: item.avatarUrl ? item.avatarUrl : `/portal/rest/v1/social/users/${username}/avatar`,
```
boubaker: Why changing this expression ? doesn't it work for bringing **Users** avatars ?
As you can see, the method name is `searchUsers`. For spaces, there is another method called `searchSpaces`

hassenelkamel-exo: I change it because in event creation when adding a participants we cannot retrieve the avatar of user, all time we retrieve the default avatar.

PR #294, opened 2020-09-21
MAINT-28485: Make sure that only the activity owner has the right to edit it and that administrators can only delete it.

mdiha:
I made sure that only the activity's owner is allowed to edit it but instead platform administrators have only the right to delete t.

### DIFF
--- a/component/webui/src/main/java/org/exoplatform/social/webui/activity/BaseUIActivity.java
+++ b/component/webui/src/main/java/org/exoplatform/social/webui/activity/BaseUIActivity.java
@@ -21,6 +21,7 @@ import org.apache.commons.lang3.StringUtils;
 
 import org.exoplatform.commons.utils.CommonsUtils;
 import org.exoplatform.commons.utils.DateUtils;
+import org.exoplatform.portal.config.UserACL;
 import org.exoplatform.portal.webui.util.Util;
 import org.exoplatform.services.log.ExoLogger;
 import org.exoplatform.services.log.Log;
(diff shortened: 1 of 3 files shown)


component/webui/src/main/java/org/exoplatform/social/webui/activity/BaseUIActivity.java
```diff
@@ -577,8 +578,9 @@ public boolean isActivityDeletable() throws SpaceException {
       }
     }
 
+    UserACL userACL = getApplicationComponent(UserACL.class);
     if (space != null) {
-      return spaceService.isManager(space, Utils.getOwnerRemoteId());
+      return userACL.isSuperUser() || spaceService.isManager(space, Utils.getOwnerRemoteId()) || ConversationState.getCurrent().getIdentity().getGroups().contains(userACL.getAdminGroups());
```
ahamdi: We need to check if the user is a super manager (member of space administration group)
```suggestion
      return userACL.isSuperUser() || spaceService.isManager(space, Utils.getOwnerRemoteId()) || spaceService.isSuperManager(Utils.getOwnerRemoteId())
```

ahamdi (changes requested):

component/webui/src/main/java/org/exoplatform/social/webui/activity/BaseUIActivity.java
mdiha: Okay i will add it

component/webui/src/main/java/org/exoplatform/social/webui/activity/BaseUIActivity.java
```diff
@@ -577,8 +578,9 @@ public boolean isActivityDeletable() throws SpaceException {
       }
     }
 
+    UserACL userACL = getApplicationComponent(UserACL.class);
     if (space != null) {
-      return spaceService.isManager(space, Utils.getOwnerRemoteId());
+      return userACL.isSuperUser() || spaceService.isSuperManager(Utils.getOwnerRemoteId()) || spaceService.isManager(space, Utils.getOwnerRemoteId()) || ConversationState.getCurrent().getIdentity().getGroups().contains(userACL.getAdminGroups());
```
ahamdi: The check on platform administrators group is no more needed as we check on Space administration group

mdiha: Done

ahamdi (approved):

PR #299, opened 2020-09-22
MAINT-28585: Display order doesn't work when space name is edited

AzmiTouil:
**activity.getActivityStream().getPrettyId()** return the space **PrettyName** , for that the **spaceGroupId** becomes null if the space name has been changed


ahamdi (approved):
LGTM

PR #306, opened 2020-09-25
MAINT-28623:  Uploading documents trigger a post on the activity stream of the space

Rdidmohamedelhadi:
**Problem :**  Uploading documents trigger a post on the activity stream of the space
**Fix :** I changed the default value of properties (f`iles:spaces.CREATION_COMMENT`,`files:spaces.RENAME_COMMENT`,`files:spaces.MOVE_COMMENT`,`files:spaces.TAG_ACTION_COMMENT`) to false


ahamdi (changes requested):
Simpler solution, put their values to false by default in configuration.properties

PR #312, opened 2020-09-28
MAINT-26920 : Make sure to use the right meta tag for link images

ahamdi:
To generate an image for shared links, The LinkShare class was using just the Link tag with "image_src" as rel, then it looks for the images inside the body of the web page.
This fix changes the current behavior to make it similar to what is used in the Oembed plugin of CKEditor (https://github.com/nfl/jquery-oembed-all) :

1. Look for Meta tag with property og:image -> the open Graph protocol https://ogp.me/
2- Otherwise, look for the Link tag with "image_src" and get its content
3- Finally look for images inside the web page


Rdidmohamedelhadi (approved):

mdiha (approved):
LGTM

PR #313, opened 2020-09-29
Task 28722 : the close search button does not respect a9n criteria 1.1.1 Non Text Content

rdenarie:
Prior to this change, the button is empty (has no text), so reading tools are not able to read it.
This fix add an aria-label on the button, then reading tools can read the button
In addition, after the fix, the wave plugin does not mark this as an error.

### DIFF
--- a/webapp/portlet/src/main/webapp/search/components/SearchToolbar.vue
+++ b/webapp/portlet/src/main/webapp/search/components/SearchToolbar.vue
@@ -15,7 +15,7 @@
       <v-btn v-if="term" text color="error" @click="clearSearchTerm">
         {{ $t('search.connector.label.clear') }}
       </v-btn>
-      <v-btn v-if="!standalone" icon class="searchCloseIcon transparent" @click="$emit('close-search')">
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/search/components/SearchToolbar.vue
```diff
@@ -15,7 +15,7 @@
       <v-btn v-if="term" text color="error" @click="clearSearchTerm">
         {{ $t('search.connector.label.clear') }}
       </v-btn>
-      <v-btn v-if="!standalone" icon class="searchCloseIcon transparent" @click="$emit('close-search')">
+      <v-btn v-if="!standalone" icon class="searchCloseIcon transparent" aria-label="Close search" @click="$emit('close-search')">
```
ahamdi: This should be i18n IMO

rdenarie: Updated

ahamdi (approved):

PR #324, opened 2020-10-04
QWIN/Maint-28474 : Improve UX for spaces drawer in welcome back widget

ahmadmido:
This PR  will add a REST service for commons spaces to have on the user profile.


ahamdi (changes requested):
We can not load all spaces for each user and check if one is member of the other one spaces.
This should be queried directly from database.

ahmadmido:
> We can not load all spaces for each user and check if one is member of the other one spaces.
> This should be queried directly from database.



> We can not load all spaces for each user and check if one is member of the other one spaces.
> This should be queried directly from database.

As discussed with Ali, we will adopt this solution and we need to implement a unit tests to not have more complexity.

ahamdi (approved):
LGTM

PR #329, opened 2020-10-05
MAINT-28977: Fix Stream name label when posting in stream directly

hbenali:
When activity composer is enabled for posting inside the stream.  "Post in Post" appears.


ahamdi (commented):
Could you create a task for this ?

ahamdi (approved):
LGTM

PR #335, opened 2020-10-09
MAINT-35249 : Add space drawer appears when making a search

ahmadmido:
This PR will fix the problem of appears drawer when marking a search.


webapp/portlet/src/main/webapp/spaces-list/components/ExoSpaceFormDrawer.vue
```diff
@@ -258,7 +258,7 @@ export default {
     },
   },
   mounted() {
-    if(location.search){
+    if(location.search === '?createSpace=true'){
```
ahamdi:
```suggestion
    if (window.location.href.indexOf('/all-spaces?createSpace=true') > -1) {
```

ahamdi (approved):

PR #349, opened 2020-10-16
Task 35529 : Space template names and description are not i18n (#348)

rdenarie:
Before this fix, space templates names and descriptions are always in english.
This fix use the user locale to resolve name label, instead of en locale hardcoded


ahamdi (approved):
LGTM

PR #352, opened 2020-10-20
MAINT-25906: Make sure that after liking a comment, Smileys doesn't get displayed as an encoded text in activity comments.

mdiha:
Avoid the unnecessary update of comment title, body and template params for the like comment action.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/manager/ActivityManagerImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/manager/ActivityManagerImpl.java
@@ -402,6 +402,10 @@ public class ActivityManagerImpl implements ActivityManager {
    * {@inheritDoc}
    */
   public void saveLike(ExoSocialActivity existingActivity, Identity identity) {
+    // in order to avoid updating unnecessarily activity title, body and template params 
+    existingActivity.setTitle(null);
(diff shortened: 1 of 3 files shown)


component/core/src/main/java/org/exoplatform/social/core/manager/ActivityManagerImpl.java
```diff
@@ -402,6 +402,9 @@ public void deleteComment(ExoSocialActivity existingActivity, ExoSocialActivity
    * {@inheritDoc}
    */
   public void saveLike(ExoSocialActivity existingActivity, Identity identity) {
+    existingActivity.setTitle(null);
+    existingActivity.setBody(null);
+    existingActivity.setTemplateParams(null);
```
ahamdi, line 408: This workarounds the problem, could we know why they are encoded and prevent it.

ahamdi (changes requested):

component/core/src/main/java/org/exoplatform/social/core/manager/ActivityManagerImpl.java
mdiha: Okay i will check

ahamdi (approved):

PR #357, opened 2020-10-22
FIX-27770 : The global search bar can't find some users

Rdidmohamedelhadi:
**Problem :** The global search bar can't find some users
**Solution :** Improve user search


ahamdi (dismissed):

ahamdi (approved):
LGTM

PR #365, opened 2020-10-26
Task-35860 Improve Loading Skeleton of Vue applications

boubaker:
Knowing that rendering of the page is made in Browser side and not Server side (No use of NodeJS as rendering engine in server side yet), a skeleton was using to let the browser render the application and displays the final result afterwards.

The old skeleton brings a lot of HTML to download from Server side, in addition, the Browser was making a lot of computing to try refresh the UI by new changes made on application, and at the same time brings data from server side.

The solution that is brought by this PR is based on a new Service worker [defined in this PR](https://github.com/Meeds-io/gatein-portal/pull/86) that will append cached DOM in HTML Fetching request using custom tag name `v-cacheable-dom-app` [defined in this PR](https://github.com/Meeds-io/gatein-portal/pull/86).

A new Vue directive is added to cache Vue application DOM when the application emits an event on `$root.$emit`:
* `$root.$emit('application-mount')`: to handle application real HTML rendering by replacing "**displayed cached DOM**" by "**real application DOM"**. This event will be handled only once by application (`$root.$once`).
* `$root.$emit('application-cache')`: refresh cached DOM by "**real application DOM**" 
* `$root.$emit('application-loaded')`: combinaison between `$root.$emit('application-mount')` and `$root.$emit('application-cache')`

The rendering of a cached application is so made in background, thus its computation is faster because DOM changes, due to Application first loading, aren't interpreted by UI Engine that will have to refresh the display while building application. In addition, the user has a first chunck of information that are outdated but real.


ahamdi (approved):
LGTM

PR #366, opened 2020-10-26
MAINT-28561: Avoid not found real users from joining a space

azayati:
This PR will avoid not real users to join spaces, a sanitized url is generated in that case

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/service/rest/NotificationsRestService.java
+++ b/component/service/src/main/java/org/exoplatform/social/service/rest/NotificationsRestService.java
@@ -19,6 +19,8 @@ package org.exoplatform.social.service.rest;
 import org.exoplatform.container.ExoContainer;
 import org.exoplatform.container.ExoContainerContext;
 import org.exoplatform.container.PortalContainer;
+import org.exoplatform.services.organization.OrganizationService;
+import org.exoplatform.services.organization.User;
 import org.exoplatform.services.rest.resource.ResourceContainer;
 import org.exoplatform.services.security.ConversationState;
 import org.exoplatform.social.core.identity.model.Identity;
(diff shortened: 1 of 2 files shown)


ahamdi (changes requested):
This will block the user from joining the space and won't give him a replacement for that. Could we have a sanitized link to let user join the space ?
Could we add a '/' at the end to make sure that the PathParam takes the correct value from URL and change accordingly in the web notification ?

component/service/src/main/java/org/exoplatform/social/service/rest/NotificationsRestService.java
```diff
@@ -255,11 +258,11 @@ public Response validateRequestToJoinSpace(@PathParam("spaceId") String spaceId,
     checkAuthenticatedRequest();
 
     Space space = getSpaceService().getSpaceById(spaceId);
-    
-    if (space == null) {
+    OrganizationService organizationService = (OrganizationService) getPortalContainer().getComponentInstanceOfType(OrganizationService.class);
+    User user = organizationService.getUserHandler().findUserByName(userId);
```
ahamdi: You could use identityManager as it performs better that OrganizationService org.exoplatform.social.core.manager.IdentityManagerImpl#getIdentity(java.lang.String, java.lang.String, boolean)

azayati: done

azayati:
@ahamdi:
> This will block the user from joining the space and won't give him a replacement for that. Could we have a sanitized link to let user join the space ?

Sanitization done for urls containing jsessionId pattern

> Could we add a '/' at the end to make sure that the PathParam takes the correct value from URL and change accordingly in the web notification ?

IMO this is not need for 6.1.x/6.0.x since IntranetNotificationRestService API is used for notification and the concerned API NotificationsRestService is no more used on space members invitation validation. We should check if it is needed and still used on 5.3.x

ahamdi (approved):

PR #374, opened 2020-10-27
MAINT-35728: Make sure to fix access to spaces with Cyrillic name.

mdiha:
I made sure to clear modifier letters from space's pretty name while adding a space.


ahamdi (changes requested):
Problem not reproduced using a Mysql DB server

PR #375, opened 2020-10-27
35824 : Add lastUpatedTime property to space

ahamdi:
A new property was added to the space entity to store the last updated time. This property will be used to add an ETAG to the Rest response when requesting a space.


component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSSpaceStorageImpl.java
```diff
@@ -505,6 +507,7 @@ public void saveSpace(Space space, boolean isNew) throws SpaceStorageException {
       if (entity != null) {
         EntityConverterUtils.buildFrom(space, entity);
         //
+        entity.setUpdatedDate(new Date());
```
boubaker, line 510: We may need to update the updatedDate when renaming a space too : 
`org.exoplatform.social.core.jpa.storage.RDBMSSpaceStorageImpl.renameSpace(String, Space, String)`

(Just to share with you some checks made too, it could help for further modification to other entities: Knowing that we should also update the date when a new user joins/leaves the space, but at the same time, the spaceService.addMember will update the space entity using the method `saveSpace`, no additional modifications are required)

ahamdi: - Yes the PR includes already updating the date when a space is renamed.
 - Yes, whenever a member is added/removed the update date will be removed too

boubaker (approved):

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSSpaceStorageImpl.java
boubaker: This could be a good pattern to use for other served `entities` by REST API. Thanks a lot @ahamdi 
( We may need to add a new guideline for this in `REST API Guideline`, WDYT ? )

ahamdi: @boubaker OK I will, good idea. I will also dedicate a training session to explain the goal and benefits of caching data and go in details on what was done here. Thanks for the review

PR #383, opened 2020-10-29
Task_35995: Make it possible le to reimport existing user by setting all attributes

azayati:


component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -1131,7 +1131,7 @@ private String importUser(UserImportResultEntity userImportResultEntity,
 
     User existingUser = organizationService.getUserHandler().findUserByName(userName, UserStatus.ANY);
     if (existingUser != null) {
-      userImportResultEntity.addErrorMessage(userName, "USERNAME:ALREADY_EXISTS");
+      organizationService.getUserHandler().saveUser(user, true);
```
rdenarie, line 1134: Why do we save user ? to update fields ?

rdenarie (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
azayati: yes

rdenarie (dismissed):

rdenarie (approved):

PR #411, opened 2020-11-12
MAINT-25851 : Space group Binding : Long group name Display issue

Rdidmohamedelhadi:
**Problem :** when the group name length is bigger than 48 char the name is not displayed correctly.
**Solution:** fix style


ahamdi:
Please provide a description

ahamdi (approved):

PR #412, opened 2020-11-16
Task-36530: Make Global site and pages accessible by /platform/externals group members

azayati:


extension/war/src/main/webapp/WEB-INF/conf/social-extension/portal/portal/global/portal.xml
```diff
@@ -2,7 +2,7 @@
 <portal-config>
     <portal-name>global</portal-name>
     <locale>en</locale>
-    <access-permissions>*:/platform/users</access-permissions>
+    <access-permissions>*:/platform/users;*:/platform/externals</access-permissions>
```
plamarque, line 5: isn't it dangerous to provide such wide access to external users ? The global navigaiton contains pages that they are not authorized to see (such as people, spaces and others). How is the permission check is done ? at page level ?

azayati: You have anticipated, despite this configuration, right now all global pages are accessible to only /platform/users. The goal is not to give externals full access to global site pages, the work is not yet finished and the appropriate permissions will be given effectively at page level.

plamarque: > the work is not yet finished and the appropriate permissions will be given effectively at page level.

OK. Just make sure you don't push this change before you safeguard page permissions.

ahmadmido (approved):
LGTM

PR #419, opened 2020-11-18
Task-36806 : When importing csv user file with a new line at the end,…

rdenarie:
… we cannot import a new file after

Before this commit, when reading the csv file, if a line is empty, we don't read a new line, so we are stuck in infinite loop, preventing the executor to start new task
This commit add a readLine instruction when the read line is empty, so that we can continue in the file and reach the end.


ahamdi (commented):
Could you rebase the branch as it contains more commits than the needed fix?

ahamdi (dismissed):

ahamdi (approved):
LGTM

PR #427, opened 2020-11-24
MAINT-35949: Add lastUpdatedDate property to identity.

mdiha:


ahamdi (changes requested):
- Description is required
- Unit tests are mandatory
- Add the same logic for all endpoints that are returning a user identity in their responses check this class UserRestResourcesV1.java

PR #442, opened 2020-12-02
Add extension point to space title

NickEngineer:
This PR moves extension point from Space Menu to Space Title.

### DIFF
--- a/webapp/portlet/src/main/webapp/space-menu/extension.js
+++ b/webapp/portlet/src/main/webapp/space-menu/extension.js
@@ -4,7 +4,7 @@ function getExtensionsByType(type) {
 
 export const spaceMenuActionComponents = getExtensionsByType('action-component');
 
-function getExtensionsByTypeCallButton(type) {
+function getSpaceTitleExtensionsByType(type) {
   return extensionRegistry.loadExtensions('space-title-action-components', type);
 }
 
--- a/webapp/portlet/src/main/webapp/space-menu/extension.js
+++ b/webapp/portlet/src/main/webapp/space-menu/extension.js
@@ -8,4 +8,4 @@ function getExtensionsByTypeCallButton(type) {
   return extensionRegistry.loadExtensions('space-title-action-components', type);
 }
 
-export const spaceMenuCallActionComponents = getExtensionsByTypeCallButton('action-component');
+export const spaceTitleActionComponents = getExtensionsByTypeCallButton('action-component');
--- a/webapp/portlet/src/main/webapp/space-menu/extension.js
+++ b/webapp/portlet/src/main/webapp/space-menu/extension.js
@@ -3,3 +3,9 @@ function getExtensionsByType(type) {
 }
 
 export const spaceMenuActionComponents = getExtensionsByType('action-component');
+
+function getExtensionsByTypeCallButton(type) {
+  return extensionRegistry.loadExtensions('space-title-action-components', type);
+}
+
+export const spaceMenuCallActionComponents = getExtensionsByTypeCallButton('action-component');
--- a/webapp/portlet/src/main/webapp/space-menu/extension.js
+++ b/webapp/portlet/src/main/webapp/space-menu/extension.js
@@ -8,4 +8,4 @@ function getSpaceTitleExtensionsByType(type) {
   return extensionRegistry.loadExtensions('space-title-action-components', type);
 }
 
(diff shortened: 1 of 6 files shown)


webapp/portlet/src/main/webapp/space-menu/extension.js
```diff
@@ -2,4 +2,10 @@ function getExtensionsByType(type) {
   return extensionRegistry.loadExtensions('space-menu', type);
 }
 
-export const spaceMenuActionComponents = getExtensionsByType('action-component');
\ No newline at end of file
+export const spaceMenuActionComponents = getExtensionsByType('action-component');
+
+function getExtensionsByTypeCallButton(type) {
```
rdenarie:
```suggestion
function getSpaceTitleExtensionsByType(type) {
```
In fact, here we are in social, so, no need to refer to something which is in an addon (Call)

rdenarie (changes requested):
Please, is it possible to add a description to this PR (by following developer guides : http://developer.exoplatform.org/#id-pull-requests-best-practices) 
Why this change, possible solutions, impacts...

This will help to understand this PR. 
Thanks

pnedonosko (dismissed):

pnedonosko:
> Please, is it possible to add a description to this PR (by following developer guides : http://developer.exoplatform.org/#id-pull-requests-best-practices)
> Why this change, possible solutions, impacts...
> 
> This will help to understand this PR.
> Thanks

We have updated the PR name and description.

rdenarie (approved):
Cleaning point are catched in another US.
Ok for me

webapp/portlet/src/main/webapp/space-menu/extension.js
```diff
@@ -2,4 +2,10 @@ function getExtensionsByType(type) {
...
 
-export const spaceMenuActionComponents = getExtensionsByType('action-component');
\ No newline at end of file
+export const spaceMenuActionComponents = getExtensionsByType('action-component');
+
+function getSpaceTitleExtensionsByType(type) {
+  return extensionRegistry.loadExtensions('space-title-action-components', type);
+}
+
+export const spaceTitleActionComponents = getExtensionsByTypeCallButton('action-component');
```
rdenarie: you need to change the function name here aldo

rdenarie (changes requested):

webapp/portlet/src/main/webapp/space-menu/extension.js
pnedonosko: Already

rdenarie (approved):

pnedonosko (approved):

webapp/portlet/src/main/webapp/space-menu/main.js
```diff
@@ -36,4 +39,13 @@ export function init(settings) {
     template: `<space-menu v-cacheable="{cacheId: '${cacheId}'}" id="${appId}" :navigations="navigations" :selected-navigation-uri="selectedNavigationUri" />`,
     vuetify,
   }).$mount(appElement);
+
+  new Vue({
+    mounted() {
```
boubaker, line 44: I didn't get it, could you please explain why this change ? and what does it affect, especially why do we need to define anew app in the same main.js file ? (This is the first I encountered such use case in fact)
@pnedonosko

pnedonosko: Thank you @boubaker for looking at the code. But please clarify the question. It's not clear what does mean the _change_ 

> could you please explain why this change ?

As for _main.js_, it's part of a place we found good for adding an extension point to deal with Social space title. Its actual content in `bannerLogo.jsp`. And two instances of Vue, what is wrong with this? Can you tell how you would do this better?

boubaker: I guess, the question is clear. But it seems that we lack of best practices here.
In fact the description didn't described well, why change the existing code, is it a bug, is it an improvement. If it's an improvement for what and why is it the best solution. Here are some guidelines  : http://developer.exoplatform.org/#id-commits-messages-best-practices
After that, as any developer looking at PRs and who can comment/approve/disapprove code (not only TL of FT BTW ;-) ), I personally developed a lot of Vue applications and the solution I found here (of what problem?) seems not coherent to what we do in product usually. So my question is simple, why this change and to do what ?

PR #451, opened 2020-12-08
MAINT-37594 : Clear cache for portlet getting started when a user close this portlet and fix css

Rdidmohamedelhadi:
when user closed portlet `getting started`, the portlet is always displayed because the portlet is saved in the cache so the solution is to delete the cache when the user close the portlet.

### DIFF
--- a/webapp/portlet/src/main/webapp/getting-started/components/ExoGettingStarted.vue
+++ b/webapp/portlet/src/main/webapp/getting-started/components/ExoGettingStarted.vue
@@ -110,16 +110,10 @@ export default {
       });
     },
     clearCache(){
-      caches.keys().then(function(cacheNames) {
-        return Promise.all(
-          cacheNames.filter(function(cacheName) {
-            return cacheName === 'portal-pwa-resources-dom';
-          }).map((cacheName) => {
-            caches.open(cacheName).then(function(cache) {
-              return cache.delete('/dom-cache?id=GettingStartedPortlet');
-            });
-          })
-        );
+      caches.open('portal-pwa-resources-dom').then(function(cache) {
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/getting-started/components/ExoGettingStarted.vue
```diff
@@ -105,6 +108,19 @@ export default {
           this.showGettingStrated = false;
         }
       });
+    },
+    clearCache(){
```
ahamdi, line 112: No need to iterate over all c aches, you can get the on,e you need and delete a specific entry inside .
check the example in  https://developer.mozilla.org/en-US/docs/Web/API/Cache/delete
```
caches.open('portal-pwa-resources-dom').then(function(cache) {
  cache.delete('/dom-cache?id=GettingStartedPortlet').then(function(response) {
    console.debug('cache deleted');
  });
})```

ahamdi (changes requested):

webapp/portlet/src/main/webapp/getting-started/components/ExoGettingStarted.vue
ahamdi: And provide a description for the Pull Request

Rdidmohamedelhadi: ok

ahamdi: @Rdidmohamedelhadi I understand that people who will click on the close button will have the widget disappear, what about people who had already clicked on the close button ?

ahamdi (approved):
LGTM

PR #453, opened 2020-12-09
MAINT-37615: Allow the space manager to enable or disable the chat of a space.

mdiha:
I added the Chat applications to the returned space applications in order to check whether it is installed or not.

### DIFF
--- a/webapp/portlet/src/main/webapp/common/js/SpaceService.js
+++ b/webapp/portlet/src/main/webapp/common/js/SpaceService.js
@@ -531,32 +531,3 @@ export function ignoreSuggestion(item) {
     }
   });
 }
-
-export function updateRoomEnabled(userSettings, spaceId, enabled) {
-  const data = {
-    user: userSettings.username,
-    spaceId: spaceId,
-    enabled: enabled,
-  };
-
-  return fetch('/chatServer/updateRoomEnabled', {
-    headers: {
-      'Authorization': `Bearer ${userSettings.token}`,
-      'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8'
-    },
-    method: 'post',
-    body: decodeURI($.param(data))
-  });
-}
-
-export function isRoomEnabled(userSettings, spaceId) {
-  return fetch(`/chatServer/isRoomEnabled?user=${userSettings.username}&spaceId=${spaceId}`, {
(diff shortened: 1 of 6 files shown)


webapp/portlet/src/main/webapp/common/js/SpaceService.js
```diff
@@ -531,3 +531,32 @@ export function ignoreSuggestion(item) {
...
 }
+
+export function updateRoomEnabled(userSettings, spaceId, enabled) {
+  const data = {
+    user: userSettings.username,
+    spaceId: spaceId,
+    enabled: enabled,
+  };
+
+  return fetch('/chatServer/updateRoomEnabled', {
```
ahamdi:
```suggestion
  return fetch('/chatServer/enableRoom', {
```

ahamdi (changes requested):
We need to make it generic and we should not add code related to an addon (Chat) inside social module.

webapp/portlet/src/main/webapp/spaces-list/components/ExoSpaceCardFront.vue
```diff
@@ -237,6 +237,7 @@ export default {
     confirmMessage: '',
     okMethod: null,
     displaySecondButton: false,
+    isChatEnabled: false,
```
ahamdi: We should manage Chat button in the extension.js file, and not in the space card

ahamdi (changes requested):

webapp/portlet/src/main/webapp/spaces-list/components/ExoSpaceCardFront.vue
mdiha: The extensions are managed in a computed property which filters on all extension to check if they are enabled. In the extension.js i added an async method to the chat extension's enabled function to check if a space's chat is enabled. Since javascript's array.filter doesn't support async functions its not possible to do so.

ahamdi: Please find a way to solve it.

mdiha: Done

ahamdi (approved):

PR #467, opened 2020-12-23
TASK-38016: Enhance suggester service to allow restrict assigning to project participants

hassenelkamel-exo:
- To allow retrieving the list of participants of project , we add a specific url to get participants from task addons;

### DIFF
--- a/webapp/portlet/src/main/webapp/common/js/SuggesterService.js
+++ b/webapp/portlet/src/main/webapp/common/js/SuggesterService.js
@@ -41,13 +41,12 @@ function searchUsers(filter, items, typeOfRelation, searchOptions) {
 
   let params = null;
   let url = null;
-  if (searchOptions && !searchOptions.projectId) {
+  if (searchOptions && !searchOptions.searchUrl) {
     Object.assign(options, searchOptions);
     params = $.param(options);
     url = '/portal/rest/social/people/suggest.json?'.concat(params);
   } else {
-    params = searchOptions.projectId;
-    url = '/portal/rest/projects/projectParticipants/'.concat(params);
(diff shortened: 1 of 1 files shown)


ahamdi (changes requested):
This is a common component, it should be independent from applications usingit. Please pass the full search URL including the project ID

hassenelkamel-exo:
> This is a common component, it should be independent from applications usingit. Please pass the full search URL including the project ID

webapp/portlet/src/main/webapp/common/js/SuggesterService.js
```diff
@@ -38,11 +38,18 @@ function searchUsers(filter, items, typeOfRelation, searchOptions) {
...
+
+  let params = null;
+  let url = null;
+  if (searchOptions && !searchOptions.searchUrl) {
     Object.assign(options, searchOptions);
+    params = $.param(options);
+    url = '/portal/rest/social/people/suggest.json?'.concat(params);
+  } else {
+    url = searchOptions.searchUrl.concat(filter);
   }
```
ahamdi, line 50: Could we let it have other olptions like this ? WDYT ?

```suggestion
  if (searchOptions) {
    if(searchOptions.searchUrl) {
      url = searchOptions.searchUrl;
      delete searchOptions.searchUrl;
    } else {
      url = '/portal/rest/social/people/suggest.json?';
    }
    Object.assign(options, searchOptions);
    params = $.param(options);
    url = url.concat(params);
  }
```

ahamdi (changes requested):

ahamdi (approved):
LGTM

PR #473, opened 2020-12-28
Fix-MAINT-37615: Make sure to display the enable/disable sapce chat setting.

mdiha:
At space setting's instance Vue creation the chat Vue instance is not already created, i made sure to dispatch the event of chat-external-updated after the creation of chat's Vue instance.

### DIFF
--- a/webapp/portlet/src/main/webapp/space-settings/components/SpaceSettings.vue
+++ b/webapp/portlet/src/main/webapp/space-settings/components/SpaceSettings.vue
@@ -29,6 +29,11 @@ export default {
     document.addEventListener('hideSettingsApps', () => this.displaySpaceExternalSettings = false);
     document.addEventListener('showSettingsApps', () => this.displaySpaceExternalSettings = true);
   },
+  destroyed() {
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/space-settings/components/SpaceSettings.vue
```diff
@@ -29,6 +29,11 @@ export default {
     document.addEventListener('hideSettingsApps', () => this.displaySpaceExternalSettings = false);
     document.addEventListener('showSettingsApps', () => this.displaySpaceExternalSettings = true);
   },
+  destroyed() {
+    document.removeEventListener('chat-external-updated');
```
ahamdi, line 33: Is this needed ? we do not want to add any specific event here

ahamdi (changes requested):

webapp/portlet/src/main/webapp/space-settings/components/SpaceSettings.vue
mdiha: Not really i will close this PR.

PR #477, opened 2021-01-05
Task-38526: Add external label for external activity commenter

ahmadmido:


component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
```diff
@@ -452,4 +462,35 @@ public static String getBaseURLSocialUserRest() {
...
+      return null;
+    }
+  }
+
+  /**
+   * Gets the resource bundle label.
+   *
+   * @return the label
+   */
+  public static String getResourceBundleLabel(Locale locale, String label) {
```
azayati, line 492: I think that there is an Utils method added for that ?

extension/war/src/main/webapp/groovy/social/webui/activity/commons/UIActivityCommentBox.gtmpl
```diff
@@ -12,6 +12,7 @@
   def labelComment = _ctx.appRes("UIActivity.label.Comment");
   def labelCancel = _ctx.appRes("UIActivity.label.Cancel");
   def labelUpdate = _ctx.appRes("UIActivity.label.Update");
+  def external = _ctx.appRes("UIActivity.label.External");
```
azayati, line 15: The () should be removed from this label, to be added directly when displaying

azayati (changes requested):
See review

component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
ahmadmido: It can't be used in core component !

azayati (approved):
LGTM

PR #481, opened 2021-01-07
MAINT-39075 : Newly created users avatar isn't displayed in people when filtering

Rdidmohamedelhadi:
**Problem :** when updating avatar of user `Elasticsearch`  don't  indexing avatar so the **solution** is when update avatar we re-index it .

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/profile/UserProfileComparator.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/profile/UserProfileComparator.java
@@ -49,7 +49,7 @@ public class UserProfileComparator {
     if (oldValue == null) {
       return newValue != null
           && (((newValue instanceof String) && StringUtils.isNotBlank(newValue.toString()))
-              || ((newValue instanceof List) && !((List<?>) newValue).isEmpty()));
+              || ((newValue instanceof List) && !((List<?>) newValue).isEmpty())) || profileToUpdate.getProperty(Profile.AVATAR) != null;
     } else if (oldValue instanceof String) {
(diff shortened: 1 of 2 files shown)


component/core/src/main/java/org/exoplatform/social/core/profile/UserProfileComparator.java
```diff
@@ -46,10 +47,19 @@ private static boolean hasChanged(Profile profileToUpdate, Profile existingProfi
     Object newValue = profileToUpdate.getProperty(key);
     Object oldValue = existingProfile.getProperty(key);
 
+    //to check if the avatar is changed 
+    boolean avatarHasChanged = false;
+    if (Profile.AVATAR.equals(key)){
```
ahamdi: Why should have a specific usage for the avatar field ? couldn't it be treated like other fields ?

ahamdi (changes requested):

component/core/src/main/java/org/exoplatform/social/core/profile/UserProfileComparator.java
Rdidmohamedelhadi: I can't use `hasChanged` without modifying it because the value of in existing profile avatar is null, so my solution when I have a value (<> null) for avatar in profile to update, I can detect that the value has changed.

ahamdi: I didn't understand, I see that you added an extra check for the avatyer even though it is almost the same as other. could you at least add it to the return expression in case oldValue == null

ahamdi (approved):
LGTM

PR #482, opened 2021-01-07
Task 39157 : When an external user in invited, there is an error disp…

rdenarie:
…layed with psql database

Prior to this change, when creating an external invitation, the line is not added in the corresponding table due to error in log.
With PSQL db, the sequence for this table is missing
This commit add the createSequence needed in liquibase


ahmadmido (dismissed):
LGTM

ahmadmido (approved):
LGTM

PR #488, opened 2021-01-12
MAINT-36333 : fix empty email notification from activity with attached files

hakermi:


hbenali (changes requested):
This integration is not allowed. ECMS is an addon. Please consider meeds packaging.

ahamdi:
Same as @hbenali said, ECMS should not be added as a dependency to Social

PR #489, opened 2021-01-12
Move Dlp quarantine portlet from ECMS

AzmiTouil:


rdenarie (changes requested):
Remove i18n files which are not en or fr

rdenarie (changes requested):
In addition you should update the translation properties files

rdenarie (approved):

PR #493, opened 2021-01-13
Task-39254: Display detected positive items on the quarantine page

AzmiTouil:


webapp/portlet/src/main/webapp/dlp-quarantine/dlpAdministrationServices.js
```diff
@@ -25,3 +25,16 @@ export function isDlpFeatureActive() {
     }
   });
 }
+
+export function getDlpPositiveItems() {
+  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/dlp/items`, {
```
rdenarie: should we use offset and limit to prevent to get ALL items ?

rdenarie (changes requested):

webapp/portlet/src/main/webapp/dlp-quarantine/dlpAdministrationServices.js
AzmiTouil: Updated !

rdenarie (dismissed):

rdenarie (dismissed):

rdenarie (approved):

PR #500, opened 2021-01-21
MAINT-39795: fix and handle a condition that prevents the succeed of  adding an application in the space

hakermi:
This PR should handle an Exception : **Tree already in the map** that raises when the application wasn't completely added to the space and the user attempts to add it again.

 and i have reproduced it only once


mkrout (approved):
LGTM

PR #510, opened 2021-01-25
TASK-39855 Improve user validators to include email check

boubaker:
Identity old enpoint is presenting a security issue and should be deleted, especially that we have adequate endpoints to update user and space identites separately.
In addition the User email redundancy check has to be done systematically when updating user fields.
This fix will update endpoint and enhance unit tests to make sure to not give possibility to a user to use an email of another existing user.


ahamdi (approved):

boubaker:
Merged using https://github.com/Meeds-io/social/pull/507

PR #519, opened 2021-01-28
MAINT-37615: Make sure to display the enable/disable space chat setting section.

mdiha:


webapp/portlet/src/main/webapp/space-settings/components/SpaceSettings.vue
```diff
@@ -19,7 +19,7 @@ export default {
   }),
   created() {
     // get external components
-    document.addEventListener('chat-external-updated', () => {
+    document.addEventListener('component-SpaceSettings-external-component-space-chat-setting-updated', () => {
```
ahamdi: This is a very long and specific event, we need to haver a generic event name like 
```suggestion
    document.addEventListener('external-space-settings-component-updated', () => {
```

ahamdi (changes requested):

webapp/portlet/src/main/webapp/space-settings/components/SpaceSettings.vue
mdiha: This is the default event which is fired from the extensionRegistry module : https://github.com/Meeds-io/gatein-portal/blob/254fd4f7dad51828f9ff5cade42a1594458166f0/web/eXoResources/src/main/webapp/javascript/eXo/core/extensionRegistry.js#L59 
So i will change the app name and componentName in Chat-Application module

ahamdi (approved):

webapp/portlet/src/main/webapp/space-settings/components/SpaceSettings.vue
ahamdi: Ok clear

PR #520, opened 2021-01-28
Secure accessing to users and spaces avatars and banners

boubaker:
When accessing an user avatar anonymously, there weren't a check made to ensure that user avatar isn't accessible only for permitted users. In order to allow accessing user avatars and especially from email notifications, a token is attached to user avatar URL that will allow access anonymously to user avatar.
( Same strategy was adopted for user banner )
For spaces, before this fix, open spaces avatars weren't accessible anonymously at all. So even on email, the avatar isn't displayed anymore. Thus the same strategy of token has been adopted for non-Hidden Spaces


mkrout (approved):
LGTM

PR #536, opened 2021-02-07
MAINT-40475: Prevent suggesting disabled users

hbenali:
Users connections suggestion includes disabled users (OK for deleted)


ahamdi (changes requested):
Could you add a UT for this fix ?

ahamdi (approved):
LGTM

PR #548, opened 2021-02-11
Task-40607 : UI feedbacks about the quarantine page

ahmadmido:


webapp/portlet/src/main/webapp/dlp-quarantine/components/DlpEditPermissions.vue
```diff
@@ -205,6 +211,12 @@ export default {
       saveDlpPermissions(this.selectedItems.map(item => item.id).join());
       this.$refs.EditPermissionsDrawer.close();
       this.permissions = this.selectedItems;
+      this.searchResults = '';
+    },
+    CleanInput() {
```
rdenarie:
```suggestion
    cleanInput() {
```

webapp/portlet/src/main/webapp/dlp-quarantine/dlpAdministrationServices.js
```diff
@@ -100,9 +100,9 @@ export function checkIsAdminMemberGroup() {
   });
 }
 
-export async function getPermissionsData(query) {
+export async function getPermissionsData(query, offset, limit) {
   try {
-    const response = await fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/v1/groups?q=${query}`, {
+    const response = await fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/dlp/items/groups?q=${query}&offset=${offset || 0}&limit=${limit|| 0}`, {
```
rdenarie: Why it was not possible to use default API ? What are the limitation which lead to create and use a custom rest service ?

rdenarie (changes requested):

webapp/portlet/src/main/webapp/dlp-quarantine/dlpAdministrationServices.js
ahmadmido: To suggest only groups in the permissions drawer, the default rest services suggest spaces and groups.

rdenarie: As explained in the task, this task is paused, because this part will be reworked and simplified. So please : do not add a new rest function in dlp/items/groups to not add deadly code, and keep using /v1/groups for this moment.
Thakns

rdenarie (approved):

PR #553, opened 2021-02-15
MAINT-7218: Searching for newly created users in order to invite them to a space is failing

Rdidmohamedelhadi:
This condition `if (this.previousSearchTerm === this.searchTerm)` is always invalid at the first letter typed, since `this.previousSearchTerm` is initialized to `null` , to fix this problem we should skip this first typed letter.

### DIFF
--- a/webapp/portlet/src/main/webapp/common/components/ExoIdentitySuggester.vue
+++ b/webapp/portlet/src/main/webapp/common/components/ExoIdentitySuggester.vue
@@ -184,7 +184,7 @@ export default {
     searchTerm(value) {
       if (value && value.length) {
         window.setTimeout(() => {
-          if (this.previousSearchTerm === null || this.previousSearchTerm === this.searchTerm) {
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/common/components/ExoIdentitySuggester.vue
```diff
@@ -184,7 +184,7 @@ export default {
     searchTerm(value) {
       if (value && value.length) {
         window.setTimeout(() => {
-          if (this.previousSearchTerm === this.searchTerm) {
+          if (this.previousSearchTerm === null || this.previousSearchTerm === this.searchTerm) {
```
ahamdi:
```suggestion
          if (! this.previousSearchTerm || this.previousSearchTerm === this.searchTerm) {
```

Rdidmohamedelhadi: Done !!

ahamdi (approved):

PR #563, opened 2021-02-16
Task-40653 : People & spaces connectors for external users and internal

AzmiTouil:
improvement :

1. As internal user, When I go to the unified search

Then I don't display external users (whether I am connected or not to the external user I don't see him)

2. As PLF external user, When I go to the unified search

Then I cannot search by people and spaces  (I don't see the tag)

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/search/impl/SearchServiceImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/search/impl/SearchServiceImpl.java
@@ -20,6 +20,8 @@ import java.util.*;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
+import org.exoplatform.commons.utils.CommonsUtils;
+import org.exoplatform.portal.config.UserACL;
 import org.picocontainer.Startable;
 
 import org.exoplatform.commons.api.settings.SettingService;
@@ -38,6 +40,8 @@ public class SearchServiceImpl implements SearchService, Startable {
   private static final Scope   SEARCH_CONNECTORS_SCOPE        = Scope.APPLICATION.id("connectors");
 
   private static final String  SEARCH_CONNECTORS_STATUS_PARAM = "enabledSearchConnectors";
+
+  private static final String  PLATFORM_EXTERNALS_GROUP  = "/platform/externals";
 
   private SettingService       settingService;
 
@@ -76,6 +80,9 @@ public class SearchServiceImpl implements SearchService, Startable {
 
   @Override
   public Set<SearchConnector> getEnabledConnectors() {
+    UserACL userACL = CommonsUtils.getService(UserACL.class);
+    boolean isExternalUser = userACL.isUserInGroup(PLATFORM_EXTERNALS_GROUP);
+    connectors.stream().filter(connector -> connector.getName().equals("space") || connector.getName().equals("people")).forEach(connector -> connector.setEnabled(!isExternalUser));
     return Collections.unmodifiableSet(connectors.stream()
                                                  .filter(SearchConnector::isEnabled)
                                                  .map(SearchConnector::clone)
--- a/component/core/src/main/java/org/exoplatform/social/core/search/impl/SearchServiceImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/search/impl/SearchServiceImpl.java
@@ -82,9 +82,8 @@ public class SearchServiceImpl implements SearchService, Startable {
   public Set<SearchConnector> getEnabledConnectors() {
     UserACL userACL = CommonsUtils.getService(UserACL.class);
     boolean isExternalUser = userACL.isUserInGroup(PLATFORM_EXTERNALS_GROUP);
-    connectors.stream().filter(connector -> connector.getName().equals("space") || connector.getName().equals("people")).forEach(connector -> connector.setEnabled(!isExternalUser));
     return Collections.unmodifiableSet(connectors.stream()
-                                                 .filter(SearchConnector::isEnabled)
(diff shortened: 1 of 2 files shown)


component/core/src/main/java/org/exoplatform/social/core/search/impl/SearchServiceImpl.java
```diff
@@ -76,6 +80,9 @@ public void addConnector(SearchConnectorPlugin connectorPlugin) {
 
   @Override
   public Set<SearchConnector> getEnabledConnectors() {
+    UserACL userACL = CommonsUtils.getService(UserACL.class);
+    boolean isExternalUser = userACL.isUserInGroup(PLATFORM_EXTERNALS_GROUP);
+    connectors.stream().filter(connector -> connector.getName().equals("space") || connector.getName().equals("people")).forEach(connector -> connector.setEnabled(!isExternalUser));
```
rdenarie: Can you ensure this : 
1) an external user goes to unified search : he sees all connectors excepts people and space
2) log with another user, which is internal. Does he see the connectors spaces and people ? 

As developed, I'm afraid that the setEnabled can disabled connectors for all users.

extension/war/src/main/webapp/WEB-INF/conf/social-extension/social/search-configuration.xml
```diff
@@ -60,7 +60,7 @@
               <string>people</string>
             </field>
             <field name="uri">
-              <string><![CDATA[/portal/rest/v1/social/users?q={keyword}&limit={limit}&expand=all,spacesCount,relationshipStatus,connectionsCount]]></string>
```
rdenarie, line 63: As I understand this is the request launched with the search app. 
you add the flag excludeExternal=true, so the request will be modified. 
But, what happens if the internal user get the request in chrome dev tool, and use excludeExternal=false ? or remove this attribute ?

rdenarie (changes requested):

component/core/src/main/java/org/exoplatform/social/core/search/impl/SearchServiceImpl.java
AzmiTouil: Yes it's , since it is not stored in the **settingService** and this method is called each time the search portlet is instantiated.

extension/war/src/main/webapp/WEB-INF/conf/social-extension/social/search-configuration.xml
AzmiTouil: it gets a full search result (with external users), I didn't change the [getUsers()](https://github.com/Meeds-io/social/blob/c7faaddf21e50eece9e0d350a986fc48d5a57f5c/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java#L262) rest method, because it's a generic method and used in other places.

component/core/src/main/java/org/exoplatform/social/core/search/impl/SearchServiceImpl.java
rdenarie: But you modify the list connectors at runtine, no ? So the next user which will call getEnabledConnectors will get a list with some connetors disabeld

extension/war/src/main/webapp/WEB-INF/conf/social-extension/social/search-configuration.xml
rdenarie: So, if a internal user get the rest service without excludeExternal, he will be able to read all users, including external. 
This should be ask to Sameh and Samuel, because I'm not sure that it is the wanted behaviour

component/core/src/main/java/org/exoplatform/social/core/search/impl/SearchServiceImpl.java
AzmiTouil: it depends on user status (internal or external) `connector.setEnabled(!isExternalUser)`

extension/war/src/main/webapp/WEB-INF/conf/social-extension/social/search-configuration.xml
rdenarie: In the web service, if excludeExternal==false and user is not admin, then the service returns 403, so it is ok with this modification

component/core/src/main/java/org/exoplatform/social/core/search/impl/SearchServiceImpl.java
rdenarie: In fact, the connectors list is an object present in SearchServiceImpl. So if you call setEnabled(false) for one connector of this list, you will set it disable for all users, and this is KO.

component/core/src/main/java/org/exoplatform/social/core/search/impl/SearchServiceImpl.java
```diff
@@ -76,6 +80,9 @@ public void addConnector(SearchConnectorPlugin connectorPlugin) {
...
     return Collections.unmodifiableSet(connectors.stream()
                                                  .filter(SearchConnector::isEnabled)
```
rdenarie:
```suggestion
                                                 .filter(c -> c.isEnabled && (
                                                    (c.getName().equals("people") || c.getName().equals("space")) && !isExternalUser)))
```
We filter connectors : it should be enable AND if connector name is people or space, user should not be external

rdenarie (changes requested):

component/core/src/main/java/org/exoplatform/social/core/search/impl/SearchServiceImpl.java
AzmiTouil: Updated!

rdenarie (approved):

PR #578, opened 2021-02-22
MAINT-27712: Make sure to be able to mention users when editing a post.

mzorai:
The problem comes from the already initialized CKEditor instance, i made sure to reinitialize it when editing a post.


ahamdi (changes requested):
Do not merge until we fix the inifinite loading isue when editing a post

ahamdi (approved):

PR #582, opened 2021-02-22
Task-39984: Improve keyword set for DLP

AzmiTouil:
Add the ability to add more than one word (at least two words) in the DLP keyword edition.


azayati (approved):
LGTM

PR #591, opened 2021-02-24
41045 : clear activity message after posting it

ahamdi:
Currently, after posting a message and then trying to navigate to another page, a popup is displayed asking if we need to keep changes. This popup comes from a CKeditor plugin that ensures users does not leave the page when changes are still in the editor.
This fix will reset the content of the activity composer editor once the activity is posted successfully.


boubaker:
Currently, when closing the drawer without posting, the alert message gets displayed as well (not same behavior as before)
I propose to cherry pick the commit https://github.com/Meeds-io/social/commit/3a33d598bb7b437d8e7be90ac1565f535296ee73 as well here to make it exactly the same behavior as before, WDYT ?

boubaker (approved):

PR #599, opened 2021-02-25
Task-40906: Remove permissions management area from the quarantine page

azayati:


hbenali (commented):
No need each time to remove non-English entries from resources bundle. Crowdin will do it for you. Just make changes on the default one which is in english.

rdenarie (approved):

PR #603, opened 2021-03-01
Task-40910: Fix enable/disable the DLP feature

ahmadmido:
- Before this fix, only members of admins groups show the switch button on DLP feature.
-This fix will assure that all members on DLP group can show the switch button.


azayati (approved):
LGTM

PR #609, opened 2021-03-04
MAINT-36023: Make sure that applications names and descriptions in spaces settings are translated.

mdiha:
Space applications names and descriptions are not translated in the space settings menu, i made sure to internationalize (I18n) these two properties.


ahamdi:
Please provide a description

mdiha:
> Please provide a description

Description added

ahamdi (changes requested):
We need to use the same Bundles for applications already defined. Each application should bring its label from its resource bundles itself and we should not put references to ecms/ chat etc ... here.

mdiha:
> We need to use the same Bundles for applications already defined. Each application should bring its label from its resource bundles itself and we should not put references to ecms/ chat etc ... here.

Okay will add a PR for each application

mdiha:
Added PRs:
https://github.com/exoplatform/task/pull/355
https://github.com/exoplatform/wiki/pull/270
https://github.com/exoplatform/ecms/pull/1180
https://github.com/exoplatform/agenda/pull/273

ahamdi (approved):

PR #610, opened 2021-03-04
Task-21244: platform/dlp members should not be able to enable/disable dlp notifications

ahmadmido:
- Before this fix, all users and members of platform/dlp can edit notifications settings of dlp feature.
- This fix will make sure that no one can edit dlp notifications it will be enabled by default.


azayati (approved):
LGTM

PR #612, opened 2021-03-08
Task-41341 : Change userManegement page style

ahmadmido:
This PR will change the style of usersManegement : 
- Separate actions delete - edit - role.
- Change the icon on the role column.
- The delete icons are available only for superusers and are red.
- Restore the display of the user's status: Reuse the switch buttons for this column and remove the switch button enable/disable from the edit user drawer.


webapp/portlet/src/main/webapp/common/js/UserService.js
```diff
@@ -11,6 +11,19 @@ export function getUser(username, expand) {
   });
 }
 
+export function checkIsSuperUser() {
+  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/v1/users/checkIsSuperUser`, {
```
rdenarie:
```suggestion
export function isSuperUser() {
  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/v1/users/isSuperUser`, {
```

rdenarie (changes requested):

rdenarie (dismissed):

rdenarie (dismissed):

rdenarie (approved):

PR #615, opened 2021-03-10
Task-41306 : Add source column on UserManegement

ahmadmido:
This PR will add a new column on UserManegmenet table -> Source.
This column will check if a user is an internal user or a synchronized user, if is internal we display an icon with eXo logo and if's synchronized LDAP/AD we display an icon with a tag SYNC.


component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
```diff
@@ -192,6 +192,7 @@ public static ProfileEntity buildEntityProfile(Profile profile, String restPath,
       User user = organizationService.getUserHandler().findUserByName(userEntity.getUsername());
       if (user != null) {
         userEntity.setIsInternal(user.isInternalStore());
+        userEntity.setCreatedDate(String.valueOf(user.getCreatedDate().getTime()));
```
rdenarie: the date we have to display as tooltip is the date of last sync for the user, not the created date.
Imagine that the user is modified in AD, then synchronised, the sync date is not the same as the created date.
As explain in the tech spec, we should add a listener on event linked to synchronisation, and store/update the sync date in the user profile

webapp/portlet/src/main/resources/locale/portlet/Portlets_en.properties
```diff
@@ -335,6 +335,8 @@ UsersManagement.lastConnexion.neverEnrolled=Never enrolled
 UsersManagement.password=Password
 UsersManagement.confirmPassword=Confirm password
 UsersManagement.status=Status
+UsersManagement.source=Source
+UsersManagement.source.syncronized=Syncronized on {0}
```
rdenarie:
```suggestion
UsersManagement.source.syncronized=Synchronized on {0}
```

webapp/portlet/src/main/webapp/idm-users-management/components/UsersManagementList.vue
```diff
@@ -52,6 +52,25 @@
...
+          <v-btn
+            primary
+            icon
+            text>
+            <i class="uiIconSoupCan"><span class="internalIconClass">eXo</span></i>
+          </v-btn>
+        </div>
+        <div v-else>
+          <v-btn
+            :title="suncronizedTitle(item.createdDate)"
```
rdenarie: you should use date-format vue component to display this date.

rdenarie (changes requested):

rdenarie (dismissed):
When you want to add a parameter in an existing method of a service widely used, it is preferable to do this : 
1) keep the existing method without your new parameter
2) add a new method with the new parameter and containing the logic of the existing method
3) modifty the existing method by calling the new method, with a default value for the new parameter (in our case : enabled=true)

So that you not have to modify everywhere in the codebase, all method without the new parameter will continue to work, and call the method with the new parameter default value

With a modification like that, your PR can be greatly simplified

rdenarie (dismissed):

rdenarie (dismissed):

rdenarie (approved):

PR #618, opened 2021-03-11
Task-41416_4117: Implement MalwareDetection feature

azayati:
This PR:
- Implements Malware detection rest service which allows to process "files" or "jcr nodes" detected as malware ,the process is right now only sending a notification to the last updater of the infected file. 
- Implements FileMalwareDetectionServiceConnector for the "files" detection case, for example after uploading a wiki attachment, a FS file example is generated into /gatein/data/files/2/2/d/b/1/2/3/3/22db1233ceb84833ac44cc36f630fb1c of the server,  the last part of the file path is the file checksum, the connector will detect this checksum, get the corresponding files using FileService then send the notification.

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/service/malwareDetection/MalwareDetectionService.java
+++ b/component/service/src/main/java/org/exoplatform/social/service/malwareDetection/MalwareDetectionService.java
@@ -1,0 +1,71 @@
+/**
+ * 
+ */
+package org.exoplatform.social.service.malwareDetection;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.exoplatform.services.log.ExoLogger;
+import org.exoplatform.services.log.Log;
+import org.exoplatform.social.service.malwareDetection.connector.MalwareDetectionServiceConnector;
+
+/**
+ * Is extended by all MalwareDetectionService connectors, and allows to build configuration needed by a list of connectors that is used for Malware detection.
+ * 
+ */
+public class MalwareDetectionService {
+  
+  private Map<String, MalwareDetectionServiceConnector> connectors = new HashMap<String, MalwareDetectionServiceConnector>();
+  private static final Log LOGGER = ExoLogger.getExoLogger(MalwareDetectionService.class);
+  
+  /**
+   * Add Malware detection Connector to the service
+   * @param malwareDetectionServiceConnector the Malware detection Connector to add
+   * @LevelAPI Experimental
+   */
+  public void addConnector (MalwareDetectionServiceConnector malwareDetectionServiceConnector) {
+    addConnector(malwareDetectionServiceConnector, false);
+  }
+  
+  /**
+   * Add malware detection Connector to the service
+   * @param malwareDetectionServiceConnector the malware detection connector to add
+   * @param override equal true if we can override an existing connector, false otherwise
+   * @LevelAPI Experimental
+   */
+  public void addConnector (MalwareDetectionServiceConnector malwareDetectionServiceConnector, Boolean override) {
+    if (connectors.containsKey(malwareDetectionServiceConnector.getType()) && override.equals(false)) {
+      LOGGER.error("Impossible to add connector {}. A connector with the same name has already been registered.", malwareDetectionServiceConnector.getType());
+    } else {
+      connectors.put(malwareDetectionServiceConnector.getType(), malwareDetectionServiceConnector);
+      LOGGER.info("A Malware detection Connector has been added: {}", malwareDetectionServiceConnector.getType());
+    }
+  }
+  
+  /**
+   * Gets all current connectors
+   * @return Connectors
+   * @LevelAPI Experimental
+   */
+  public Map<String, MalwareDetectionServiceConnector> getConnectors() {
+    return connectors;
+  }
+  
+  public void processInfectedItem(String infectedItemPath) {
+    if (infectedItemPath.contains(MalwareDetectionServiceConnector.FILES_CONNECTOR)) {
+      MalwareDetectionServiceConnector fileMalwareDetectionServiceConnector = (MalwareDetectionServiceConnector) getConnectors().get(MalwareDetectionServiceConnector.FILES_CONNECTOR);
+      if(fileMalwareDetectionServiceConnector != null && fileMalwareDetectionServiceConnector.isEnable()) {
+        fileMalwareDetectionServiceConnector.sendInfectedItemNotification(infectedItemPath);
+        fileMalwareDetectionServiceConnector.cleanInfectedItem(infectedItemPath);
+      }
+    }
+    else if (infectedItemPath.contains(MalwareDetectionServiceConnector.JCR_CONNECTOR)) {
+      MalwareDetectionServiceConnector jcrMalwareDetectionServiceConnector = (MalwareDetectionServiceConnector) getConnectors().get(MalwareDetectionServiceConnector.JCR_CONNECTOR);
+      if(jcrMalwareDetectionServiceConnector != null && jcrMalwareDetectionServiceConnector.isEnable()) {
+        jcrMalwareDetectionServiceConnector.sendInfectedItemNotification(infectedItemPath);
+        jcrMalwareDetectionServiceConnector.cleanInfectedItem(infectedItemPath);
+      }
+    }
+  }
+}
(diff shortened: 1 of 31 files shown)


component/service/src/main/java/org/exoplatform/social/service/malwareDetection/connector/FileMalwareDetectionServiceConnector.java
```diff
@@ -0,0 +1,68 @@
...
+    String fileChecksum = getInfectedFileChecksum(infectedFilePath);
+    try {
+      List<FileItem> infectedFileItems = fileService.getFilesByChecksum(fileChecksum);
+      for (FileItem infectedFileItem : infectedFileItems) {
+        if (infectedFileItem != null && infectedFileItem.getFileInfo() != null && infectedFileItem.getFileInfo().getId() != null) {
+          fileService.deleteFile(infectedFileItem.getFileInfo().getId());
+        }
+      }
+    } catch (Exception e) {
+      LOGGER.error("Error when trying to clean the infected item", e);
```
plamarque: shouldn't we log some identifier of the infected item to help troubleshoot ?

component/service/src/main/java/org/exoplatform/social/service/malwareDetection/connector/FileMalwareDetectionServiceConnector.java
```diff
@@ -0,0 +1,68 @@
...
+      for (FileItem infectedFileItem : infectedFileItems) {
+        if (infectedFileItem != null && infectedFileItem.getFileInfo() != null) {
+          Map<String, String> infectedItem = new HashMap<String, String>();
+          infectedItem.put(INFECTED_ITEM_NAME, infectedFileItem.getFileInfo().getName());
+          infectedItem.put(INFECTED_ITEM_LAST_MODIFIER, infectedFileItem.getFileInfo().getUpdater());
+          infectedItems.add(infectedItem);
+        }
+      }
+    } catch (Exception e) {
+      LOGGER.error("Error when trying to get the infected item informations", e);
```
plamarque: shouldn't we log some identifier of the infected item to help troubleshoot ?

extension/notification/src/main/resources/locale/notification/template/Notification_en.properties
```diff
@@ -278,3 +278,15 @@ UINotification.title.DlpUserRestoredItemPlugin=Item has been validated
...
 Notification.subject.DlpUserRestoredItemPlugin=Item has been validated
 Notification.intranet.message.DlpUserRestoredItemPlugin=The item named <strong>{0}</strong> has been analyzed during a quarantine phase. It has been validated and restored in his initial location.
+
+
+#############################################################################
+#                         MalwareDetectionPlugin                         #
+#############################################################################
+UINotification.label.group.MalwareDetection=Malware detection
+UINotification.label.MalwareDetectionPlugin=Malware detection
+UINotification.title.MalwareDetectionPlugin=Item is infected
```
plamarque: should we simply replace "item" by "file" or "document" ? seems more user friendly

component/service/src/main/java/org/exoplatform/social/service/malwareDetection/connector/FileMalwareDetectionServiceConnector.java
azayati: yes I will rework the logs with the good format, it is just the draft :)

extension/notification/src/main/resources/locale/notification/template/Notification_en.properties
azayati: I have used the same label of DLP, I will see with PO to put the good label

azayati: Modified @plamarque

component/service/src/main/java/org/exoplatform/social/service/malwareDetection/connector/FileMalwareDetectionServiceConnector.java
azayati: I have added one log [here](https://github.com/Meeds-io/social/pull/618/files#diff-8c188b1765721990c9d74eed2edd9399e5a54bc0e38a7488a31da8df96fcf6beR38) when detecting a file as a malware

component/service/src/main/java/org/exoplatform/social/service/malwareDetection/connector/FileMalwareDetectionServiceConnector.java
rdenarie: :+1:

component/service/src/test/java/org/exoplatform/social/service/malwareDetection/MalwareDetectionServiceTest.java
```diff
@@ -0,0 +1,27 @@
...
+  
+  private MalwareDetectionService malwareDetectionService;
+
+  public void setUp() throws Exception {
+    super.setUp();
+    malwareDetectionService = getContainer().getComponentInstanceOfType(MalwareDetectionService.class);
+  }
+
+  public void testProcessInfectedItem() {
+    String infectedItemPath = "/gatein/data/files/d/e/4/c/e/2/d/7/dedeerere673r777";
```
rdenarie, line 18: Can you make a test with more than one infected item

rdenarie (changes requested):

component/service/src/test/java/org/exoplatform/social/service/malwareDetection/MalwareDetectionServiceTest.java
azayati: Right now the test done is for the MalwareDetectionService which process only one infected item. We can test MalwareDetectionServiceRestService also in order to test the case of a list of infected items

rdenarie (approved):

component/service/src/main/java/org/exoplatform/social/service/malwareDetection/MalwareDetectionService.java
```diff
@@ -0,0 +1,89 @@
...
+    else if (infectedItemPath.contains(getJcrNodePathSeparator(infectedItemPath) + JCR_CONNECTOR + getJcrNodePathSeparator(infectedItemPath) + VALUES + getJcrNodePathSeparator(infectedItemPath))) {
+      MalwareDetectionServiceConnector jcrMalwareDetectionServiceConnector = (MalwareDetectionServiceConnector) getConnectors().get(JCR_CONNECTOR);
+      if(jcrMalwareDetectionServiceConnector != null && jcrMalwareDetectionServiceConnector.isEnable()) {
+        jcrMalwareDetectionServiceConnector.sendInfectedItemNotification(infectedItemPath);
+        jcrMalwareDetectionServiceConnector.cleanInfectedItem(infectedItemPath);
+      }
+    }
+  }
+  
+  public static String getJcrNodePathSeparator(String infectedFilePath) {
```
rdenarie:
```suggestion
  public static String getPathSeparator(String infectedFilePath) {
```
As this is not related to JCR we should not use 'JCR' in method name.

component/service/src/main/java/org/exoplatform/social/service/malwareDetection/MalwareDetectionService.java
```diff
@@ -0,0 +1,89 @@
...
+        fileMalwareDetectionServiceConnector.cleanInfectedItem(infectedItemPath);
+      }
+    }
+    else if (infectedItemPath.contains(getJcrNodePathSeparator(infectedItemPath) + JCR_CONNECTOR + getJcrNodePathSeparator(infectedItemPath) + VALUES + getJcrNodePathSeparator(infectedItemPath))) {
+      MalwareDetectionServiceConnector jcrMalwareDetectionServiceConnector = (MalwareDetectionServiceConnector) getConnectors().get(JCR_CONNECTOR);
+      if(jcrMalwareDetectionServiceConnector != null && jcrMalwareDetectionServiceConnector.isEnable()) {
+        jcrMalwareDetectionServiceConnector.sendInfectedItemNotification(infectedItemPath);
+        jcrMalwareDetectionServiceConnector.cleanInfectedItem(infectedItemPath);
+      }
+    }
```
rdenarie: getPathSeparator should be called once at the beginning of the function, and stored in a local variable, so that it will prevent to call a lot of contains on String

rdenarie: Second question : 
I think it should be preferable to do something like this : 

for connector : connectors {
   if (connector.isConcernedBy(infectedFilePath)) {
    do stuff
  }
}

So that, you have not the logic in the master MalwareDetectionService and each connector can say if he have authority to clean the infected file.

My idea is to remove all information about JCR from social module

rdenarie (changes requested):

rdenarie (approved):

PR #619, opened 2021-03-11
Task-41300: Add enrollment column in UserManagement

AzmiTouil:
This PR will add a new column in the user management that will describe the enrollment status of each user


component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -1377,17 +1378,10 @@ private String importUser(UserImportResultEntity userImportResultEntity,
         LOG.error("Failure to retrieve portal config", e);
       }
       passwordRecoveryService.sendOnboardingEmail(user, locale, url);
+      if (profileEntity != null) {
+        profileEntity.setEnrollmentDate(String.valueOf(Calendar.getInstance().getTimeInMillis()));
+      }
```
rdenarie: This should be in sendOnboardEmail method, so that, you can check if the sent is ok or KO

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -1461,7 +1455,7 @@ private void logMetrics(ExoSocialActivity activity) {
   }
 
   private void saveProfile(String username, ProfileEntity profileEntity) throws Exception {
-    Identity userIdentity = getUserIdentity(username);
+    Identity userIdentity = identityManager.getOrCreateIdentity(OrganizationIdentityProvider.NAME, username);
```
rdenarie: Can you explain why you change this line ?

webapp/portlet/src/main/webapp/idm-users-management/components/UsersManagementList.vue
```diff
@@ -324,10 +376,26 @@ export default {
...
+              user.enrollmentDetails= this.$t('UsersManagement.enrollment.alreadyConnected');
+            }
           } else if (user.external === 'true') {
             user.lastConnexion = this.$t('UsersManagement.lastConnexion.neverConnected');
+            user.enrollmentStatus = 'cannotBeEnrolled';
+            user.enrollmentDetails= this.$t('UsersManagement.enrollment.cannotBeEnrolled');
+          }else if (user.enrollmentDate != null) {
+            user.lastConnexion = this.$t('UsersManagement.lastConnexion.invitedToJoin');
+            user.enrollmentStatus = 'reInviteToJoin';
+            user.enrollmentDetails= this.$t('UsersManagement.enrollment.reInviteToJoin', {0: this.formatDate(Number(user.enrollmentDate))});
```
rdenarie, line 385: is possible to use date-format vue component to display the date ? So that you not have to format yourself the date

rdenarie (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
AzmiTouil: I can't update profile inside sendOnboardEmail method , I added a check if the sent is ok or not

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -828,6 +819,27 @@ public Response updateUserById(@Context UriInfo uriInfo,
     return EntityBuilder.getResponse(EntityBuilder.buildEntityProfile(id, uriInfo.getPath(), expand), uriInfo, RestUtils.getJsonMediaType(), Response.Status.OK);
   }
 
+  @GET
+  @Path("onBoarding/{id}")
```
plamarque: /onboard  (prefer loser case in rest endpoints)

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -828,6 +819,27 @@ public Response updateUserById(@Context UriInfo uriInfo,
...
+  @Produces(MediaType.APPLICATION_JSON)
+  @RolesAllowed("administrators")
+  @ApiOperation(value = "Send onBoarding email to a specific user", 
+                httpMethod = "GET", 
+                response = Response.class,
```
plamarque, line 828: not sure GET is the right verb semantic for an endpoint that actually changes state. I would use POST here to  follow RESTful recommends

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -828,6 +819,27 @@ public Response updateUserById(@Context UriInfo uriInfo,
...
+                notes = "This send onBoarding email to a specific user.")
+  public Response sendOnBoardingEmail(@Context UriInfo uriInfo,
+                                      @ApiParam(value = "User name", required = true) @PathParam("id") String id) throws Exception {
+    UserHandler userHandler = organizationService.getUserHandler();
+    User user = userHandler.findUserByName(id);
+    if (user == null) {
+      throw new WebApplicationException(Response.Status.BAD_REQUEST);
+    }
+    String uri = uriInfo.getBaseUri().toString().substring(0, uriInfo.getBaseUri().toString().lastIndexOf("/"));
+    StringBuilder url = new StringBuilder(uri);
```
rdenarie: in ImportUsers, uri is build like this :
StringBuilder url = new StringBuilder();
    url.append(request.getScheme()).append("://").append(request.getServerName());
    if (request.getServerPort() != 80 && request.getServerPort() != 443) {
      url.append(':').append(request.getServerPort());
    }
    PortalContainer container = PortalContainer.getCurrentInstance(request.getServletContext());
    url.append(container.getPortalContext().getContextPath());

I thinnk we should follow the same process

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -828,6 +819,27 @@ public Response updateUserById(@Context UriInfo uriInfo,
...
+    sendOnBoardingEmail((UserImpl) user, url);
+    return Response.noContent().build();
```
rdenarie: Can you confirm that Response.noContent() return a 200
Is it better than Response.ok().build() ?

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
rdenarie: :+1:

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
rdenarie: :+1:

rdenarie (dismissed):

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
AzmiTouil: updated!

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
AzmiTouil: I think the most correct is to use PATCH , as we used when update profile property in updateUserProfileAttribute

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
AzmiTouil: no , it return 204.
Updated to Response.ok().build() !

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
AzmiTouil: They return the same result , moreover, I don't have the object HttpServletRequest request

AzmiTouil: They return the same result

AzmiTouil: Done!

rdenarie (approved):

rdenarie (dismissed):

rdenarie (approved):

PR #621, opened 2021-03-12
MAINT-36023: Make sure that applications names and descriptions in spaces settings are translated.

mzorai:
Space applications names and descriptions are not translated in the space settings menu, i made sure to internationalize (I18n) these two properties.

### DIFF
--- a/webapp/portlet/src/main/webapp/space-settings/components/SpaceSettingApplications.vue
+++ b/webapp/portlet/src/main/webapp/space-settings/components/SpaceSettingApplications.vue
@@ -60,13 +60,27 @@ export default {
 
       this.$spaceService.getSpaceApplications(this.spaceId)
         .then(applications => {
-          this.applications = applications;
+          this.applications = applications.map(app => ({
+            applicationName: app.applicationName,
+            contentId: app.contentId,
+            description: this.$t(`SpaceSettings.application.${/\s/.test(app.displayName) ? app.displayName.replace(/ /g,'.').toLowerCase() : app.displayName.toLowerCase()}.description`),
+            displayName: this.$t(`SpaceSettings.application.${/\s/.test(app.displayName) ? app.displayName.replace(/ /g,'.').toLowerCase() : app.displayName.toLowerCase()}.title`),
+            id: app.id(),
+            removable: app.removable,
+          }));
         });
     },
     openDetail() {
       this.$spaceService.getSpaceApplications(this.spaceId)
         .then(applications => {
-          this.applications = applications;
+          this.applications = applications.map(app => ({
+            applicationName: app.applicationName,
+            contentId: app.contentId,
+            description: this.$t(`SpaceSettings.application.${/\s/.test(app.displayName) ? app.displayName.replace(/ /g,'.').toLowerCase() : app.displayName.toLowerCase()}.description`),
+            displayName: this.$t(`SpaceSettings.application.${/\s/.test(app.displayName) ? app.displayName.replace(/ /g,'.').toLowerCase() : app.displayName.toLowerCase()}.title`),
+            id: app.id,
+            removable: app.removable,
+          }));
 
           document.dispatchEvent(new CustomEvent('hideSettingsApps', {detail: this.id}));
           this.displayDetails = true;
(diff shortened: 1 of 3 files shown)


ahamdi (approved):

webapp/portlet/src/main/webapp/space-settings/components/SpaceSettingApplications.vue
```diff
@@ -60,13 +60,27 @@ export default {
...
     },
     openDetail() {
       this.$spaceService.getSpaceApplications(this.spaceId)
         .then(applications => {
-          this.applications = applications;
+          this.applications = applications.map(app => ({
+            applicationName: app.applicationName,
+            contentId: app.contentId,
+            description: this.$t(`SpaceSettings.application.${/\s/.test(app.displayName) ? app.displayName.replace(/ /g,'.').toLowerCase() : app.displayName.toLowerCase()}.description`),
+            displayName: this.$t(`SpaceSettings.application.${/\s/.test(app.displayName) ? app.displayName.replace(/ /g,'.').toLowerCase() : app.displayName.toLowerCase()}.title`),
```
boubaker, line 80: why not having used the space portlet name instead, it will be simple to understand what to put as label in I18N of other addons.
Here the developers will have to make a computing switch display name of the application (replace "a space" by a "." & lowercase everything)
Wouldn't it be more simple to use `SpaceSettings.application.PORTLET_NAME.title` & `SpaceSettings.application.PORTLET_NAME.description` ?

So without reading this code, I will never remember how to make new apps translated
This means, each time I have a new application, I have to search for this code to know how to translate the label of the application, which is not convenient IMO.
This is ok to fix the problem right now, but for the future development, it's not easily maintainable

webapp/portlet/src/main/webapp/space-settings/components/SpaceSettingApplications.vue
```diff
@@ -60,13 +60,27 @@ export default {
 
       this.$spaceService.getSpaceApplications(this.spaceId)
         .then(applications => {
-          this.applications = applications;
+          this.applications = applications.map(app => ({
+            applicationName: app.applicationName,
+            contentId: app.contentId,
+            description: this.$t(`SpaceSettings.application.${/\s/.test(app.displayName) ? app.displayName.replace(/ /g,'.').toLowerCase() : app.displayName.toLowerCase()}.description`),
+            displayName: this.$t(`SpaceSettings.application.${/\s/.test(app.displayName) ? app.displayName.replace(/ /g,'.').toLowerCase() : app.displayName.toLowerCase()}.title`),
+            id: app.id(),
```
boubaker, line 68: A regression is made by this, the `app.id` is an attribute not a function.
(
Try to add an application in a space or delete it => 
```
Uncaught (in promise) TypeError: app.id is not a function
    at eval (SpaceSettingApplications.vue?7190:68)
    at Array.map (<anonymous>)
    at eval (SpaceSettingApplications.vue?7190:63)
```)

boubaker: Instead of altering retrieved data from backend, you should have added a `computed` attribute and displayed in this sub Vue component :
https://github.com/Meeds-io/social/blob/1e9ac24366785aa625da87a86ff80edd4f792899/webapp/portlet/src/main/webapp/space-settings/components/SpaceSettingApplicationCard.vue#L13-L22

This will reduce Object instanciations and add clarity to your code with a real separation between "Data" and "View".
(
How to think it next time: I need to display an I18N of applications retrieved from Backend.
* (**OK**) => So just in place where you make the display you do the I18N of the retrieved applications
* (**KO**) => In all methods where I retrieve applications, I will make the computing of the label and prepare it for display.

mdiha: > A regression is made by this, the `app.id` is an attribute not a function.
> (
> Try to add an application in a space or delete it =>
> 
> ```
> Uncaught (in promise) TypeError: app.id is not a function
>     at eval (SpaceSettingApplications.vue?7190:68)
>     at Array.map (<anonymous>)
>     at eval (SpaceSettingApplications.vue?7190:63)
> ```)
> ```

I didn't notice it, this is probably due to the auto-completion.

webapp/portlet/src/main/webapp/space-settings/components/SpaceSettingApplications.vue
mdiha: Yes you are right i agree

PR #626, opened 2021-03-13
Task-41284 : Add the information internal/exteral in a new column "Type".

ahmadmido:
- This PR will add a new column on UserManegement table "Type".
- This column will add the internal/external information for users.


webapp/portlet/src/main/webapp/idm-users-management/components/UsersManagementList.vue
```diff
@@ -204,6 +213,12 @@ export default {
         align: 'center',
         width: '10%',
         sortable: false,
+      }, {
```
rdenarie: is there a way to reduce duplicated code ? 
Can function superHeaders return something like 
headers() +delete column

So that, all the part present in both is not duplicated

rdenarie (changes requested):

rdenarie (approved):

PR #634, opened 2021-03-17
Task-41620 : LDAP/AD synchronized user informations should not be editable

ahmadmido:
- Before this fix synchronized users are able to edit those pieces of information.
- This fix will assure that synchronized users can't edit pieces of information, a tooltip is displayed if users are synchronized via LDAP/AD with this message: "Impossible to edit synchronized users".


rdenarie (changes requested):
This PR have to be changed in accordance with samuel modification : we have not to add a new attribute. Just check if user source is external and in this case, disable edit butotn

rdenarie (dismissed):

rdenarie (approved):

PR #645, opened 2021-03-23
Task-42088 :  The search in the disabled users view does not work correctly

AzmiTouil:
To fix this issue, we need to update this method to only search in disabled users when users are filtered by disabled


component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -297,11 +298,33 @@ public Response getUsers(@Context UriInfo uriInfo,
...
+          limitToFetch = totalSize - offset;
+        }
+        if (limitToFetch <= 0) {
+          users = new User[0];
+        } else {
+          users = usersListAccess.load(offset, limitToFetch);
+          if (!returnSize) {
+            totalSize = 0;
+          }
+        }
```
```suggestion
          
        }
```

rdenarie (changes requested):

rdenarie (approved):

PR #646, opened 2021-03-23
TASK-41621: Import users via csv their status

hassenelkamel-exo:
**Problems:** 
- Error returned by server is not handled in front.
- null pointer exception when add users from csv file.
-  Cannot set the enabled value from csv.
**Solution:**
- `window.setTimeout` to ensure that error sent by server will be handled.
- set value of enabled when parsing csv file

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
@@ -1414,7 +1414,7 @@ public class UserRestResourcesV1 implements UserRestResources, Startable {
       return userName;
     }
     boolean onboardUser = !userObject.isNull("onboardUser") && userObject.getString("onboardUser").equals("true");
-    boolean enabledUser = !userObject.isNull("enabled") && userObject.getString("enabled").toLowerCase().equals("false");
+    boolean userStatus = !userObject.isNull("enabled") && ( "true".equalsIgnoreCase(userObject.getString("enabled")) ||"false".equalsIgnoreCase(userObject.getString("enabled")));
 
     User existingUser = organizationService.getUserHandler().findUserByName(userName, UserStatus.ANY);
     if (existingUser != null ) {
@@ -1423,6 +1423,9 @@ public class UserRestResourcesV1 implements UserRestResources, Startable {
       }
       // skipping password overwrite from csvLine
       user.setPassword(null);
+      if (userStatus) {
+        organizationService.getUserHandler().setEnabled(userName, Boolean.valueOf(userObject.getString("enabled")), true);
+      }
       organizationService.getUserHandler().saveUser(user, true);
       onboardUser = onboardUser && existingUser.isEnabled() && (existingUser.getLastLoginTime().getTime() == existingUser.getCreatedDate().getTime());
     }
@@ -1475,10 +1478,10 @@ public class UserRestResourcesV1 implements UserRestResources, Startable {
       sendOnBoardingEmail(user, url);
     }
 
-    if (enabledUser) {
+    if (userStatus) {
       organizationService.getUserHandler().setEnabled(userName, Boolean.parseBoolean(userObject.getString("enabled")), true);
     }
-    
+
     // Delete imported User object properties
     userObject.remove("userName");
     userObject.remove("firstName");
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
@@ -1413,9 +1413,8 @@ public class UserRestResourcesV1 implements UserRestResources, Startable {
       userImportResultEntity.addErrorMessage(userName, errorMessage);
       return userName;
     }
-    String status = userObject.getString("status").toLowerCase();
     boolean onboardUser = !userObject.isNull("onboardUser") && userObject.getString("onboardUser").equals("true");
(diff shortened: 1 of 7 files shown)


component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -1338,7 +1338,9 @@ private String importUser(UserImportResultEntity userImportResultEntity,
       userImportResultEntity.addErrorMessage(userName, errorMessage);
       return userName;
     }
+    String status = userObject.getString("status").toLowerCase();
```
plamarque: is « status » name of the column in the file? If so why is is status and not active? or enabled ?

plamarque (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
hassenelkamel-exo: yes, i will change it to enabled

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -1394,6 +1394,7 @@ private String importUser(UserImportResultEntity userImportResultEntity,
       return userName;
     }
     boolean onboardUser = !userObject.isNull("onboardUser") && userObject.getString("onboardUser").equals("true");
+    boolean enabledUser = !userObject.isNull("enabled") && userObject.getString("enabled").toLowerCase().equals("false");
```
ahamdi: We may rename the variable to be more precise with the value that it will get
```suggestion
    boolean userStatus = !userObject.isNull("enabled") && ("true".equalsIgnoreCase(userObject.getString("enabled") || "false".equalsIgnoreCase(userObject.getString("enabled"));
```

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -1420,35 +1421,43 @@ private String importUser(UserImportResultEntity userImportResultEntity,
...
       }
     }
     //onboard user if the onboardUser csv field is true, the user is enabled and not yet logged in 
     if (onboardUser) {
       sendOnBoardingEmail(user, url);
     }
+
+    if (enabledUser) {
+      organizationService.getUserHandler().setEnabled(userName, Boolean.parseBoolean(userObject.getString("enabled")), true);
+    }
```
ahamdi:
```suggestion
    if (userStatus) {
      organizationService.getUserHandler().setEnabled(userName, Boolean.parseBoolean(userObject.getString("enabled")), true);
    }
```

component/service/src/test/java/org/exoplatform/social/rest/impl/users/UserRestResourcesTest.java
```diff
@@ -659,6 +668,33 @@ public void testImportUsers() throws Exception {
     assertTrue(result);
     }
 
+    // Test import enabled or disabled users
```
ahamdi, line 674: Could you add UT for updating user status after importing them.

ahamdi (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -1553,6 +1566,9 @@ private void logMetrics(ExoSocialActivity activity) {
 
   private void saveProfile(String username, ProfileEntity profileEntity) throws Exception {
     Identity userIdentity = getUserIdentity(username);
+    if(userIdentity == null) {
```
ahamdi: This will return NULL if the user is anonymous or is not present -> IdentityManager is enabled to retrieve him. I am not sure if we need to recheck again in this function.

component/service/src/test/java/org/exoplatform/social/rest/impl/users/UserRestResourcesTest.java
```diff
@@ -655,10 +666,66 @@ public void testImportUsers() throws Exception {
...
+    importResultEntity = (UserImportResultEntity) response.getEntity();
+    assertEquals(2, importResultEntity.getCount());
+    assertEquals(importResultEntity.getCount(), importResultEntity.getProcessedCount());
+
+    assertNull(importResultEntity.getErrorMessages());
+    assertNull(importResultEntity.getWarnMessages());
+    // Status after update
+    boolean userStatusUpdated1 = organizationService.getUserHandler().findUserByName("usera", UserStatus.DISABLED).isEnabled();
+    boolean userStatusUpdated2 = organizationService.getUserHandler().findUserByName("userb", UserStatus.ENABLED).isEnabled();
+    assertEquals(userStatusUpdated1, false);
```
ahamdi: assertFalse

component/service/src/test/java/org/exoplatform/social/rest/impl/users/UserRestResourcesTest.java
```diff
@@ -655,10 +666,66 @@ public void testImportUsers() throws Exception {
...
+    assertEquals(userStatusUpdated2, true);
```
ahamdi: assertTrue

ahamdi (approved):

hassenelkamel-exo:
@plamarque we need your review to merge this PR

plamarque (approved):

PR #647, opened 2021-03-24
Task-42177: Type is not correctly displayed when searching external users in usermanagment

AzmiTouil:
To fix this problem we need to load profile for each identity to get the identity type

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
@@ -679,18 +679,12 @@ public class RDBMSIdentityStorageImpl implements IdentityStorage {
                                                  long offset,
                                                  long limit,
                                                  boolean forceLoadOrReloadProfile) throws IdentityStorageException {
-    Identity viewerIdentity = null;
+    Identity identity = null;
     if (profileFilter.getViewerIdentity() != null) {
-      viewerIdentity = profileFilter.getViewerIdentity();
+      identity = profileFilter.getViewerIdentity();
     }
     if (OrganizationIdentityProvider.NAME.equals(providerId)) {
-      List<Identity> identities = getProfileSearchConnector().search(viewerIdentity, profileFilter, type, offset, limit);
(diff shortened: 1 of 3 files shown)


component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
```diff
@@ -679,12 +679,18 @@ public void processEnabledIdentity(Identity identity, boolean isEnable) {
...
     }
     if (OrganizationIdentityProvider.NAME.equals(providerId)) {
-      return getProfileSearchConnector().search(identity, profileFilter, type, offset, limit);
+      List<Identity> identities = getProfileSearchConnector().search(viewerIdentity, profileFilter, type, offset, limit);
+      if (forceLoadOrReloadProfile) {
+        for (Identity identity : identities) {
+          identity.setProfile(loadProfile(identity.getProfile()));
+        }
+      }
+      return identities;
```
rdenarie: Revert theses modifications and add the need profile field in ProfileSearchConnector as we did for syncDate and enrollmentDate

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -298,7 +298,7 @@ public Response getUsers(@Context UriInfo uriInfo,
       filter.setSkills(q == null || q.isEmpty() ? "" : q);
       filter.setExcludeExternal(excludeExternal);
       filter.setEnabled(isEnabled == null ? true : Boolean.valueOf(isEnabled));
-      ListAccess<Identity> list = identityManager.getIdentitiesByProfileFilter(OrganizationIdentityProvider.NAME, filter, false);
+      ListAccess<Identity> list = identityManager.getIdentitiesByProfileFilter(OrganizationIdentityProvider.NAME, filter, true);
       identities = list.load(offset, limit);
```
rdenarie: revert also this modification

rdenarie (changes requested):

rdenarie (approved):

PR #650, opened 2021-03-25
Task-28314: Fix deactivated Admin notifications settings still appears in users notifications settings UI

NaderSai:
Problem: There wasn't a condition to check if the option was disabled by the admin.
How it was solved: By adding a condition in the switch tag to disable the disabled options by the admin.


ahamdi (approved):
LGTM

PR #651, opened 2021-03-25
Task-42088 : The search in the disabled users view does not work correctly

AzmiTouil:
To fix this issue, we need to update this method to only search in disabled users when users are filtered by disabled


rdenarie (changes requested):
- Add a test to check isDisabled take correct default value
- Add missing change reverted in UsersManagementList.vue

rdenarie (approved):

PR #652, opened 2021-03-26
TASK-41621: Import users via csv their status

mzorai:
**Problems:** 
- Error returned by server is not handled in front.
- null pointer exception when add users from csv file.
-  Cannot set the enabled value from csv.
**Solution:**
- `window.setTimeout` to ensure that error sent by server will be handled.
- set value of enabled when parsing csv file


hbenali (approved):

ahamdi (approved):
LGTM

PR #662, opened 2021-04-01
Task-41420: Secure malware detection rest endpoint call by token

azayati:
This PR will ensure the security of malware detection rest endpoint call by a token passed as query param. the token should match the following property:
> exo.malwareDetection.passphrase

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/service/malwareDetection/rest/MalwareDetectionRestService.java
+++ b/component/service/src/main/java/org/exoplatform/social/service/malwareDetection/rest/MalwareDetectionRestService.java
@@ -28,12 +28,15 @@ import org.exoplatform.social.service.malwareDetection.MalwareDetectionService;
 @Api(tags = "/malware", value = "/malware", description = "Malware detection")
 public class MalwareDetectionRestService implements ResourceContainer {
 
-  private MalwareDetectionService malwareDetectionService;
+
   private static final String MALWARE_DETECTION_PASSPHRASE_PROPERTY = "exo.malwareDetection.passphrase";
   private static final String MALWARE_DETECTION_PASSPHRASE_PROPERTY_DEFAULT = "@malwareDetection/TOCHANGE!@";
+  private MalwareDetectionService malwareDetectionService;
+  private String malwareDetectionPassphrase;
   
   public MalwareDetectionRestService(MalwareDetectionService malwareDetectionService) {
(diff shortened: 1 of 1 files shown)


component/service/src/main/java/org/exoplatform/social/service/malwareDetection/rest/MalwareDetectionRestService.java
```diff
@@ -28,21 +29,27 @@
...
       httpMethod = "POST",
       response = Response.class,
       notes = "This processes infected items")
   @ApiResponses(value = {
-      @ApiResponse(code = 200, message = "Request fulfilled") })
-  public Response processInfectedItems(List<String> infectedItemsPaths) throws Exception {
+      @ApiResponse(code = 200, message = "Request fulfilled"),
+      @ApiResponse(code = 403, message = "Unauthorized to process infected items")})
+  public Response processInfectedItems(List<String> infectedItemsPaths, @QueryParam("token") String token) throws Exception {
+    String malwareDetectionPassphrase = System.getProperty(MALWARE_DETECTION_PASSPHRASE_PROPERTY, MALWARE_DETECTION_PASSPHRASE_PROPERTY_DEFAULT);
```
rdenarie: Should this property be read in constructor ? so that you dont have to reread it at each call

rdenarie (changes requested):

component/service/src/main/java/org/exoplatform/social/service/malwareDetection/rest/MalwareDetectionRestService.java
```diff
@@ -27,22 +28,30 @@
 @Api(tags = "/malware", value = "/malware", description = "Malware detection")
 public class MalwareDetectionRestService implements ResourceContainer {
 
+
+  private static final String MALWARE_DETECTION_PASSPHRASE_PROPERTY = "exo.malwareDetection.passphrase";
+  private static final String MALWARE_DETECTION_PASSPHRASE_PROPERTY_DEFAULT = "@malwareDetection/TOCHANGE!@";
```
rdenarie: Should we put a default value ? 

In fact, if someone forget to change that, the service is accessible with the defualt passphrase.
If we dont put a default passphrase, and do not let the access to the service is the passphrase is not configured, it is more secured.

rdenarie (changes requested):

rdenarie (approved):

PR #667, opened 2021-04-07
TASK-42704: The space manager is redactor by default

hassenelkamel-exo:


component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
```diff
@@ -249,9 +249,10 @@ public Response createSpace(@Context UriInfo uriInfo,
     space.setMembers(members);
 
     //
-    spaceService.createSpace(space, authenticatedUser, model.getInvitedMembers());
+    Space createdSpace = spaceService.createSpace(space, authenticatedUser, model.getInvitedMembers());
+    spaceService.addRedactor(createdSpace, authenticatedUser);
```
boubaker: This is a business logical, so it should be handled in Service layer not in REST. In addition, this should be added in the same place where the space creator is promoted as manager by default.

webapp/portlet/src/main/webapp/people-list/components/PeopleCardFront.vue
```diff
@@ -258,7 +259,7 @@ export default {
       return this.profileActionExtensions.slice().filter(extension => extension.enabled(this.user));
     },
     canUseActionsMenu() {
-      return this.user && this.enabledProfileActionExtensions.length;
+      return this.user && this.user.username !== this.currentUserName  && this.enabledProfileActionExtensions.length;
```
boubaker: Why this ? If I'm manager of the space, I shouldn't be able to promote myself or revoke myself as redactor ?

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
hassenelkamel-exo: ok, i will make the changes

webapp/portlet/src/main/webapp/people-list/components/PeopleCardFront.vue
hassenelkamel-exo: The option "Set as a redactor" must be deleted for space managers,

azayati: If the authenticated user is a manager, he shouldn't be able to promote himself or any other manager as a redactor because all space managers will be by default and permanent redactors. @hassenelkamel-exo will rework the condition

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
azayati: yes normally @hassenelkamel-exo we should not touch the space creation, we should only change the condition of post button display in order to be always accessible to space managers

PR #668, opened 2021-04-07
Task 42671 : Change the UI of the filter field

Rdidmohamedelhadi:


boubaker (approved):

webapp/portlet/src/main/webapp/common/components/ExoIdentitySuggester.vue
```diff
@@ -182,6 +183,12 @@ export default {
         return null;
       },
     },
+    innerIcon: {
```
boubaker: We will have to differentiate the display of the suggester input by using an explicit and meaningful variable name like `filterStyle` prop of type boolean that will have `false` as default value.

webapp/portlet/src/main/webapp/common/components/ExoIdentitySuggester.vue
```diff
@@ -15,7 +15,8 @@
       :height="height"
       :filter="filterIgnoredItems"
       :hide-no-data="hideNoData"
-      :class="(required && !value && 'required-field invalid') || (required && 'required-field')"
+      :class="(required && !value && 'required-field invalid') || (required && 'required-field') || (innerIcon && 'customIdentitySuggester' && 'customIconIdentitySuggester') || (!innerIcon && 'identitySuggesterStyle')"
```
boubaker: Add a computed property to : 
1/ use required style or not
2/ use `identitySuggesterFilterStyle` class

```vue.js
computed: {
  autocompleteClass() {
    const requiredClass = (this.required && !this.value && 'required-field invalid') || (this.required && 'required-field') || '';
    const sugesterStyleClass = this.filterStyle && 'identitySuggesterFilterStyle' || 'identitySuggesterInputStyle';
    return `${requiredClass} ${sugesterStyleClass}`;
  }
}
```

webapp/portlet/src/main/webapp/common/components/ExoIdentitySuggester.vue
```diff
@@ -15,7 +15,8 @@
...
+      :prepend-inner-icon="innerIcon ? 'fa-filter' : ''"
```
boubaker: Same here, you can use a computed property

webapp/portlet/src/main/webapp/common/components/ExoIdentitySuggester.vue
```diff
@@ -195,6 +202,14 @@ export default {
     };
   },
   computed: {
+    icon() {
```
boubaker:
```suggestion
    prependInnerIcon() {
```

webapp/portlet/src/main/webapp/common/components/ExoIdentitySuggester.vue
```diff
@@ -182,6 +183,12 @@ export default {
         return null;
       },
     },
+    filterStyle: {
+      type: Boolean,
+      default: function () {
+        return false;
+      },
```
boubaker: When it's boolean, you can avoid declaring the default value as a function
```suggestion
      default: false,
```

Rdidmohamedelhadi: done

boubaker (approved):

PR #671, opened 2021-04-07
Task-42676: Add possibility to search users by email on usersManagement

ahmadmido:
- Before this fix, search users by email isn't working on users Management portlet.
- This fix will ensure that filtered users by emails are working for enabled users by setting searchByemail property as true on the filter and for disabled users.


component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -296,9 +296,17 @@ public Response getUsers(@Context UriInfo uriInfo,
       filter.setSkills(q == null || q.isEmpty() ? "" : q);
       filter.setEnabled(!isDisabled);
       filter.setUserType(userType);
+      filter.setSearchEmail(true);
       if (isDisabled && q != null && !q.isEmpty()) {
+        ListAccess<User> usersListAccess;
         User[] users;
-        ListAccess<User> usersListAccess = userSearchService.searchUsers(q, UserStatus.DISABLED);
+        if(q.contains("@")) {
```
rdenarie, line 303: as userSearchService calls orgSe'rvice.getUserHandler.findUsersByQuery, it is possible to remove this if ?
I dont understand why we make a different treatment when query contains @ or not, can you explain ?

rdenarie (dismissed):

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
azayati: to be treated in the new US as explained by @ahmadmido

azayati (approved):
LGTM

PR #680, opened 2021-04-13
TASK-40949: fixed unchanging space description value in space stream

nidhalnidhal:
Before this fix , when changing a space description in space settings , this value remains the same in the space bloc description in space stream which is caused by the fact that the browser is always rendering a stale cache value . So i made few changes in the method getspacebyID in SpaceRestResourcesV1 so that the browser demands a new fresh version of the resource in case it changed based on the value of the etag related to the resource.


component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
```diff
@@ -285,15 +285,10 @@ public Response getSpaceById(@Context UriInfo uriInfo,
     Response.ResponseBuilder builder = request.evaluatePreconditions(eTag);
     if (builder == null) {
       builder = EntityBuilder.getResponseBuilder(EntityBuilder.buildEntityFromSpace(space, authenticatedUser, uriInfo.getPath(), expand), uriInfo, RestUtils.getJsonMediaType(), Response.Status.OK);
-      builder.tag(eTag);
+      return builder.cacheControl(cc).tag(eTag).build();
```
ahamdi: remove this line as it is duplicated.

ahamdi (changes requested):

ahamdi (approved):

PR #682, opened 2021-04-13
Task-42425: Allow searching redactional spaces for super admin, super spaces manager and space manager

hassenelkamel-exo:
**Current behavior:** The spaces suggestor will not display:
- Redactional spaces (with at least one redactor) on which I'm manager or when I'm a super spaces manager or super admin

**New behavior:** The spaces suggestor will not display:
- Redactional spaces (with at least one redactor) on which I'm manager or when I'm a super spaces manager or super admin

### DIFF
--- a/webapp/portlet/src/main/webapp/common/js/SuggesterService.js
+++ b/webapp/portlet/src/main/webapp/common/js/SuggesterService.js
@@ -24,7 +24,6 @@ function searchSpaces(filter, items, excludeNonRedactor, noRedactorSpace, exclud
     .then(resp => resp && resp.ok && resp.json())
     .then(data => {
       data.spaces.forEach((item) => {
-      // canEdit = true if manager or super manager
         if ((!noRedactorSpace || !item.redactorsCount) && (!excludeNonRedactor || item.isRedactor || !item.redactorsCount) || (!excludeNonManager || item.canEdit)) {
           items.push({
             id: `space:${item.prettyName}`,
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/common/js/SuggesterService.js
```diff
@@ -24,7 +24,8 @@ function searchSpaces(filter, items, excludeNonRedactor, noRedactorSpace) {
     .then(resp => resp && resp.ok && resp.json())
     .then(data => {
       data.spaces.forEach((item) => {
-        if ((!noRedactorSpace || !item.redactorsCount) && (!excludeNonRedactor || item.isRedactor || !item.redactorsCount)) {
+      // canEdit = true if manager or super manager
```
azayati: to be removed

webapp/portlet/src/main/webapp/common/js/SuggesterService.js
```diff
@@ -24,7 +24,8 @@ function searchSpaces(filter, items, excludeNonRedactor, noRedactorSpace) {
...
+        if ((!noRedactorSpace || !item.redactorsCount) && (!excludeNonRedactor || item.isRedactor || !item.redactorsCount) || (!excludeNonManager || item.canEdit)) {
```
azayati, line 27: Why adding item.canEdit here ?

azayati (changes requested):
See review

webapp/portlet/src/main/webapp/common/js/SuggesterService.js
hassenelkamel-exo: it return true if user is manager or super manager to allow suggest space for them

azayati (approved):
LGTM

PR #691, opened 2021-04-20
Task-43073: Add user management navigation for delegated administrators

AzmiTouil:


azayati (dismissed):
LGTM waiting QA validation

azayati (dismissed):
LGTM after QA validation

rdenarie (commented):
Hello,
I have few questions : 

1. We are creating a new navigation. Are we sure that all existing instance, with existing data will automatically create the new navigation/page without additionnal step/property ?
2. An administrator which is also delegated will 2 user management page. Do you ask the question to the PO to confirm that this point is ok ?

AzmiTouil:
> Hello,
> I have few questions :
> 
> 1. We are creating a new navigation. Are we sure that all existing instance, with existing data will automatically create the new navigation/page without additionnal step/property ?
> 2. An administrator which is also delegated will 2 user management page. Do you ask the question to the PO to confirm that this point is ok ?

1.Yes
2.We have already discussed it

rdenarie (approved):

PR #694, opened 2021-04-27
Task-43074: users management page for delegated administrators

AzmiTouil:


component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -846,7 +876,7 @@ public Response updateUserById(@Context UriInfo uriInfo,
   @PATCH
   @Path("onboard/{id}")
   @Produces(MediaType.APPLICATION_JSON)
-  @RolesAllowed("administrators")
+  @RolesAllowed("users")
```
rdenarie, line 882: you should check that the current user is admin or admin delegate.
Else a simple user could send onboarding mail to anybody

rdenarie (changes requested):

rdenarie (dismissed):

rdenarie (approved):

PR #700, opened 2021-05-04
42662 : fix changing start/due date for composite language codes

ahamdi:
an error was thrown when we try to pass a composite language to get its language formatted value.
The fix will strip the language code and remove the Country code if it exists.


lbengayed (approved):
LGTM

Rdidmohamedelhadi (approved):

PR #701, opened 2021-05-05
Task-42849: Implement share regular activity V1

azayati:
This PR will implement:

- shareActivityOnSpaces rest service which will be invoked by the share activity on spaces drawer which will be implemented later)
- first version of the shared regular activity template (to be updated later with the final UI)

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
@@ -16,21 +16,55 @@
  */
 package org.exoplatform.social.rest.impl.space;
 
-import java.io.*;
-import java.util.*;
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
+
+import java.io.ByteArrayInputStream;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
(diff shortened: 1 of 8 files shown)


hassenelkamel-exo (approved):

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
```diff
@@ -1153,6 +1195,60 @@ public Response postActivityOnSpace(@Context UriInfo uriInfo,
...
+
+    if (sharedActivityRestIn == null || sharedActivityRestIn.getTargetSpaces() == null || sharedActivityRestIn.getTargetSpaces().isEmpty() || sharedActivityRestIn.getType() == null) {
+      return Response.status(Response.Status.BAD_REQUEST).build();
+    }
+    
+    Identity authenticatedUserIdentity = identityManager.getOrCreateIdentity(OrganizationIdentityProvider.NAME, authenticatedUser);
+    List<ActivityEntity> sharedActivitiesEntities = new ArrayList<ActivityEntity>();
+    for (String targetSpaceName : sharedActivityRestIn.getTargetSpaces()) {
+      Space targetSpace = spaceService.getSpaceByPrettyName(targetSpaceName);
+      if (SpaceUtils.isSpaceManagerOrSuperManager(authenticatedUser, targetSpace.getGroupId()) || (spaceService.isMember(targetSpace, authenticatedUser) && SpaceUtils.isRedactor(authenticatedUser, targetSpace.getGroupId()))) {
```
plamarque, line 1232: this is silently ignoring attempts to share in spaces where user is not supposed to.
first it's questionable wether HTTP 200 should be used in such situation since, it was not fully completed (you have no way to tell the user that the share didn't work on all spaces). But more importantly  IMO, we should at least log it server side as it may be some kind of attack or denoting an inconsistency bug

PR #704, opened 2021-05-06
QWIN-44121 : Make the read more notification label translatable

hakermi:
Make the read more notification label translatable


extension/notification/src/main/resources/locale/notification/template/Notification_en.properties
```diff
@@ -138,6 +138,7 @@ Notification.intranet.message.one.EditCommentPlugin={0} edited a comment on your
 Notification.intranet.message.two.EditCommentPlugin={0} and {1} edited their comments on your post.
 Notification.intranet.message.more.EditCommentPlugin={0}, {1} and {2} edited comments.
 Notification.intranet.message.one.EditActivityPlugin={0} edited a post.
+Notification.intranet.readMore=Read More
```
ahamdi: To avoid having intranet in the key which is the old default site.
```suggestion
Notification.label.readMore=Read More
```

ahamdi (changes requested):

ahamdi (approved):

PR #705, opened 2021-05-07
Task-44094 : User Type on UserManegement search is not updated when an internal user become external

ahmadmido:
- Before this fix, when we add an internal user to an external group, the profile property external isn't updated.
- This fix will ensure that when we add users to an external group, the profile external property is changed to true.


rdenarie (changes requested):
I think you should use an existing event : your listener should extends MembershipEventListener, and make action if group is /platform/externals

rdenarie (approved):

PR #707, opened 2021-05-07
Task-43958 : If the external group is provided ignore it from imported CSV users

ahmadmido:
- Before this fix, if a CSV file containing a user on /platform/externals, it's will be added to the external group.
- This fix will ensure that if a user on CSV file is a member on /platform/externals ignore it.


component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -1498,6 +1498,7 @@ private String importUser(UserImportResultEntity userImportResultEntity,
       if (StringUtils.isNotBlank(groups)) {
         List<String> groupsList = Arrays.asList(groups.split(";"));
         for (String groupMembershipExpression : groupsList) {
+          if (groupMembershipExpression.equals("/platform/externals")) continue;
```
```suggestion
          if (groupId.equals("/platform/externals")) continue;
```
This line should be after the groupId definition

rdenarie (changes requested):

rdenarie (approved):

PR #717, opened 2021-05-21
Task 41382: Fix External users are searchable in people directory

NaderSai:
Problem: after the delete of user from membership the indexing in the Elastic Search was the same as before.
How it was solved: by unindexing this user after its delete.

### DIFF
--- a/component/common/pom.xml
+++ b/component/common/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <artifactId>social-component</artifactId>
     <groupId>org.exoplatform.social</groupId>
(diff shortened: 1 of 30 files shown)


ahamdi (changes requested):
This can not be the fix for this issue, the issue is about external users. But the fix is adding an already existing logic for user removal .

rdenarie:
Close this PR, as the fix was done by https://community.exoplatform.com/portal/dw/tasks/taskDetail/44094
Backport is done here : https://github.com/Meeds-io/social/pull/932

PR #726, opened 2021-05-26
Task-43673: Clean space suggester of activity share drawer after share activity

hassenelkamel-exo:
After sharing an activity or news in space stream we will clear the suggester of space


azayati (approved):
LGTM after QA validation

PR #732, opened 2021-05-28
Task-44606: Write an article button is not shown until an explicit refresh of the space activity stream

AzmiTouil:
The problem is that the newsComposerPlugin is loaded only at the first site page access and not when switching from one page to another, so the "enabled" parameter will keep the initial value.
To fix this problem, we have to call the "enabled" method each time the component loads.

### DIFF
--- a/webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
+++ b/webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
@@ -236,7 +236,7 @@ export default {
       const urls = [];
       this.activityComposerActions = getActivityComposerActionExtensions();
       this.activityComposerActions.forEach(action => {
-        this.checkEnabled(action);
+        this.setExtensionEnabled(action);
         if (action.component) {
           this.actionsData[action.key] = action.component.model.value;
           this.actionsEvents[action.key] = action.component.events;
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
```diff
@@ -183,6 +184,11 @@ export default {
     },
     activityBodyEdited: function() {
       return this.activityId ? this.escapeHTML(this.message.split('<oembed>')[0]) === this.escapeHTML(this.cleanInitialActivityBody(this.activityBody)) : false;
+    },
+    enabledActivityComposerActions: function () {
```
azayati, line 188:
```suggestion
    enabledActivityComposerActions: function () {
    return this.activityComposerActions && this.activityComposerActions.filter(activityComposerAction => isExtensionEnabled(activityComposerAction));
    
```

webapp/portlet/src/main/webapp/activity-composer-app/components/ExoActivityComposer.vue
```diff
@@ -240,6 +247,23 @@ export default {
       });
       exoi18n.loadLanguageAsync(eXo.env.portal.language, urls);
     },
+    checkEnabled(action) {
```
azayati: keep the same function isExtensionEnabled of extension.js

azayati (changes requested):
See review

PR #748, opened 2021-06-03
Task-38140: Fix When searching a user and clicking on i the about me field is empty

NaderSai:
Problem: The about me property is not displayed when searching user from the unified search after clicking on the "i" icon.
How it was solved: by setting the about me property in the user object from the search result returned by the Elastic Search engine.


ahamdi (approved):
LGTM

PR #750, opened 2021-06-04
TASK-45123 : Deleted ActivitiesRestService service

nidhalnidhal:
i deleted ActivitiesRestService class and all its dependencies as it is no longer used to fetch data and it comes with a huge  security hole because all end points are not marked with @RolesAllowed .


ahamdi (changes requested):
The PR proposed changes that will revert what was done lately to cleanup dependencies. Please recreate it based on latest version on feature/maintenance

ahamdi (approved):

PR #755, opened 2021-06-04
MAINT-44337: fix edit short message activity when it contains a link with preview plus an attached file

hakermi:
**FIX** edit short message activity when it contains a link with preview plus an attached file by using adapting the `UIlinkActivtyBuilder` to use the **MESSAGE** param as it will be available when the activity contains an attached file and it's extended in the same time by the `LinkActivity` params, also updating the rest service to be able to update the **MESSAGE** param when editing the activity as the old implementation will **never** update and only updating the title.

### DIFF
--- a/component/common/pom.xml
+++ b/component/common/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <artifactId>social-component</artifactId>
     <groupId>org.exoplatform.social</groupId>
-    <version>6.2.x-SNAPSHOT</version>
+    <version>6.2.x-maintenance-SNAPSHOT</version>
   </parent>
   <groupId>org.exoplatform.social</groupId>
   <artifactId>social-component-common</artifactId>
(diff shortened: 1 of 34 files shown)


ahamdi (changes requested):
Please add a Unit test

hakermi:
> Please add a Unit test

done

ahamdi:
It seems my request was not clear, I wanted a UT for org.exoplatform.social.rest.impl.activity.ActivityRestResourcesV1.java more than the UILinkActivity.java. could you cover that class too ?

hakermi:
> It seems my request was not clear, I wanted a UT for org.exoplatform.social.rest.impl.activity.ActivityRestResourcesV1.java more than the UILinkActivity.java. could you cover that class too ?

Okay

hakermi:
> It seems my request was not clear, I wanted a UT for org.exoplatform.social.rest.impl.activity.ActivityRestResourcesV1.java more than the UILinkActivity.java. could you cover that class too ?

Done

ahamdi (approved):

PR #759, opened 2021-06-07
Task-45005: Allow a super user or a super spaces admin to view space activities on which he is not a member

hassenelkamel-exo:
**Current behavior:**  When a super admin or super spaces admin try to access to an space activity on which he is not a member, he will be redirected to a page "Activity not found"
**New behavior:** Super admin or space admin will see space activities on which he is not a member


azayati (approved):
LGTM after QA validation

PR #763, opened 2021-06-08
TASK-45373: Fix Applications' name and description are not displayed by their value in space settings interface

NaderSai:
Problem: the display name and description of an application wasn't displayed correctly in space settings application.
How it was solved: by replacing the static values read from the .properties files by dynamic values read in the SpaceSettingApplicationCard component for installed applications and in the SpaceSettingAddApplicationDrawer component for other applications.


webapp/portlet/src/main/webapp/space-settings/components/SpaceSettingAddApplicationDrawer.vue
```diff
@@ -60,8 +60,8 @@ export default {
         this.applications = data.map(app => ({
           applicationName: app.applicationName,
           contentId: app.contentId,
-          description: this.$t(`SpaceSettings.application.${/\s/.test(app.displayName) ? app.displayName.replace(/ /g,'.').toLowerCase() : app.displayName.toLowerCase()}.description`),
-          displayName: this.$t(`SpaceSettings.application.${/\s/.test(app.displayName) ? app.displayName.replace(/ /g,'.').toLowerCase() : app.displayName.toLowerCase()}.title`),
```
ahamdi, line 64: How will those applications names and descriptions will be translated ?

ahamdi (changes requested):

webapp/portlet/src/main/webapp/space-settings/components/SpaceSettingAddApplicationDrawer.vue
NaderSai: This two lines are the cause of the problem of displaying the name and description. But right, i will try to find a solution to recover the two options.

PR #768, opened 2021-06-10
Fix the backport of MAINT-38140

hakermi:
Fix the backport of MAINT-38140  as it needs more changes in the 6.1.x version

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
@@ -24,6 +24,8 @@ import java.util.Map;
 
 import org.exoplatform.commons.search.es.ElasticSearchException;
 import org.exoplatform.commons.search.es.client.ElasticSearchingClient;
+import org.exoplatform.commons.utils.CommonsUtils;
+import org.exoplatform.social.core.manager.IdentityManager;
 import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
 import org.json.simple.parser.JSONParser;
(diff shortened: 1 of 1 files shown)


component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
```diff
@@ -142,10 +144,13 @@ private int getCount(String jsonResponse) {
       String avatarUrl = (String) hitSource.get("avatarUrl");
       String email = (String) hitSource.get("email");
       String identityId = (String) ((JSONObject) jsonHit).get("_id");
-      String aboutMe = (String) hitSource.get("aboutMe");
       identity = new Identity(OrganizationIdentityProvider.NAME, userName);
       identity.setId(identityId);
       p = new Profile(identity);
+      Profile profile = getIdentityProfile(userName);
+      if (profile == null) {
+        continue;
```
plamarque, line 152: I think this is swallowing information when some data retrieved from ES is not wellformed (userName expected, but not in the result). Shouldn't we log (log.warn if unexpected,  log.debug() if it's common and) the fact ?  
ex: "WARN : ES hit $HIT contained invalid userName information $USER and was ignored"

component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
```diff
@@ -142,10 +144,13 @@ private int getCount(String jsonResponse) {
       String avatarUrl = (String) hitSource.get("avatarUrl");
       String email = (String) hitSource.get("email");
       String identityId = (String) ((JSONObject) jsonHit).get("_id");
-      String aboutMe = (String) hitSource.get("aboutMe");
       identity = new Identity(OrganizationIdentityProvider.NAME, userName);
       identity.setId(identityId);
       p = new Profile(identity);
+      Profile profile = getIdentityProfile(userName);
```
ahamdi, line 150: We are getting information from ES, no need to reload the profile from storage

component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
```diff
@@ -142,10 +144,13 @@ private int getCount(String jsonResponse) {
...
+      if (profile == null) {
```
ahamdi, line 151: @hakermi This should not b e a normal case, do you have a steps to reproduce such a behaviour ?

ahamdi (changes requested):

PR #775, opened 2021-06-11
Task-45373: Fix Applications' name and description are not displayed by their value in space settings interface

NaderSai:
Problem: the name and the description of apps were not displayed by their values in the addApplication drawer.
How it was solved: delete the duplicated values, already read from the SpaceSettingApplicationCard component, from the SpaceSettingAddApplicationDrawer component.


ahamdi (changes requested):
This will break the intyernationalization of th edescription and title keys of all applications. Please make sure that the fix will work and display correctly each application's title and description

NaderSai:
> This will break the intyernationalization of th edescription and title keys of all applications. Please make sure that the fix will work and display correctly each application's title and description
the content of apps is charged from the SpaceSettingApplicationCard in which the internationalization is respected. i verified that and it works.

ahamdi (dismissed):

ahamdi (approved):

PR #783, opened 2021-06-17
Add Sonar Cloud configuration

rdenarie:


.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
+name: Build
```
hbenali:
```suggestion
name: Sonar Analysis
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  push:
+    branches:
+      - develop
```
hbenali, line 5:
```suggestion
```
No need to run sonar after PR merge on develop branch

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened]
```
hbenali:
```suggestion
```
Let's keep the default behavior of PR state

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Build and analyze
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        run: mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar -Dmaven.test.skip=true
```
hbenali: Not sure if we need to keep Unit tests and include the exo-release profile.

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
+name: Build
+on:
+  push:
+    branches:
+      - develop
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+  build:
+    name: Build
```
hbenali:
```suggestion
    name: Analyse Module with SonarCloud
```

.github/workflows/build.yml
rdenarie: I think it is necessary, as we need  to have a view on current sonar status on develop. So I prefer to let it on develop

.github/workflows/build.yml
rdenarie: we dont want unit test here, because tests are run on jenkins. We only need code examination

hbenali: Okey thanks.

hbenali (dismissed):

pom.xml
```diff
@@ -268,5 +274,32 @@
       </plugins>
     </pluginManagement>
   </build>
-
+  <profiles>
+    <profile>
+      <id>project-repositories</id>
```
boubaker, line 278: no need to add this since it's already declared in parent POM and thus will be inherited, see : 
https://github.com/Meeds-io/maven-parent-pom/blob/develop/pom.xml#L1894

pom.xml
```diff
@@ -52,6 +52,12 @@
 
     <!-- For surefire and failsafe to be compatible with jacoco -->
     <argLine>-Xmx1024m -XX:MaxPermSize=512m</argLine>
+    
+    <!-- Sonar properties -->
+    <sonar.projectKey>Meeds-io_social</sonar.projectKey>
```
boubaker: we can declare less variables, by adding in parent pom properties : 
```xml
<sonar.projectKey>${project.groupId}</sonar.projectKey>
<sonar.host.url>https://sonarcloud.io</sonar.host.url>
<sonar.moduleKey>${project.artifactId}</sonar.moduleKey>
```

pom.xml
rdenarie: It is necessary as maven-parent-pom is in our public repository. The github action is unable to find it if we dont put the repository here

.github/workflows/build.yml
```diff
@@ -0,0 +1,35 @@
+name: Sonar Analysis
+on:
+  push:
+    branches:
+      - develop
+  pull_request:
+jobs:
+  build:
+    name: Analyse Module with SonarCloud
+    runs-on: ubuntu-latest
```
boubaker, line 10: I think we have a lighter linux distribution for docker: may be Alpine

.github/workflows/build.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+      - name: Set up JDK 11
```
boubaker, line 15: This is not yet compatible with all projects.
It's still in progress.

pom.xml
rdenarie: ok, I will do that.

.github/workflows/build.yml
rdenarie: done

.github/workflows/build.yml
rdenarie: needed for sonar

boubaker: ok, so we need to wait on JDK 11 support for projects, isn't it ?

.github/workflows/build.yml
hbenali: We should respect the PRD

.github/workflows/build.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Build and analyze
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        run: mvn -B compile org.sonarsource.scanner.maven:sonar-maven-plugin:sonar -DskipTests
```
boubaker:
```suggestion
        run: mvn -B compile org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
```
No need to  `-DskipTests` now that we don't use `verify` anymore

.github/workflows/build.yml
rdenarie: we can activate the sonar analysis now. It is not blocker if fail on PR. And it will indicate us if something is not building on jdk 11.
I think we should not wait

.github/workflows/build.yml
rdenarie: I not sure to get in what this not respect the PRD ? 
It is the image used to build the PR (or the branch) not to run exo

pom.xml
rdenarie: done

.github/workflows/build.yml
rdenarie: done

hbenali (approved):

.github/workflows/build.yml
boubaker: ok, let's do it :-)

boubaker (approved):

PR #794, opened 2021-06-24
Task-45959 : AdminMFA add the two factor authentication page under data leak prevention

ahmadmido:
This PR will add the two-factor authentication page under data leak prevention.


extension/war/src/main/resources/locale/navigation/group/platform/administrators_fr.properties
```diff
@@ -32,3 +32,4 @@ administration.notification=Notifications
 administration.spaces=Gestion des Espaces
 administration.spacesTemplates=Mod\u00E8les
 administration.spacesAdministration=G\u00E9rer les espaces
+administration.multifactorAuthentication=Authentification multi-factor
```
rdenarie:
```suggestion
administration.multifactorAuthentication=Authentification multi-facteur
```

rdenarie (changes requested):

rdenarie (changes requested):
There is an empty file : multifactorAuthenticationServices.js 
can you check it ?

In addition, as you add new properties files, you should add it in https://github.com/Meeds-io/social/blob/develop/translations.properties to push it in crowdin

rdenarie (approved):

PR #805, opened 2021-06-30
Task46338 : Add the groups' Ids on users profile

ahmadmido:
This PR will add a new property groups ids on the user profile which contains all users groups ids.


rdenarie (changes requested):
Fix code smells from sonar

PR #809, opened 2021-07-05
MAINT-38140 : Add the aboutMe profile property to the elastic search indexing profile mapping

hakermi:
Add the aboutMe profile property to the elastic search indexing profile mapping


extension/war/src/main/webapp/WEB-INF/conf/social-extension/social/indexing-configuration.xml
```diff
@@ -38,7 +38,7 @@
           <property name="type" value="profile"/>
           <property name="enable" value="${exo.unified-search.connector.people.enable:true}"/>
           <property name="titleField" value="name"/>
-          <property name="searchFields" value="name,firstName,lastName,position,skills"/>
+          <property name="searchFields" value="name,firstName,lastName,position,skills,aboutMe"/>
```
plamarque, line 41: Don't we need an upgrade plugin to reindex old profiles ?
CC @ahamdi

hakermi: it's already handled by the ProfileIndexingServiceConnector , it' updating the indexes of the old profiles with the new mapping 
cc @plamarque

extension/war/src/main/webapp/WEB-INF/conf/social-extension/social/indexing-configuration.xml
```diff
@@ -127,8 +127,8 @@
         <properties-param>
           <name>constructor.params</name>
           <property name="index_alias" value="profile_alias"/>
-          <property name="index_previous" value="profile"/>
-          <property name="index_current" value="profile_v2"/>
+          <property name="index_previous" value="profile_v2"/>
+          <property name="index_current" value="profile_v3"/>
```
ahamdi: Let's avoid this, when a user has an updat on his profile, the profile will be indexed automatically on user login. All active users will have their profiles reidexed for 6.2

ahamdi: Please makje sure to announce this so that it is considered when the fix is delivered.

hakermi: done

ahamdi (approved):
LGTM

PR #814, opened 2021-07-06
MAINT-46589 : Add Mention Processor and Html sanitize processor to the MESSAGE template param used in the activity with attached file

hakermi:
This PR should fix user mention in the activities with attached files and the activities with attached files + link preview and the display of the emojis in activities description

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/processor/MentionsProcessor.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/processor/MentionsProcessor.java
@@ -23,6 +23,7 @@ import java.util.regex.Pattern;
 
 import org.apache.commons.lang.StringEscapeUtils;
 
+import org.exoplatform.commons.utils.CommonsUtils;
 import org.exoplatform.container.xml.InitParams;
 import org.exoplatform.portal.config.UserPortalConfigService;
 import org.exoplatform.portal.webui.util.Util;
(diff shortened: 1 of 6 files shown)


component/core/src/main/java/org/exoplatform/social/core/processor/MentionsProcessor.java
```diff
@@ -41,6 +42,8 @@
 public class MentionsProcessor extends BaseActivityProcessorPlugin {
 
   private static final Pattern pattern = Pattern.compile("@([^\\s]+)|@([^\\s]+)$");
+  private static final String MESSAGE_PARAM = "MESSAGE";
```
ahamdi: COuld you move this variable to a common place, check if it is not already available

boubaker (dismissed):
It does work now on new AS, knowing that the old AS will disappear for 6.2, I think this should be committed and tested for 6.1.x only. WDYT ?

hakermi:
> It does work now on new AS, knowing that the old AS will disappear for 6.2, I think this should be committed and tested for 6.1.x only. WDYT ?

Agreed.
fyi @mzorai 
cc @ahamdi WDYT?

ahamdi:
@boubaker Yes that was what we agreed on Maintenance team 2 days ago. @hakermi please update the PR for 6.1

hakermi:
> @boubaker Yes that was what we agreed on Maintenance team 2 days ago. @hakermi please update the PR for 6.1

Done

ahamdi (approved):

component/core/src/main/java/org/exoplatform/social/core/processor/MentionsProcessor.java
hakermi: done

PR #821, opened 2021-07-07
Task-46613: Scan QR code to connect

NaderSai:


webapp/portlet/src/main/webapp/mobile-settings/components/MobileSettingsWindow.vue
```diff
@@ -51,14 +51,24 @@
...
-      </div>
+      <div class="mx-16" id="qrcode"></div>
     </div>
   </v-card>
-</template>
\ No newline at end of file
+</template>
+
+<script>
+import * as QRCode from 'easyqrcodejs';
```
ahamdi: We need to use AMD to inject the qrCode library

ahamdi (changes requested):

ahamdi (approved):

PR #868, opened 2021-08-02
TASK-40995:External user flag is still displayed next to the user fullname

GouadriaHanen:
user && user.external was treated as the user.external is boolean the case was that user.external is string 
fixed by comparing user.external with string value


ahamdi (commented):
could you do the same for the enabled property.

GouadriaHanen:
> could you do the same for the enabled property.

enabled is already boolean!

ahamdi:
I was not clear, sorry, I meant extracting the label calculation into a computed property

GouadriaHanen:
> I was not clear, sorry, I meant extracting the label calculation into a computed property

done!

ahamdi (approved):

PR #893, opened 2021-08-06
Task-47662 : Wrong color of Update button is displayed when editing a post  in Streams page

ahmadmido:
Task-47662: Before this fix, when the user edits a post from the activity stream page the update button is disabled, this fix fix will ensure that button is still enabled when the user tries to edit the post by fixing the method activityBodyEdited.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-composer-app/components/ExoActivityComposer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-composer-app/components/ExoActivityComposer.vue
@@ -195,7 +195,7 @@ export default {
       }).reduce((a, b) => a + b, 0);
     },
     activityBodyEdited: function() {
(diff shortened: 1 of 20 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-composer-app/components/ExoActivityComposer.vue
```diff
@@ -369,7 +369,7 @@ export default {
     },
     cleanInitialActivityBody: function()  {
       const div = document.createElement('div');
-      div.innerHTML = this.activityBody;
+      div.innerHTML = this.message;
```
boubaker, line 372: @ahmadmido I have fixed it in this PR : https://github.com/Meeds-io/social/pull/899/commits/74dbbe653411cf2244846eb057cdb7eca4d28cdf
You can close this one, thanks.

boubaker (changes requested):

PR #911, opened 2021-08-11
MAINT-36782: Fix space redactors filter when user role is added from group management

hakermi:
Fix space redactors filter when user role is added from group management

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/listeners/SocialMembershipListenerImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/listeners/SocialMembershipListenerImpl.java
@@ -118,6 +118,11 @@ public class SocialMembershipListenerImpl extends MembershipEventListener {
             return;
           }
           spaceService.addMember(space, userName);
+        } else if (SpaceUtils.REDACTOR.equalsIgnoreCase(m.getMembershipType())) {
+          if (spaceService.isRedactor(space, userName)) {
+            return;
+          }
+          spaceService.addRedactor(space, userName);
         }
         
         //Refresh GroupNavigation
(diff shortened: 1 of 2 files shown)


component/core/src/main/java/org/exoplatform/social/core/listeners/SocialMembershipListenerImpl.java
```diff
@@ -118,6 +118,10 @@ public void postSave(Membership m, boolean isNew) throws Exception {
             return;
           }
           spaceService.addMember(space, userName);
+        } else if (SpaceUtils.REDACTOR.equalsIgnoreCase(m.getMembershipType())) {
+          if (spaceService.isRedactor(space, userName))
+            return;
+          spaceService.addRedactor(space, userName);
```
hbenali:
```suggestion
        } else if (SpaceUtils.REDACTOR.equalsIgnoreCase(m.getMembershipType())) {
          if (spaceService.isRedactor(space, userName)) {
            return;
          }    
          spaceService.addRedactor(space, userName);
```

hbenali (changes requested):
Please add a unit test case.

component/core/src/main/java/org/exoplatform/social/core/listeners/SocialMembershipListenerImpl.java
hakermi: does this make difference ?

hbenali: No change, but I suggest being in harmony with the above implementation.

hbenali (approved):

PR #923, opened 2021-08-12
TASK-45845 Fix Who is online caching switch displayed space (#884)

mzorai:
Prior to this change, the Who Is Online Widget uses a common cache independently from the page where it's displayed. Thus, the list of displayed users in a space, for this widget, will come from the last display made of it (it can be displayed last time in a stream of another space in the same browser, or from stream page). This will ensure to use the Skeleton V3 cache key to introduce spaceId into it, so that the display of the widget will depends on current page's space identifier (as made for all other widgets)


boubaker:
It's already applied as change in fact

PR #932, opened 2021-08-13
Task-44094 : User Type on UserManegement search is not updated when a…

rdenarie:
…n internal user become external (#705)

Before this fix, when we add an internal user to an external group, the profile property external isn't updated.
This fix will ensure that when we add users to an external group, the profile external property is changed to true.


azayati (approved):
LGTM

PR #964, opened 2021-08-24
Task-48190 : The switch to an article button shouldn't be displayed when exceed 1300 chars in stream main page

ahmadmido:
Issue: When adding a post with exceeding 1300 chars, the switch to an article button is displayed on mainstream and space stream.
Solution: I have added a check to verify if is mainstream or space stream, and I display the switch to an article button only on space stream.


rdenarie (changes requested):
In social in webapp/src/main/webapp/news-extensions/extensions.js we have something speaking about enabling or not activity-composer-action "switch to an article" : 

>         switchToArticleActivityComposerPlugin.enabled = newsActivityComposerPlugin.enabled = eXo.env.portal.spaceId && canCreateNews;

As the action comes from news, I think that the fix should occurs in news modules, and we should check the condition "enabled" here.

rdenarie (approved):

PR #967, opened 2021-08-26
Task-48242: Make sure to display disabled and external users labels in user profile popup.

mdiha:
When hovering over disabled or external user profile the displayed popup shows the indication between parentheses as undefined.
These changes make sure to add lables for disabled and external users to be displayed in the popup.

### DIFF
--- a/webapp/portlet/src/main/webapp/js/eXo/social/webui/UIUserProfile.js
+++ b/webapp/portlet/src/main/webapp/js/eXo/social/webui/UIUserProfile.js
@@ -118,7 +118,10 @@
               //
               UserProfile.loadingProfileSize(componentId);
               //
-              uiProfile.initUserProfilePopup(componentId, {});
+              uiProfile.initUserProfilePopup(componentId, {
+                'External': this.$t('UserProfilePopup.label.External'),
+                'Disabled': this.$t('UserProfilePopup.label.Disabled')
+              });
             } else {
               container.hide();
               portlet.find('div.borderContainer.empty:first').show();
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/js/eXo/social/webui/UIUserProfile.js
```diff
@@ -118,7 +118,10 @@
               //
               UserProfile.loadingProfileSize(componentId);
               //
-              uiProfile.initUserProfilePopup(componentId, {});
+              uiProfile.initUserProfilePopup(componentId, {
+                'External': this.$t('UserProfilePopup.label.External'),
+                'Disabled': this.$t('UserProfilePopup.label.Disabled')
+              });
```
rdenarie, line 124: Instead of duplicate this code, I think we should put it directly in UIProfile.js.
During initialization, there labels which are defined : 
`function($) { 
var UIProfile = {
  labels: {},
  KEYS : {
    ENTER : 13
  },`

We could add theses default labels here, so that it will work everywhere

rdenarie (changes requested):

webapp/portlet/src/main/webapp/js/eXo/social/webui/UIUserProfile.js
mdiha: Yes you are right.

rdenarie (approved):
LGTM.
Adding label in UIProfile is more complicated than the proposed solution

PR #968, opened 2021-08-27
Task-48341 : Externals users are displayed on the unified search

rdenarie:
This commit should also fix
Task-47922 : External users are displayed in people list
Task-47770 : Wrong behaviors when search with an advanced filter applied

Before this fix, when doing an es request containing a should option, if none of should option match, the condition stay true.
This fix add the option minimum_should_match to 1, so that, at least one should option is true.
This modification applies on requests about external/internal users, connected/not connected users, and request about enrollment status

In addition the request to get notEnrolled users have a syntax error because of multiple must_not part.
This change reworks this request to make a correct one

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
@@ -71,7 +71,8 @@ public class ProfileSearchConnector {
                                      long offset,
                                      long limit) {
     if(identity == null && filter.getViewerIdentity() != null) {
-      identity = filter.getViewerIdentity();}
+      identity = filter.getViewerIdentity();
+    }
     String esQuery = buildQueryStatement(identity, filter, type, offset, limit);
     String jsonResponse = this.client.sendRequest(esQuery, this.index);
     return buildResult(jsonResponse);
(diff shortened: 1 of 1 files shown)


ahmadmido (dismissed):

component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
```diff
@@ -71,8 +71,7 @@ public ProfileSearchConnector(InitParams initParams, ElasticSearchingClient clie
                                      long offset,
                                      long limit) {
     if(identity == null && filter.getViewerIdentity() != null) {
-      identity = filter.getViewerIdentity();
-    }
+      identity = filter.getViewerIdentity();}
```
Rdidmohamedelhadi: Fix indentation, please

Rdidmohamedelhadi (changes requested):

component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
rdenarie: done, thanks

Rdidmohamedelhadi (approved):
LGTM

PR #969, opened 2021-08-27
Task-48221 : Download option is still displayed after suspend temporarily download documents on transfer rules page.

ahmadmido:
Issue: Before this fix, when suspending temporarily download documents on the transfer rules page, the download action (download or downloadAll)is still displayed on the activity stream (3 dots button).
Solution: Check the transfer rules download action status and filter enabled action according to this value, if suspend is enabled the download actions are displayed(download or downloadAll) otherwise isn't displayed.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityHeadMenu.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityHeadMenu.vue
@@ -52,13 +52,14 @@ export default {
   data: () => ({
     menu: false,
     loading: false,
+    downloadAllowed: true,
   }),
   computed: {
     menuButtonStyle() {
       return this.$vuetify && this.$vuetify.rtl && 'top:8px;left:4px;' || 'top:8px;right:4px;';
     },
     enabledActions() {
-      return this.activityActions && Object.values(this.activityActions).filter(action => action.isEnabled && action.id && action.click && action.isEnabled(this.activity, this.activityTypeExtension));
+      return this.activityActions && Object.values(this.activityActions).filter(action => action.isEnabled && action.id && action.click && action.isEnabled(this.activity, this.activityTypeExtension)).filter(action => !(this.downloadAllowed  && (action.id === 'download'|| action.id === 'downloadAll'))) ;
     },
   },
   created() {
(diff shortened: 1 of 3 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityHeadMenu.vue
```diff
@@ -52,14 +52,18 @@ export default {
...
     menu: false,
     loading: false,
+    transferRulesDownloadStatus: false,
   }),
   computed: {
     menuButtonStyle() {
       return this.$vuetify && this.$vuetify.rtl && 'top:8px;left:4px;' || 'top:8px;right:4px;';
     },
-    enabledActions() {
+    actions() {
```
rdenarie: Can you explain what is "actions" stand for ? I'm not sure to understand.
Thanks

ahmadmido: > Can you explain what is "actions" stand for ? I'm not sure to understand.
> Thanks

Actions are all  the actions available when suspend document is disabled and on enabledActions we do a filter to check if all actions of download are enabled or not

rdenarie: can you show me where "actions()" functione is used ?

rdenarie (approved):

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityHeadMenu.vue
```diff
@@ -70,6 +71,10 @@ export default {
         }, 200);
       }
     });
+    this.$transferRulesService.getTransfertRulesDownloadDocumentStatus().then(
+      (data) => {
+        this.downloadAllowed = data.downloadDocumentStatus === 'true';
+      });
```
boubaker, line 77: @ahmadmido please rework this PR to avoid introducing `download` term in Social, while the download buttons are defined in ecms.
 you can instead of adding here a check, adding this code snippet [here](https://github.com/exoplatform/ecms/blob/6.2.0-RC04/apps/portlet-documents/src/main/webapp/vue-app/attachment/extensions.js#L56-L81) where the download buttons are defined :


```javascript
getTransfertRulesDownloadDocumentStatus()
  .then(data => {
    if (data && data.downloadDocumentStatus === 'true') {
      extensionRegistry.registerExtension('activity', 'action', Object.assign({
        id: 'download',
        ....
      extensionRegistry.registerExtension('activity', 'action', Object.assign({
        id: 'downloadAll',
        ....
    }
  });

```

PR #970, opened 2021-08-31
TASK-48336 : Problems about indication(External) when hovering a mentioned external user in a comment

nidhalnidhal:
Before this fix , when mentioning an external user in a comment the indication external appears in blue color instead of grey color . This was fixed by applying the correct class and style to every mentioned user when processing the comment  message .

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
@@ -170,7 +170,7 @@ public class LinkProvider {
                 .append(buildProfileUri(identity.getRemoteId(), null, portalOwner)).append("\" target=\"_parent\">")
                 .append(StringEscapeUtils.escapeHtml(identity.getProfile().getFullName()));
     if(identity.getProfile().getProperty("external") != null && identity.getProfile().getProperty("external").equals("true")){
-      profileLink = profileLink.append("<span \" class=\"externalTagClass\">").append(" (").append(getResourceBundleLabel(new Locale(lang), "external.label.tag")).append(")").append("</span>");
+      profileLink = profileLink.append("<span \" class=\"externalFlagClass\">").append(" (").append(getResourceBundleLabel(new Locale(lang), "external.label.tag")).append(")").append("</span>");
     }
     return profileLink.append("</a>").toString();
   }


component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
```diff
@@ -170,7 +170,7 @@ public static String getProfileLink(final String username, final String portalOw
                 .append(buildProfileUri(identity.getRemoteId(), null, portalOwner)).append("\" target=\"_parent\">")
                 .append(StringEscapeUtils.escapeHtml(identity.getProfile().getFullName()));
     if(identity.getProfile().getProperty("external") != null && identity.getProfile().getProperty("external").equals("true")){
-      profileLink = profileLink.append("<span \" class=\"externalTagClass\">").append(" (").append(getResourceBundleLabel(new Locale(lang), "external.label.tag")).append(")").append("</span>");
+      profileLink = profileLink.append("<span \" class=\"externalFlagClass\" style=\"color:#999999\">").append(" (").append(getResourceBundleLabel(new Locale(lang), "external.label.tag")).append(")").append("</span>");
```
ahamdi: Please add the style to the CSS class, we should not add styles directly on HTML code.

ahamdi (changes requested):

component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
nidhalnidhal: i tried many solutions but it seems that the vue component is not recognizing the class "externalTagClass" or "externalflagclass" because it's injected from the java backend code as belonging to  the entire mention message, so the mention message style is applied.

ahamdi: We need to have a better way to fix this, no style to be put directly in the DOM

ahamdi (dismissed):

component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
AzmiTouil: @nidhalnidhal  you should force style color in the CSS class

AzmiTouil (changes requested):

AzmiTouil (dismissed):

AzmiTouil (approved):

PR #977, opened 2021-09-03
TASK-48563 Fix High Memory Consumption when reindexing User Profile

boubaker:
Due to recent change made in https://github.com/Meeds-io/social/commit/c146347542fbc4f9519b0166df65a5d59be45969, a high memory consumption has been observed on production. In fact, assuming that a user can have tens of thousands of users as 'relationship', retrieving a list of connection IDs in a list of String will consume much more Memory than appending the list of Ids in one single String. Thus this fix will revert to compute the IDs inside a String instead (Old behavior). In addition two methods has been deprecated knowing that it's not used, in order to avoid indexing & searching to/from ES the pending (incoming & outgoing) relationships.


bourasom (approved):
LGTM

PR #981, opened 2021-09-06
46374: Make sure that returned users profiles contains accentueted characters if present.

mdiha:
When searching for users identities via the elastic search connector the user's name, last name and full name aren't accentueted.
These changes make sure to return the correct user's name, last name and full name with accentueted characters if present by loadig user's profiles from the identity manager.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
@@ -683,7 +683,11 @@ public class RDBMSIdentityStorageImpl implements IdentityStorage {
       identity = profileFilter.getViewerIdentity();
     }
     if (OrganizationIdentityProvider.NAME.equals(providerId)) {
-      return getProfileSearchConnector().search(identity, profileFilter, type, offset, limit);
+      List<Identity> identities = getProfileSearchConnector().search(identity, profileFilter, type, offset, limit);
+      identities.stream().forEach(identity1 -> {
+        identity1.setProfile(loadProfile(identity1.getProfile()));
+      });
+      return identities;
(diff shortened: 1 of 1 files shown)


component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
```diff
@@ -683,7 +683,11 @@ public void processEnabledIdentity(Identity identity, boolean isEnable) {
       identity = profileFilter.getViewerIdentity();
     }
     if (OrganizationIdentityProvider.NAME.equals(providerId)) {
-      return getProfileSearchConnector().search(identity, profileFilter, type, offset, limit);
+      List<Identity> identities = getProfileSearchConnector().search(identity, profileFilter, type, offset, limit);
```
AzmiTouil, line 686: IMO you should use forceLoadOrReloadProfile param to not force the profile loading in all cases.

AzmiTouil (changes requested):

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
ahamdi: we should not query each identity from database, this will cause performance regression. Are firstname/lastname stored in ES without accents ?

ahamdi (changes requested):

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
mdiha: No in ES nothing is stored with accents.

PR #984, opened 2021-09-07
Task-48597: Fix space popover without actions for managers

NaderSai:
Problem: Space popover appears without actions only for managers.
How it was solved: by taking the "1" value in the condition json.manager >1 to build an action div because the build of the popup requires action different from null.

### DIFF
--- a/webapp/portlet/src/main/webapp/js/eXo/UISpaceInfoPopup.js
+++ b/webapp/portlet/src/main/webapp/js/eXo/UISpaceInfoPopup.js
@@ -307,7 +307,7 @@
 
                     tiptip_content.empty();
 
-                    if ((isManager && json.manager =>1)|| (!isManager && isMember)) {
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/js/eXo/UISpaceInfoPopup.js
```diff
@@ -307,7 +307,7 @@
 
                     tiptip_content.empty();
 
-                    if ((isManager && json.manager >1)|| (!isManager && isMember)) {
+                    if ((isManager && json.manager =>1)|| (!isManager && isMember)) {
```
AzmiTouil:
```suggestion
                    if ((isManager && json.manager >= 1) || (!isManager && isMember)) {
```

AzmiTouil (changes requested):

AzmiTouil (approved):
LGTM

PR #1000, opened 2021-09-14
Task 48683: Fix space and user profile picture doesn't updated on stream page

NaderSai:
Problem: space and user profile picture doesn't updated on stream page.
How it was solved: by deleting the cache once the user update the avatar.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/profile-header/components/ProfileHeaderAvatar.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/profile-header/components/ProfileHeaderAvatar.vue
@@ -96,7 +96,18 @@ export default {
               this.$emit('input', uploadId);
             }
           })
-          .then(() => this.$emit('refresh'))
+          .then(() => {this.$emit('refresh');
+            if (navigator.serviceWorker && window.caches) {
+              window.caches.open('portal-pwa-resources-image')
+                .then(cache => {
+                  if (cache) {
+                    window.setTimeout(() => {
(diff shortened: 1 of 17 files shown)


AzmiTouil (approved):
LGTM

webapp/portlet/src/main/webapp/vue-apps/profile-header/components/ProfileHeaderAvatar.vue
```diff
@@ -96,7 +96,18 @@ export default {
               this.$emit('input', uploadId);
             }
           })
-          .then(() => this.$emit('refresh'))
+          .then(() => {this.$emit('refresh');
+            if (navigator.serviceWorker && window.caches) {
```
boubaker, line 100:
```suggestion
            if (navigator.serviceWorker && window.caches) {
```
This is just a workaround and we shouldn't manage service worker lifecycle inside apps

To make a better fix, you can delete Service Worker on 6.2.x, in fact, it's not used anymore. To do it, you can, as a start, just change default value of :
https://github.com/Meeds-io/gatein-portal/blob/811013bc654cd35f7100049d8f0275862be0bc35/web/portal/src/main/webapp/WEB-INF/conf/pwa/pwa-configuration.xml#L11

to make it `false` instead of `true`

boubaker (changes requested):

webapp/portlet/src/main/webapp/vue-apps/profile-header/components/ProfileHeaderAvatar.vue
NaderSai: should i delete the workaround or not?

PR #1001, opened 2021-09-14
Task-47412: Allow all space members to see space redactors by the redactors filter of space members application

azayati:
Prior to this change, only space managers can see space redactors using the redactors filter of space members application.
After this change, any space member have access to the redactors filter of space members application in order to see space redactors.


hassenelkamel-exo (approved):
LGTM

PR #1011, opened 2021-09-22
Task-49104 Implement Metadata API and Activity Favorites as first example

boubaker:


component/service/src/main/java/org/exoplatform/social/rest/impl/metadata/FavoriteRest.java
```diff
@@ -0,0 +1,165 @@
...
+import org.exoplatform.services.log.Log;
+import org.exoplatform.services.rest.resource.ResourceContainer;
+import org.exoplatform.social.common.ObjectAlreadyExistsException;
+import org.exoplatform.social.metadata.favorite.FavoriteService;
+import org.exoplatform.social.rest.api.RestUtils;
+import org.exoplatform.social.service.rest.api.VersionResources;
+
+import io.swagger.annotations.*;
+
+@Path(VersionResources.VERSION_ONE + "/social/favorites")
```
plamarque, line 23: couldn't we have /favorites without prefix? it's not related to social fesures amd supposed to be transversal even for obbjects outside social

plamarque (commented):
i'd prefer /favorites instead of /social/favorites to make it clear that it's a transversal service (not tied to "social" features)

component/service/src/main/java/org/exoplatform/social/rest/impl/metadata/FavoriteRest.java
boubaker: Unfortunately we have already a `favorites` endpoint : https://github.com/exoplatform/ecms/blob/develop/core/connector/src/main/java/org/exoplatform/wcm/connector/collaboration/FavoriteRESTService.java#L37

Besides, adding a parent category for REST endpoints ( like 'social' here to reference the module ), will give us more flexibility for routing rules, such as in Microservices Architecture (where we will be able to route all 'social' endpoints to the Monolith, a dedicated Microservice, applying a specific rule ...)

PR #1016, opened 2021-10-07
Task-49264: Space members shouldn't be able to create projects

AzmiTouil:
Prior this change, every login/logout with a user we need to empty the browser cache to display or hide Add project button, The problem is that when the authenticated User has been changed, the cache does not updated.
We should integrate the authenticated user into the cache calculation to get the correct result.


ahamdi:
Do you mean that the restriction is already implemented but there was a cache that was not updated that made it possible for users to create projects under spaces? Could you add more details on the problem and why will it be fixed with this change ?

ahamdi (commented):
Do you mean that the restriction is already implemented but there was a cache that was not updated that made it possible for users to create projects under spaces? Could you add more details on the problem and why will it be fixed with this change ?

ahamdi (approved):

PR #1019, opened 2021-10-08
Task-49602: Fix the problem of hidden datepicker when implement stepper

hassenelkamel-exo:
Prior to this fix, when implementing stepper in `ExoScheduleNewsDrawer.vue`, we noticed that the datepicker is not well displayed. After this change, the problem is fixed by disabling attach property.


azayati (dismissed):
LGTM after QA validation

azayati (dismissed):
LGTM

azayati (approved):
LGTM

PR #1020, opened 2021-10-12
QWIN-42333: Implement a new space Top banner Logo popover

hakermi:
Implement a new space Top banner Logo popover


webapp/portlet/src/main/webapp/WEB-INF/jsp/bannerLogo.jsp
```diff
@@ -35,13 +36,20 @@
     logoTitle = space.getDisplayName();
     String permanentSpaceName = space.getGroupId().split("/")[2];
     portalPath = "/portal/g/:spaces:" + permanentSpaceName + "/" + space.getPrettyName();
-    imageClass="spaceAvatar";
-    titleClass = "space";
+    membersNumber = space.getMembers().length;
```
ahamdi, line 48: This will be a performance killer. Could you find another way to do it ?

webapp/portlet/src/main/webapp/vue-apps/common/initComponents.js
```diff
@@ -19,7 +19,6 @@ import ExoGroupSuggester from './components/ExoGroupSuggester.vue';
 import CardCarousel from './components/CardCarousel.vue';
 import DrawersOverlay from './components/DrawersOverlay.vue';
 import ActivityShareDrawer from './components/ActivityShareDrawer.vue';
-
```
ahamdi: unneeded change

ahamdi (changes requested):

webapp/portlet/src/main/webapp/WEB-INF/jsp/bannerLogo.jsp
hakermi: the `space.getMmebers` is an array of String (usernames of members) , when call that method no rest call will be invoked and no computed functions will be called. all the **model** (space data) already loaded in the cache of the server, and  i just access and use these already loaded data. and same for `space.getManagers`.

webapp/portlet/src/main/webapp/vue-apps/common/initComponents.js
hakermi: changes reverted

webapp/portlet/src/main/webapp/WEB-INF/jsp/bannerLogo.jsp
ahamdi: Did you test it with a crowded space ? 5000 members for example ? will it be quick or will it take time to create an array , fill it with usernames to just count the number of members ?

hakermi: > Did you test it with a crowded space ? 5000 members for example ? will it be quick or will it take time to create an array , fill it with usernames to just count the number of members ?

The array is already created and loaded before opening the popover, and the array length depends on the JAVA mechanism is just a fieled in the array object, doesn't need a calculation and it's a super fast operation.

and there is no other way to do this unless if you want me to change this by a rest call of members-count and update the spaceDao and the storage and add a query of members count ?

ahamdi: I am for removing the getMembers() and get the count of members in the DTO. Getting the members info could be done with search queries with pagination.
We can let it as is now , we will do it in another task

ahamdi (approved):

AzmiTouil (approved):
LGTM

PR #1025, opened 2021-10-14
Task-50193: Add web contributor group and publisher role on social

azayati:
Prior to this change, web contributor group and publisher role are not configured on social thus not created on Meeds, after this change web contributor group and publisher role are well created on Meeds which allows publisher on web contributor group to write a post on a redactional space


samehaidi (approved):

rdenarie:
@azayati must this pr be closed ?

PR #1044, opened 2021-10-25
Task-50098 : Article is not raised to the top of the space when it is edited with the option update and post.

ahmadmido:
Prior to this fix, managers of space can't edit news activity on redactional or normal space, and the article is not raised to the top of the activity stream.
This fix will make sure that the manager of space has the possibility to edit news articles, and the activity is raised to the top of the activity stream.

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
@@ -386,6 +386,12 @@ public class ActivityRestResourcesV1 implements ResourceContainer {
                                      @QueryParam("expand")
                                      String expand,
                                      @ApiParam(
+                                         value = "Is current activity editable",
+                                         required = false
+                                     )
+                                     @QueryParam("isEditabled")
+                                     boolean isEditabled,
+                                     @ApiParam(
                                          value = "Activity object to be updated, ex: <br/>{<br/>\"title\" : \"My activity\"<br/>}",
                                          required = true
                                      )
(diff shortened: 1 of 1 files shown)


component/core/src/main/java/org/exoplatform/social/core/manager/ActivityManagerImpl.java
```diff
@@ -708,6 +710,12 @@ public boolean isActivityEditable(ExoSocialActivity activity, org.exoplatform.se
         enableEdit = enableEditActivity;
       }
       Identity identity = identityManager.getOrCreateIdentity(OrganizationIdentityProvider.NAME, viewer.getUserId());
+      if(activity.getType() != null && activity.getType().equals(NEWS_TYPE)) {
```
azayati: mentioning news on social code is a kind of forbidden modifications, we should find another smart solution

azayati (changes requested):
See review

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
```diff
@@ -400,7 +406,7 @@ public Response updateActivityById(
     Identity currentUser = identityManager.getOrCreateIdentity(OrganizationIdentityProvider.NAME, authenticatedUser);
 
     ExoSocialActivity activity = activityManager.getActivity(activityId);
-    if (!activityManager.isActivityEditable(activity, authenticatedUserIdentity)) {
+    if (!isEditabled && !activityManager.isActivityEditable(activity, authenticatedUserIdentity)) {
```
boubaker, line 409: Security breach here, please delete PR as agreed with @azayati

boubaker (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
azayati: @ahmadmido as discussed the PR will be reworked from news part only

PR #1051, opened 2021-10-26
Task-47724: Fix connections number is not displayed correctly

NaderSai:
Prior this change, the number of connections in the profile widget is not updated when a new connection is added, this due to that the identity cache is not updated in this case.

Fix : Clear identity cache when save or delete a relationShip.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/storage/cache/CachedRelationshipStorage.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/storage/cache/CachedRelationshipStorage.java
@@ -23,12 +23,10 @@ import java.util.Map.Entry;
 import org.apache.commons.lang.StringUtils;
 
 import org.exoplatform.commons.cache.future.FutureExoCache;
-import org.exoplatform.container.PortalContainer;
 import org.exoplatform.services.cache.ExoCache;
 import org.exoplatform.services.log.ExoLogger;
 import org.exoplatform.services.log.Log;
 import org.exoplatform.social.core.identity.model.Identity;
-import org.exoplatform.social.core.identity.provider.OrganizationIdentityProvider;
 import org.exoplatform.social.core.jpa.storage.RDBMSRelationshipStorageImpl;
 import org.exoplatform.social.core.profile.ProfileFilter;
 import org.exoplatform.social.core.relationship.model.Relationship;
@@ -197,18 +195,13 @@ public class CachedRelationshipStorage implements RelationshipStorage {
   }
 
   /**
-   * Remove identity and profile cache.
+   * Remove identity cache.
    *
    * @param identity
    */
   private void removeIdentityCache(Identity identity) {
-    CachedIdentityStorage cachedIdentityStorage =
-                                                (CachedIdentityStorage) PortalContainer.getInstance()
-                                                                                       .getComponentInstanceOfType(IdentityStorage.class);
-    Identity cachedIdentity = cachedIdentityStorage.findIdentity(OrganizationIdentityProvider.NAME, identity.getRemoteId());
-
-    if (cachedIdentity != null) {
-      cachedIdentityStorage.clearIdentityCache(cachedIdentity, false);
+    if (identityStorage instanceof CachedIdentityStorage) {
+      ((CachedIdentityStorage) identityStorage).clearIdentityCache(identity, false);
     }
   }
 
--- a/component/core/src/main/java/org/exoplatform/social/core/storage/cache/CachedRelationshipStorage.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/storage/cache/CachedRelationshipStorage.java
@@ -208,7 +208,7 @@ public class CachedRelationshipStorage implements RelationshipStorage {
     Identity cachedIdentity = cachedIdentityStorage.findIdentity(OrganizationIdentityProvider.NAME, identity.getRemoteId());
 
     if (cachedIdentity != null) {
-      cachedIdentityStorage.clearIdentityCached(cachedIdentity, identity.getRemoteId());
+      cachedIdentityStorage.clearIdentityCache(cachedIdentity, false);
     }
   }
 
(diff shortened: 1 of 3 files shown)


AzmiTouil:
IMO, we have to clear the identity cache when updating the relationship.
cc @NaderSai

component/core/src/main/java/org/exoplatform/social/core/manager/RelationshipManagerImpl.java
```diff
@@ -468,6 +470,8 @@ public void confirm(Identity invitedIdentity, Identity invitingIdentity) {
    */
   public void delete(Relationship existingRelationship) {
     storage.removeRelationship(existingRelationship);
+    identityStorage.updateIdentity(existingRelationship.getSender());
+    identityStorage.updateIdentity(existingRelationship.getReceiver());
```
boubaker: you mean clearing cache instead of making a DB call to update two identities ?

boubaker (changes requested):

component/core/src/main/java/org/exoplatform/social/core/manager/RelationshipManagerImpl.java
```diff
@@ -48,6 +49,8 @@
 
   protected IdentityStorage identityStorage;
 
+  protected CachedIdentityStorage cachedIdentityStorage;
```
boubaker: You have already the declaration of `IdentityStorage identityStorage`, no need to retrieve it again, you just need to test if it's of type `CachedIdentityStorage`.

BTW, please give a good title and purpose of your commit and avoid messages such as `commit 2`

NaderSai: okay well noted

component/core/src/main/java/org/exoplatform/social/core/manager/RelationshipManagerImpl.java
```diff
@@ -459,6 +460,10 @@ public void confirm(Identity invitedIdentity, Identity invitingIdentity) {
     if (relationship != null && relationship.getStatus() == Relationship.Type.PENDING) {
       relationship.setStatus(Relationship.Type.CONFIRMED);
       this.update(relationship);
+      if (identityStorage instanceof CachedIdentityStorage){
+        identityStorage.updateIdentity(invitedIdentity);
```
boubaker: you kept the fact that we update the identity on RDBMS to clear the cache

boubaker:
BTW, can you please add a Unit Test on REST endpoint to return 304 when no modifications on relationships and 200 when a relationship has changed + the update on connectionsInCommon attribute.

component/core/src/main/java/org/exoplatform/social/core/manager/RelationshipManagerImpl.java
NaderSai: IMO we can do the clearing of cache directly from the CachedRaltionshipStorage after saving new relationship.
WDYT?

component/core/src/main/java/org/exoplatform/social/core/storage/cache/CachedRelationshipStorage.java
```diff
@@ -194,6 +196,22 @@ private SuggestionsData buildIdMap(Map<Identity, Integer> map) {
...
+   * @param identity
+   */
+  private void removeIdentityCache(Identity identity) {
+    CachedIdentityStorage cachedIdentityStorage =
+                                                (CachedIdentityStorage) PortalContainer.getInstance()
+                                                                                       .getComponentInstanceOfType(IdentityStorage.class);
+    Identity cachedIdentity = cachedIdentityStorage.findIdentity(OrganizationIdentityProvider.NAME, identity.getRemoteId());
+
+    if (cachedIdentity != null) {
+      cachedIdentityStorage.clearIdentityCached(cachedIdentity, identity.getRemoteId());
```
boubaker: It's better to not clear the whole identities cache when the relationship changes between two users. In fact the method `cachedIdentityStorage.clearIdentityCached` will delete all user identities from cache. The clear operation is useful when the identity itself changes, but for relationship, you can just :
1. update the cache time of cached identity only
2. clear only the identity from cache

To do this, I suggest to add a new method similar to clearIdentityCached to clear the cached identity only:
```java
  public void clearIdentityCache(Identity identity, boolean clearList) {
    IdentityKey key = new IdentityKey(new Identity(identity.getId()));
    exoIdentityCache.remove(key);
    IdentityCompositeKey compositeKey = new IdentityCompositeKey(identity.getProviderId(), identity.getRemoteId());
    if (exoIdentityIndexCache.get(compositeKey) != null) {
      exoIdentityIndexCache.remove(compositeKey);
    }
    if (clearList) {
      clearCache();
    }
  }
```

AzmiTouil: Updated!

component/core/src/main/java/org/exoplatform/social/core/storage/cache/CachedRelationshipStorage.java
```diff
@@ -194,6 +196,22 @@ private SuggestionsData buildIdMap(Map<Identity, Integer> map) {
...
+   */
+  private void removeIdentityCache(Identity identity) {
+    CachedIdentityStorage cachedIdentityStorage =
+                                                (CachedIdentityStorage) PortalContainer.getInstance()
+                                                                                       .getComponentInstanceOfType(IdentityStorage.class);
+    Identity cachedIdentity = cachedIdentityStorage.findIdentity(OrganizationIdentityProvider.NAME, identity.getRemoteId());
+
+    if (cachedIdentity != null) {
+      cachedIdentityStorage.clearIdentityCache(cachedIdentity, false);
+    }
```
boubaker:
```suggestion
    if (identityStorage instanceof CachedIdentityStorage) {
      ((CachedIdentityStorage)identityStorage).clearIdentityCache(identity, false);
    }
```

boubaker (approved):

PR #1052, opened 2021-10-27
TASK-49958 : fix kudos mention not functional (#1036)

aelbenney:


component/core/src/main/java/org/exoplatform/social/core/processor/MentionsProcessor.java
```diff
@@ -68,6 +68,9 @@ public void processActivity(ExoSocialActivity activity) {
       if (templateParams.containsKey("default_title")) {
         templateParams.put("default_title", MentionUtils.substituteUsernames(portalOwner, templateParams.get("default_title")));
       }
+      if (templateParams.containsKey("kudosMessage")) {
```
boubaker, line 71: kudos addon in social :-)

boubaker (changes requested):

PR #1057, opened 2021-11-02
Task-47727: Spaces number is not displayed correctly in profile gadget in snapshot

AzmiTouil:
Prior to this change, when the user added to a space, the displayed number of spaces is not updated.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/space/SpaceUtils.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/space/SpaceUtils.java
@@ -2037,15 +2037,10 @@ public class SpaceUtils {
     return trimmed.toArray(new String[trimmed.size()]);
   }
 
-  private static void clearIdentityCaching(String username) {
+  private static void clearIdentityCaching(String remoteId) {
     ExoContainer container = ExoContainerContext.getCurrentContainer();
-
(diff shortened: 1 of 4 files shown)


component/core/src/main/java/org/exoplatform/social/core/space/SpaceUtils.java
```diff
@@ -2033,4 +2036,16 @@ public static ResourceBundleService getResourceBundleService() {
...
     return trimmed.toArray(new String[trimmed.size()]);
   }
+
+  private static void clearIdentityCaching(String username) {
+    ExoContainer container = ExoContainerContext.getCurrentContainer();
+
+    IdentityManager identityManager = container.getComponentInstanceOfType(IdentityManager.class);
+    CachedIdentityStorage cachedIdentityStorage = container.getComponentInstanceOfType(CachedIdentityStorage.class);
+
+    Identity userIdentity = identityManager.getOrCreateIdentity(OrganizationIdentityProvider.NAME, username);
```
ahamdi: Could you make org.exoplatform.social.core.storage.cache.CachedIdentityStorage#clearIdentityCached get a Long as ID instead of the whole Identity. This will help avoid reloading the identity here.

ahamdi (changes requested):

ahamdi (approved):

PR #1058, opened 2021-11-03
Task-50744: Don't allow admins to post on redactional space activity stream

hassenelkamel-exo:
Prior to this change, administrators can post on redactional space activity stream. After this change, we remove the possibility to access to composer for administrators when they are a simple space members.


azayati (approved):
LGTM after QA validation

PR #1068, opened 2021-11-09
50499 : remove access restriction from Mobile section in user settings

ahamdi:
This fix will make Mobile settings appear for all authenticated users. This settings was hidden until the release of iOS application.


AzmiTouil (approved):
LGTM

PR #1071, opened 2021-11-12
Task-50286 : Add translation extension

rdenarie:
Before this fix, when an activity action or comment is added with extensionRegistry.registerExtension, the listener configured in ActivityStream.vue is not launched, due to the fact that the listener must have one parameter for the event.
As the 3 refresh* functions have not parameter, they are never launched.

This fix modify the addListener to register an anonymous fucntion which call the register function, so that the event correctly call the refresh function


azayati (approved):
LGTM

PR #1074, opened 2021-11-15
TASK-51257 :  Redirect space wallet setting to the space's settings

GouadriaHanen:


webapp/portlet/src/main/webapp/WEB-INF/gatein-resources.xml
```diff
@@ -223,6 +223,10 @@
       <depends>
         <module>userPopupPlugin</module>
       </depends>
+      <depends>
+        <module>walletCommon</module>
```
ahamdi: No we can't do this. Wallet depends on Social, not the opposite

Rdidmohamedelhadi (approved):
LGTM

PR #1087, opened 2021-11-19
Task-47492: Badges by domain is not displayed

AzmiTouil:
Prior this change,gamification points for normal activity in space are not earned, this is due to NPE when checking activity type


hakermi (approved):
LGTM

PR #1102, opened 2021-11-24
Task-47492: Badges by domain is not displayed

mzorai:
Prior this change,gamification points for normal activity in space are not earned, this is due to NPE when checking activity type


component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
```diff
@@ -283,7 +283,7 @@ public Response postActivity(
       activity.setTitle(model.getTitle());
     }
     activity.setBody(model.getBody());
-    activity.setType(model.getType());
+    activity.setType(model.getType() != null ? model.getType() : "");
```
boubaker, line 286: This seems to be a **workaround**, the Service & Storage layers has to handle this attribute when it's null. Same for gamification which should be improved to handle when type is null.

boubaker (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
ahamdi: @mzorai @AzmiTouil  Could you take care of the comment of @boubaker  ? The fix should be done on Gamification if the NPE is thrown there.

AzmiTouil: https://github.com/Meeds-io/gamification/pull/165 cc @ahamdi

AzmiTouil: https://github.com/Meeds-io/social/pull/1107

PR #1127, opened 2021-12-02
added styles to display v-chip on mobiles with ellipsis

Amine0008:
ISSUE: in the settings page we need to display the blue label chips with ellipsis for mobile web
FIX: added custom style that will be applied to these chips only when the application is opened on a mobile browser

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/user-setting-notifications/components/UserSettingNotificationPlugin.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/user-setting-notifications/components/UserSettingNotificationPlugin.vue
@@ -20,7 +20,12 @@
         <v-list-item-title class="text-wrap">
           <template v-if="enabledDigestLabel">
             <v-chip class="ma-2" color="primary">
-              {{ enabledDigestLabel }}
+               <span v-if="isMobile" class="mobile-chip-ellipsis">
+                  {{ enabledDigestLabel }}
+              </span>
+              <span v-else>
+                {{ enabledDigestLabel }}
+              </span>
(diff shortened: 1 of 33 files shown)


webapp/portlet/src/main/webapp/vue-apps/user-setting-notifications/components/UserSettingNotificationPlugin.vue
```diff
@@ -20,7 +20,12 @@
         <v-list-item-title class="text-wrap">
           <template v-if="enabledDigestLabel">
             <v-chip class="ma-2" color="primary">
-              {{ enabledDigestLabel }}
+               <span v-if="isMobile" class="mobile-chip-ellipsis">
+                  {{ enabledDigestLabel }}
+              </span>
+              <span v-else>
+                {{ enabledDigestLabel }}
+              </span>
```
ahamdi: Following https://vuejs.org/v2/guide/class-and-style.html , this can be replacesd with
```suggestion
               <span :class="{ mobile-chip-ellipsis : isMobile }">
                  {{ enabledDigestLabel }}
              </span>
```

ahamdi: Following https://vuejs.org/v2/guide/class-and-style.html , this can be replacesd with
```suggestion
               <span :class="{ mobile-chip-ellipsis : isMobile }">
                  {{ enabledDigestLabel }}
              </span>
```

ahamdi (changes requested):

PR #1138, opened 2021-12-08
Task-51370: Analytics to know the use of unified search filtering

AzmiTouil:
Add statistics when user use the filter in the unified search

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/search/components/SearchResults.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/search/components/SearchResults.vue
@@ -290,39 +290,10 @@ export default {
         return;
       }
       if (!selectedConnector.enabled || this.connectors.length === this.enabledConnectors.length) {
-        let name;
-        switch (selectedConnector.name) {
-        case 'activity':
-          name = 'searchActivities';
-          break;
-        case 'news':
-          name = 'searchNews';
-          break;
-        case 'wiki':
-          name = 'searchNotes';
-          break;
-        case 'people':
-          name = 'searchPeople';
-          break;
-        case 'perkstore':
-          name = 'searchProducts';
-          break;
-        case 'agenda':
-          name = 'searchEvents';
-          break;
-        default:
-          name = `search${selectedConnector.name.charAt(0).toUpperCase()}${selectedConnector.name.slice(1)}s`;
-        }
-        const searchFilter = {
-          'module': 'portal',
-          'subModule': 'ui',
-          'userId': eXo.env.portal.userIdentityId,
-          'userName': eXo.env.portal.userName,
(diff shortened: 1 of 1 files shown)


component/core/src/main/java/org/exoplatform/social/core/processor/MentionsProcessor.java
```diff
@@ -62,6 +62,9 @@ public void processActivity(ExoSocialActivity activity) {
       if (templateParams.containsKey("default_title")) {
         templateParams.put("default_title", MentionUtils.substituteUsernames(portalOwner, templateParams.get("default_title")));
       }
+      if (templateParams.containsKey("kudosMessage")) {
+        templateParams.put("kudosMessage", MentionUtils.substituteUsernames(portalOwner, templateParams.get("kudosMessage")));
```
boubaker: plz no Kudos in Social layer

webapp/portlet/src/main/webapp/vue-apps/search/components/SearchResults.vue
```diff
@@ -420,6 +422,41 @@ export default {
           });
       });
     },
+    addSearchFilterAnalytics(connectorName) {
```
boubaker: plz no Analytics in Social layer

webapp/portlet/src/main/webapp/vue-apps/search/components/SearchResults.vue
```diff
@@ -290,16 +290,39 @@ export default {
...
           'subModule': 'ui',
           'userId': eXo.env.portal.userIdentityId,
           'userName': eXo.env.portal.userName,
           'operation': 'click',
-          'name': `search${selectedConnector.name.charAt(0).toUpperCase()}${selectedConnector.name.slice(1)}`,
+          'name': name,
           'timestamp': Date.now()
         };
-        document.dispatchEvent(new CustomEvent('exo-statistic-message', {detail: searchFilterAnalytics}));
+        document.dispatchEvent(new CustomEvent('exo-statistic-message', {detail: searchFilter}));
```
boubaker: We can't add statistics collection here, all you can do is to trigger a generic event that should be catched by eXo Analytics addon to collect statistics (No dependency from social to Analytics should be visible in source code of Meeds).

```suggestion
        document.dispatchEvent(new CustomEvent('search-filter-selected', {detail: selectedConnector.name}));
```

And we should delete all the upper computation that builds an analytics object `searchFilter` which should be made in Analytics addon only.

AzmiTouil: Updated !

nidhalnidhal (approved):

PR #1153, opened 2021-12-13
task#37195: Mobile profile page issue with call icon

Amine0008:
ISSUE: margin ma-2 causes misalignment for mobile browsers
FIX: changed margin to ma-1 for mobile browsers


ahamdi:
@Amine0008 I think this is not sufficient as solution, IMO we need to remove the text and just display Call icons in case of Mobile

ahamdi (dismissed):

ahamdi (approved):

PR #1169, opened 2021-12-20
TASK-52377 : In user administration, the confirm password field is already filled.

IlhemEssaadi:
ISSUES : the field confirm password is already filled when adding a new user.
FIX : change v-model to take as value user.confirmNewPassword instead of confirmNewPassword because all the modifications are carried out on the user variable as well as the reset when saving a new user.


rdenarie (changes requested):
Hello.
The fix seems okay. But can you add more detail about how you fix the problem ? why replacing confirmNewPassword by user.confirmNewPassword help to fix ?
In addition can you update you commit message in order to respect the format defined in developer guide  :
https://developer.exoplatform.org/#id-contributions

IlhemEssaadi:
hello.
I modified the commit and I added more details to the description.

rdenarie (approved):

PR #1180, opened 2021-12-22
TASK-51126 : User's profile: Spaces in common: incorrect values

abenmbarek:
ISSUES : incorrect counting of common spaces values in the rest service
FIX : add a new SQL query that returns the common spaces between two given users and use it into a new added service + add unit tests.

### DIFF
--- a/component/api/src/main/java/org/exoplatform/social/core/space/SpaceListAccess.java
+++ b/component/api/src/main/java/org/exoplatform/social/core/space/SpaceListAccess.java
@@ -40,7 +40,7 @@ public class SpaceListAccess implements ListAccess<Space> {
   
   /** The type. */
   Type type;
-           
+
   /**
    * The space list access Type Enum.
    */
--- a/component/api/src/main/java/org/exoplatform/social/core/space/SpaceListAccess.java
+++ b/component/api/src/main/java/org/exoplatform/social/core/space/SpaceListAccess.java
@@ -31,13 +31,16 @@ public class SpaceListAccess implements ListAccess<Space> {
   
   /** The user id. */
   private String userId;
+
+  /** The visited profile id, used while getting list of common spaces between two users. */
+  private String otherUserId;
   
   /** The space filter */
   private SpaceFilter spaceFilter;
   
   /** The type. */
   Type type;
-  
+           
   /**
    * The space list access Type Enum.
    */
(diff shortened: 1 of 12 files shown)


component/api/src/main/java/org/exoplatform/social/core/space/SpaceListAccess.java
```diff
@@ -31,13 +31,16 @@
...
+  /** The visited profile id, used while getting list of common spaces between two users. */
+  private String otherUserId;
   
   /** The space filter */
   private SpaceFilter spaceFilter;
   
   /** The type. */
   Type type;
-  
+
```
rdenarie: this modification is not needed

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/SpaceDAO.java
```diff
@@ -32,4 +32,26 @@
...
+  /**
+   * Get common spaces between two users
+   *
+   * @param userId connected user id
+   * @param otherUserId visited profile user id
+   * @param offset
+   * @param limit
+   * @return list of common spaces between two users in param
+   */
+  List<SpaceEntity> getCommonSpaces(String userId,String otherUserId, int offset, int limit);
```
rdenarie:
```suggestion
  default List<SpaceEntity> getCommonSpaces(String userId,String otherUserId, int offset, int limit) {
    throw new UnsupportedOperationException();
  }
```

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/SpaceDAO.java
```diff
@@ -32,4 +32,26 @@
...
+
+  /**
+   * Count common spaces between two users
+   *
+   * @param userId connected user id
+   * @param otherUserId visited profile user id
+   * @return list of common spaces between two users in param
+   */
+  int countCommonSpaces(String userId,String otherUserId);
```
rdenarie:
```suggestion
  default int countCommonSpaces(String userId,String otherUserId)  {
    throw new UnsupportedOperationException();
  }
```

rdenarie (changes requested):

component/api/src/main/java/org/exoplatform/social/core/space/SpaceListAccess.java
```diff
@@ -31,13 +31,16 @@
...
```
rdenarie: Please remove this modification

rdenarie (changes requested):

rdenarie (approved):

PR #1184, opened 2021-12-23
TASK-52300 : Display when the original space is hidden for non members

nidhalnidhal:
Before these changes , when sharing an activity in the activitystream , the originial Activity space name and log is always displayed even if the current user is not a member of that space . so i added the necessary implementation in order to check the membership of the user in the space of a shared Activity.

### DIFF
--- a/webapp/portlet/src/main/resources/locale/portlet/social/SpacesListApplication_en.properties
+++ b/webapp/portlet/src/main/resources/locale/portlet/social/SpacesListApplication_en.properties
@@ -30,6 +30,7 @@ spacesList.label.spaceTemplate=Template
 spacesList.label.yes=Yes
 spacesList.label.no=No
 spacesList.label.hidden=Hidden
+spacesList.label.hiddenSpace=Hidden space
 spacesList.label.registration=Registration
 spacesList.label.open=Open
 spacesList.label.validation=Validation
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityHead.vue
```diff
@@ -71,9 +74,21 @@ export default {
...
+      return eXo.env.portal.userId;
+    },
+    spaceId() {
+      return this.space.id;
+    },
     posterIdentity() {
       return this.activity && this.activity.identity;
     },
+    isSpaceHidden() {
+      return this.isActivityShared && !this.isSpaceMember;
```
AzmiTouil:
```suggestion
      return this.isActivityShared && !this.$spaceService.isSpaceMember(eXo.env.portal.userId, this.space.id);
```

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityHead.vue
```diff
@@ -71,9 +74,21 @@ export default {
...
+    userId() {
+      return eXo.env.portal.userId;
+    },
+    spaceId() {
+      return this.space.id;
+    },
     posterIdentity() {
       return this.activity && this.activity.identity;
     },
+    isSpaceHidden() {
```
AzmiTouil:
```suggestion
    isHiddenSpace() {
```

AzmiTouil (changes requested):

AzmiTouil (approved):
LGTM

PR #1196, opened 2022-01-03
Task-52602: Pass the current app from which the activity will be shared to the emitted event

hassenelkamel-exo:
Prior to this change, we don't indicate in the share activity emitted event the current app from which the activity will be shared which causes the display many times of the share activity notification alert. To fix this problem, we pass the current app to the emitted event in order to be used when displaying the share activity notification alert.


webapp/portlet/src/main/webapp/vue-apps/common/components/ActivityShareDrawer.vue
```diff
@@ -111,8 +112,9 @@ export default {
       this.description = '';
       this.sharing = false;
     },
-    open(activityId) {
+    open(activityId, selectedApps) {
```
azayati:
```suggestion
    open(activityId, currentApp) {
```

azayati (changes requested):
see review

azayati (approved):
LGTM after QA validation

PR #1202, opened 2022-01-04
Task-52443: Space popover label display problem (#1194)

eromdhaneexo:
Problem: "Administrateurs" label is displayed as divided into two lines, while it should be displayed in only one line.
Fix: "Administrateurs" label is displayed in only one line.


hbenali (dismissed):
Please update the pull request and remove the unnecessary description.

wassimmhamdi (approved):

PR #1217, opened 2022-01-07
Task-50756: Update Administration navigation entry

ahmadmido:


azayati (approved):
LGTM after QA validation

PR #1223, opened 2022-01-11
task45533: added a settings page where the user can download the mobile

Amine0008:
added a settings page where the user can download the mobile application by either scanning a QR code or navigating directly to the relevant app store by clicking on the appropriate store icon


ahamdi (changes requested):
We will change the PR to consider a feedback and move this settings

PR #1233, opened 2022-01-17
TASK-52323 : Suggester does not show name of member in suggestion list unless the fullname is written

abenmbarek:
**ISSUE** : when typing slowly the searched user's name in the space invitation drawer's input, no suggestions displayed because of the wrong test condition which, while typing a new chars in the search input, it tests the previous input text **previousSearchTerm** if it's equal to the new typed one **searchTerm** which is wrong.
In fact, while typing a new text to search, the variable **searchTerm** gets the current typed text so it won't get the same value as the variable **previousSearchTerm** to compare.

**FIX** : change the test condition to compare if the previous typed text variable is not equal to the current text variable before calling the api service **searchSpacesOrUsers** so it can make the search after every typed text regardless to the typing speed.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/common/components/ExoIdentitySuggester.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/common/components/ExoIdentitySuggester.vue
@@ -237,7 +237,7 @@ export default {
       if (value && value.length) {
         window.setTimeout(() => {
           this.focus();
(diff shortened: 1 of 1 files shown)


rdenarie (changes requested):
Can you improve your descrition ?
In the first part, you explain what is the issue, but not what is the root cause.
The second part, explaining how you fix the problem could be also improved

rdenarie (approved):

PR #1237, opened 2022-01-18
TASK-53062 Upgrade to Hibernate 5.6

boubaker:
To make the upgrade to Hibernate 5.6, this commit will:
1/ Delete files org.hibernate.integrator.spi.Integrator that isn't used anymore for @ExoEntity definition injection
2/ Delete definition of persistence.xml that is already made in gatein-portal. For each Persistence Unit, we have to define this file only once
3/ Generate Sequences for hsqldb which wasn't mandatory in previous versions of Hibernate and ensure to manage Sequence allocation size correctly switch Liquibase Configured incrementation (Default = 1)


ahamdi (approved):
LGTM

rdenarie (approved):

PR #1246, opened 2022-01-20
Task-50874: Update createMetadata API to allow creating metadata with properties fields

hassenelkamel-exo:
After this change, we update create metadata API to allow creating metadata with properties fields.


azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM

PR #1257, opened 2022-01-26
TASK-53143: Import users from CSV Files, Non understandable message.

IlhemEssaadi:
ISSUES : Import users from CSV Files, Non understandable message.
FIX : the problem that the old message is not understandable with a bad format and missing the username, so it was solved by modifying the error message to be more understandable and returning the username.


rdenarie (changes requested):
Can you update the description ?
You should explain WHAT is the problem, what is the root cause, and how yuo fix it.

Your current description is just a generic description of the problem

rdenarie (approved):

PR #1259, opened 2022-01-26
Task-53134: Consecutive comments of an activity are grouped in one and only notification.

montassarhamadi:
Past Behavior: Before this commit, when user A has commented many times consecutively on a post in a specific space, One notification received by destinations like the owner of space ,the likers of post etc ... this notification contain the last comment of user A written  on this post without displayed others comments because the past process was getting the previous notification related by same activity and same user if exists then  update it and finally sent it to the destinations and all that implemented in method addUserToPreviousNotification(NotificationInfo notification, String propertyName, String activityId, String userId) in Class SocialNotificationUtils.java.

Current Behavior: In this commit, we simply remove the override function getNotificationToStore(NotificationInfo notification) in created object comment of type AbstractTemplateBuilder  to use the original one which returns object NotificationInfo that does not have id and is isolated from the previous notification that's why it will be created like a new notification then be sent to destinations separately.


component/notification/src/main/java/org/exoplatform/social/notification/channel/template/WebTemplateProvider.java
```diff
@@ -273,10 +273,7 @@ protected boolean makeDigest(NotificationContext ctx, Writer writer) {
      */
     @Override
     public NotificationInfo getNotificationToStore(NotificationInfo notification) {
-      String activityId = notification.getValueOwnerParameter(SocialNotificationUtils.ACTIVITY_ID.getKey());
-      String parameterName = SocialNotificationUtils.POSTER.getKey();
-      String parameterValue = notification.getValueOwnerParameter(parameterName);
-      return SocialNotificationUtils.addUserToPreviousNotification(notification, parameterName, activityId, parameterValue);
+      return notification;
```
rdenarie: Why don't you simply remove this function ? the method AbstractTemplateProvider.getNotificationToStore make exactly the same thing as you.

rdenarie (changes requested):

PR #1263, opened 2022-01-26
Task-53253: Double delete action to delete a space group

wassimmhamdi:
Problem: The space group is deleted after clicking  twice on delete button
Fix: we set the broadcast event listener that it deletes a space without deleting the user's group. When a space is deleted, all of its page navigations and its group will be deleted and the listener get not started for a second time


rdenarie (dismissed):

rdenarie (changes requested):
Can you fix the sonar feedback ? Thanks

rdenarie (approved):

PR #1272, opened 2022-01-28
Task-50800: Implement updateMetadata service API

hassenelkamel-exo:


component/core/src/main/java/org/exoplatform/social/core/metadata/MetadataServiceImpl.java
```diff
@@ -80,6 +80,24 @@ public Metadata createMetadata(Metadata metadata, long userIdentityId) {
     }
     return metadata;
   }
+
+  @Override
+  public Metadata updateMetadata(Metadata metadata, long userIdentityId) {
```
azayati, line 85: userIdentityId not needed IMHO

azayati (changes requested):
see review

component/core/src/main/java/org/exoplatform/social/core/metadata/MetadataServiceImpl.java
hassenelkamel-exo: used in broadcast event

azayati: ok we can keep it

azayati (approved):
LGTM after QA validation

PR #1292, opened 2022-02-08
Task-53542: Hidden Space's Activity: Hidden Space ID can be found in code

wassimmhamdi:
Problem: by inspecting the page source code, of a Hidden Space, the space URL is displayed to users who's are not in these space
Fix: No informations related to hidden spaces are displayed for users that are not its members


rdenarie (changes requested):
KO as waiting QA test

rdenarie (approved):

PR #1297, opened 2022-02-10
Task-53906: Review the way we attach files to an activity

AzmiTouil:
Add attach file option to activity editor


SaraBoutej (approved):
LGTM

PR #1298, opened 2022-02-11
Task-54171 : DatePicker problem when user timezone is negative

rdenarie:
Before this fix, when the user timezone is negative (UTC-5 for example) the date-picker component to not work well :
an infinite loop is generated when editing experiences in profile, a lot of notifications are generated when editing a task ...

This is due to a misusage of the date object and string representation which was not consistent in the component.
The current fix defines a new String object to keep the date value for the v-date-picker, with format ISO8601 (short: YYYY-MM-DD), and keep the current date in a date object.
In addition, functions for date manipulation are updated to be able to work with different date formats (ISO8601 short, ISO8601 Extended, Date Object, timestamp ...)


ghost:
### CodeSee Review Map:
[Review these changes using an interactive CodeSee Map](https://app.codesee.io/r/reviews?pr=1298&src=https%3A%2F%2Fgithub.com%2FMeeds-io%2Fsocial)

Review [in an interactive map](https://app.codesee.io/r/reviews?pr=1298&src=https%3A%2F%2Fgithub.com%2FMeeds-io%2Fsocial)

View [more CodeSee Maps](https://app.codesee.io/r/maps?src=https%3A%2F%2Fgithub.com%2FMeeds-io%2Fsocial)

#### Legend
CodeSee Map Legend

wassimmhamdi (approved):

PR #1305, opened 2022-02-16
TASK-54025 Rework Login Process to make it more extensible using Regular Vue & Vuetify Application

boubaker:
Prior to this change, the login.jsp file doesn't allow extensible content and in addition it doesn't define an Application using common UX using Vue & Vuetify. This enhancement will :

- Move JSP definitions into Social layer to make sure that UI layer is defined in the adequate layer
- Define Login Application using Vue & Vuetify and make it extensible
- Define OAuth Buttons in Login page with new extension mechanism


AzmiTouil (approved):
LGTM

AzmiTouil (approved):

PR #1308, opened 2022-02-17
Task-53659: Prevent closing message composer drawer when other drawers are opened

azayati:
Prior to this change, when the message composer drawer is opened with other drawers coming from its actions, the message composer drawer is closed before the other opened drawers. This is due to the fact that the message composer drawer is not an exo:drawer component which escapes it from the closing rule of exo-drawer components (from last open to first opened).
After this change, we ensure that closing message composer drawer is done only when no other drawer is opened.


azayati (commented):
This PR will wait until activity composer refactoring cc @AzmiTouil

PR #1309, opened 2022-02-17
Task-54024: Inform users about new way of attaching files

AzmiTouil:
This PR will add a popup that contain a message to explain for user the new way of attaching files to an activity


SaraBoutej (approved):
LGTM

PR #1320, opened 2022-02-18
Task-54499: Allow to customize rows and row-height propos of extended-textarea component

abenmbarek:
Prior to this change, extended-textarea component is always displayed with fixed row height and rows number props values. After this change, we will be able to customize these props values when using the extended-textarea component.


webapp/portlet/src/main/webapp/vue-apps/common/components/ExtendedTextarea.vue
```diff
@@ -6,6 +6,8 @@
     :placeholder="placeholder"
     counter
     auto-grow
+    :rows="rows"
+    :row-height="rowsHeight"
```
azayati:
```suggestion
    :row-height="rowHeight"
```

webapp/portlet/src/main/webapp/vue-apps/common/components/ExtendedTextarea.vue
```diff
@@ -23,7 +25,15 @@ export default {
     maxLength: {
       type: Number,
       default: () => 0
-    }
+    },
+    rows: {
+      type: Number,
+      default: () => 5
+    },
+    rowsHeight: {
```
azayati:
```suggestion
    rowHeight: {
```

azayati (changes requested):
see review

azayati (approved):
LGTM after QA validation

PR #1333, opened 2022-02-24
TASK-54243 : search news by tags

nidhalnidhal:
enable unified search based on tags for news .

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/metadata/MetadataServiceImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/metadata/MetadataServiceImpl.java
@@ -184,9 +184,7 @@ public class MetadataServiceImpl implements MetadataService, Startable {
     }
     metadataItem = metadataStorage.createMetadataItem(metadataItem);
     try {
-      if(!metadataItem.getObjectType().equals(MetadataActivityProcessor.NEWS_METADATA_OBJECT_TYPE)) {
         this.listenerService.broadcast("social.metadataItem.created", userIdentityId, metadataItem);
-      }
     } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


component/core/src/main/java/org/exoplatform/social/core/metadata/MetadataServiceImpl.java
```diff
@@ -183,7 +184,9 @@ public MetadataItem createMetadataItem(MetadataObject metadataObject,
     }
     metadataItem = metadataStorage.createMetadataItem(metadataItem);
     try {
-      this.listenerService.broadcast("social.metadataItem.created", userIdentityId, metadataItem);
+      if(!metadataItem.getObjectType().equals(MetadataActivityProcessor.NEWS_METADATA_OBJECT_TYPE)) {
```
AzmiTouil: should be handled in news side

AzmiTouil (changes requested):

component/core/src/main/java/org/exoplatform/social/core/processor/MetadataActivityProcessor.java
```diff
@@ -36,6 +36,8 @@
 
   public static final String ACTIVITY_METADATA_OBJECT_TYPE = "activity";
 
+  public static final String NEWS_METADATA_OBJECT_TYPE = "news";
```
AzmiTouil, line 39: To be removed please!

AzmiTouil (changes requested):

component/core/src/main/java/org/exoplatform/social/core/processor/MetadataActivityProcessor.java
nidhalnidhal: Done!

PR #1343, opened 2022-03-02
TASK-52834: Feature flag for tags set to true for any user

nidhalnidhal:
enable Tag feature for any user


extension/war/src/main/webapp/WEB-INF/conf/social-extension/social/core-configuration.xml
```diff
@@ -724,7 +724,7 @@
             <name>ActivityTagsFeatureProperties</name>
             <description>New Activity Stream Feature enablement flag</description>
             <property name="exo.feature.ActivityTags.enabled" value="${exo.feature.ActivityTags.enabled:false}"/>
```
AzmiTouil: IMO, you just need to remove the permission and set the default value to true

AzmiTouil (changes requested):

extension/war/src/main/webapp/WEB-INF/conf/social-extension/social/core-configuration.xml
nidhalnidhal: I agree . Done!

AzmiTouil (approved):

PR #1355, opened 2022-03-07
TASK-53670 : Tooltip proposed from the post

nidhalnidhal:
propose the new  search tooltip when hovering on a tag in any type of activity


hbenali (commented):
@nidhalnidhal Please close this PR if it is no longer needed cc @AzmiTouil

PR #1357, opened 2022-03-08
Task-54742: Dispatch post-activity event when the activity type is not simple, file or link activity

abenmbarek:
Before this change, for any type of posted activity, create activity rest service is invoked from activity composer.
After this change, we will dispatch a post-activity event when the activity type is not simple, file or link activity in order to allow invoking custom rest services in that cases.


azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM

PR #1370, opened 2022-03-15
Task-55243: Avoid to display link activity while updating a not simple, file or link activity activity

abenmbarek:


azayati (approved):
LGTM after QA validation

PR #1374, opened 2022-03-18
Task-55414: Enable posting activities without message

abenmbarek:
Prior to this fix, we cannot post poll activity without a message from the composer. After this fix, we will be able to post an activity even without adding a message in the case of file activity for example.


azayati (approved):
LGTM after QA validation

PR #1383, opened 2022-03-24
Task-53140: Shared activities - display of hidden space

AzmiTouil:
Display "Hidden space" when the space from where the activity has been shared is hidden and the user is not a member of space


SaraBoutej (approved):
LGTM

PR #1390, opened 2022-03-28
Task-55684: Current activity template params are lost when liking or disliking

azayati:
Prior to this change, when liking or disliking an activity with attached files, its template params are erased with files ones even if it has other template params. After this commit, we ensure to preserve the current template params when liking or disliking.


abenmbarek (approved):
LGTM after QA validation

PR #1393, opened 2022-03-30
Task-54817: Avoid link redirection when opening notification details drawer

hassenelkamel-exo:


azayati (approved):
LGTM after QA validation

PR #1400, opened 2022-04-01
TASK-54726: fix bad behaviours with dates (#1367)

GouadriaHanen:
fix bad behavior for dates have a null negative timezone
fix clearing date (make dateShortIso null)


boubaker:
As seen together, the bug is coming from backend of challenges. To avoid regression on other applications using this re-usable component, please close this issue and open a new one on gamification.

boubaker (commented):
As seen together, the bug is coming from backend of challenges. To avoid regression on other applications using this re-usable component, please close this issue and open a new one on gamification.

PR #1401, opened 2022-04-04
Task-54819: Mark as read details drawer notifications

hassenelkamel-exo:
Prior to this change, notifications opening details drawer are not mark as read when viewed. After this change, we ensure to mark as read this kind of notifications.


azayati (approved):
LGTM

PR #1417, opened 2022-04-15
Task-53938: Add loading effect to favorite drawer

SaraBoutej:
Prior to this change, the loading effect doesn't exist in the favorites drawer, also when there is no favorite added by the user, an empty drawer was opened.
After this change, we add loading effect to the favorites drawer and also add an interface indicates that the user has not yet added a favorite when he opens the drawer


webapp/portlet/package-lock.json
```diff
@@ -478,7 +478,6 @@
         "merge-source-map": "^1.1.0",
         "postcss": "^7.0.36",
         "postcss-selector-parser": "^6.0.2",
-        "prettier": "^1.18.2 || ^2.0.0",
```
AzmiTouil: To be reverted please!

AzmiTouil (changes requested):

webapp/portlet/package-lock.json
SaraBoutej: done

AzmiTouil (approved):

PR #1434, opened 2022-04-26
TASK-56601: Duplication and bad detection and display of tags

hakermi:
Prior to this changes, when creation a tag and then press the space button each time will create another tag with number of spaces converted to `&nbsp; `as the matcher of tags pattern will consider it as a normal character follows the tag text which will duplicate the tag depends on the number of spaces that follows it and display it with the `&nbsp;` in the unified search when selecting tags, this because of the returned html content of the activity when searching for tags,.
This PR will sanitize the content to correctly unescape the` &nbsp;` and replacing it by spaces to allow the tags pattern matcher to work as expected.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/listeners/ActivityTagMetadataListener.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/listeners/ActivityTagMetadataListener.java
@@ -23,6 +23,9 @@ import java.util.Set;
 import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang3.StringUtils;
 
+import org.exoplatform.commons.utils.HTMLSanitizer;
+import org.exoplatform.services.log.ExoLogger;
+import org.exoplatform.services.log.Log;
 import org.exoplatform.social.core.activity.ActivityLifeCycleEvent;
 import org.exoplatform.social.core.activity.ActivityListenerPlugin;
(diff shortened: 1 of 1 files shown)


component/core/src/main/java/org/exoplatform/social/core/listeners/ActivityTagMetadataListener.java
```diff
@@ -75,6 +80,11 @@ private void updateActivityTags(ExoSocialActivity activity) {
     Identity audienceIdentity = activityManager.getActivityStreamOwnerIdentity(activity.getId());
     long audienceId = Long.parseLong(audienceIdentity.getId());
     String content = getActivityBody(activity);
+    try {
+      content = HTMLSanitizer.sanitize(content);
+    } catch (Exception e) {
+      LOG.error("Error while sanitizing activity content", e);
```
boubaker:
```suggestion
      LOG.warn("Error while sanitizing activity content {}", content, e);
```
It's better to use LOG.warn level when the error is not blocker to the whole system and that it doesn't need attention from Ops.
BTW, it's better to display additional information concerning the error for better problem analysis

hakermi: Okay

boubaker (approved):

PR #1437, opened 2022-04-28
Task-56521: Fix duplication of news and activity items in favorites drawer

SaraBoutej:
Before this Fix when we bookmark a news, we have the related activity that is bookmarked too, so in the drawer of recent favorites we have duplication of items: activity + news.
With this fix, we add a boolean property "hidden" when bookmark an item to avoid the duplication of an item in the favorites drawer.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/metadata/favorite/FavoriteServiceImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/metadata/favorite/FavoriteServiceImpl.java
@@ -19,6 +19,7 @@
 package org.exoplatform.social.core.metadata.favorite;
 
 import java.util.List;
+import java.util.Map;
 
 import org.apache.commons.collections.CollectionUtils;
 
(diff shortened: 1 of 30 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/footer/actions/ActivityFavoriteAction.vue
```diff
@@ -60,6 +61,9 @@ export default {
     extensionName() {
       return this.activityTypeExtension && this.activityTypeExtension.name;
     },
+    hiddenFavorite() {
+      return this.activity && this.activity.type && this.activity.type === 'news' ? true : false;
```
boubaker, line 65: shouldn't embed eXo addon name inside Meeds code. Meeds code should be flexible to allow the addon to parameter this kind of properties through extensionRegistry

component/core/src/main/java/org/exoplatform/social/core/metadata/favorite/FavoriteServiceImpl.java
```diff
@@ -39,15 +40,13 @@ public FavoriteServiceImpl(MetadataService metadataService) {
   }
 
   @Override
-  public void createFavorite(Favorite favorite) throws ObjectAlreadyExistsException {
+  public void createFavorite(Favorite favorite, Map<String, String> favoriteProperties) throws ObjectAlreadyExistsException {
```
boubaker, line 43:
```suggestion
  @Override
  public void createFavorite(Favorite favorite) throws ObjectAlreadyExistsException {
    this.createFavorite(favorite, false);
  }

  @Override
  public void createFavorite(Favorite favorite, boolean hiddeInList) throws ObjectAlreadyExistsException {
```
In term of API definition, multiple things:
1/ It's better to keep the original operation with simple input without options `createFavorite(Favorite favorite)`. This will allow to use the API for simple use cases without complexity.
2/ It's better to not add a generic `properties` Map while our use case is simple: whether to hide favorites while listing or not.
3/ You added an option for favorite creation, but it's not clear (in term of API perspective) where it will be used. In fact, a new method to list favorites should be added in Service layer to allow filter by visibility.

PR #1448, opened 2022-05-09
Task-56663: initialize the user name input field with the url property username if its available

Amine0008:
ISSUE: when we scan the QR code in the mobile settings page, the expected behavior is for the user to be redirected to the login page where the user name field of the login form should be auto filled with the username  url property (..../portal/login?username=... )
FIX: check if the username property exists in the login url and if this is the case we simply set the user name input field value to this paramater

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/login/components/Login.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/login/components/Login.vue
@@ -93,6 +93,7 @@
                         id="username"
                         :placeholder="$t('portal.login.Username')"
                         name="username"
+                        v-model="username"
                         tabindex="1"
                         type="text"
                         aria-required="true">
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/login/components/Login.vue
```diff
@@ -218,7 +218,16 @@ export default {
...
   },
   mounted() {
+    this.setupUserName();
     this.$root.$applicationLoaded();
   },
+  methods: {
+    setupUserName(){
+      const urlParams = new URLSearchParams(window.location.search);
+      if (urlParams.has('username')) {
+        document.getElementById('username').value = urlParams.get('username');
```
boubaker: it's better to define a `v-model` for username input and benefit from double binding of Vue

boubaker (approved):

PR #1453, opened 2022-05-10
Task-57018: Hidden space indication in an event for a member . (#1452)

montassarhamadi:
Problem: when opening the event1 it's well indicated that it's added under a hidden space instead of display name for validation and hidden spaces.
 Fix: replace the empty value of parameter userId with the correct id of the current user in the method buildEntityFromSpace () to check if this user is a member of the space or not.

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
@@ -178,8 +178,10 @@ public class EntityBuilder {
     identityEntity.setRemoteId(identity.getRemoteId());
     identityEntity.setDeleted(identity.isDeleted());
     if (SpaceIdentityProvider.NAME.equals(identity.getProviderId())) {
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/common/components/ExoSpaceAvatar.vue
```diff
@@ -32,7 +32,7 @@
             class="pull-left text-truncate ms-2">
             <p
               class="text-truncate subtitle-2 my-auto hidden-space">
-              {{ $t('spacesList.label.hiddenSpace') }}
+              {{ displayName }}
```
boubaker: Since, this is a new PR for this old task, I add a remark on it.  The space name shouldn't be accessible when it's hidden and the user isn't member of, please check the property `notAccessibleSpace`. Please check with @AzmiTouil to get more information.

boubaker (approved):

PR #1455, opened 2022-05-10
Task-56986: Prevent updating activity stream when posting hidden activities 

hchaabene:
Prior to this fix, when posting a hidden activity, a blue flag "new post" is displayed on the space activity stream.
After this fix, we avoid displaying the flag when posting a hidden activity


azayati (changes requested):
IMHO, Since it is a general problem of any type of hidden activity, the title and the description should be generic to be not news activity related only (Also to not mention an exo addon on a meeds codebase)

boubaker (approved):

azayati (approved):
LGTM

azayati:
@hchaabene don't forget to take into account the update of the title and the description according to our last discussion

PR #1457, opened 2022-05-11
TASK-56601: Unescaped html characters in unified search last added tags list

hakermi:
Prior to this change, last added tags list are displayed with unescaped html characters, because the HTMLSanitizer was converting some char like apostrophe to html chars.
This PR use StringEscapeUtils.unescapeHtml to avoid such unwanted conversion

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/listeners/ActivityTagMetadataListener.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/listeners/ActivityTagMetadataListener.java
@@ -21,11 +21,9 @@ package org.exoplatform.social.core.listeners;
 import java.util.Set;
 
 import org.apache.commons.collections.MapUtils;
+import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.commons.lang3.StringUtils;
 
-import org.exoplatform.commons.utils.HTMLSanitizer;
-import org.exoplatform.services.log.ExoLogger;
-import org.exoplatform.services.log.Log;
 import org.exoplatform.social.core.activity.ActivityLifeCycleEvent;
 import org.exoplatform.social.core.activity.ActivityListenerPlugin;
 import org.exoplatform.social.core.activity.model.ExoSocialActivity;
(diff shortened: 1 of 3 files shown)


webapp/portlet/src/main/webapp/vue-apps/search/components/SearchTagList.vue
```diff
@@ -37,7 +37,10 @@
             :value="tag"
             @click="handleTag(tag)"
             @keyup.enter="handleTag(tag)">
-            <span :class="`${isMobile ? 'primary--text' : ''}`"> {{ tag }}</span>
+            <span
+              :class="`${isMobile ? 'primary--text' : ''}`"
+              v-sanitized-html="tag">
```
boubaker: The tag shouldn't contain any html character. so the fix should be in tag creation process

hakermi: PR updated

component/core/src/main/java/org/exoplatform/social/core/listeners/ActivityTagMetadataListener.java
```diff
@@ -79,12 +75,7 @@ private void updateActivityTags(ExoSocialActivity activity) {
 
     Identity audienceIdentity = activityManager.getActivityStreamOwnerIdentity(activity.getId());
     long audienceId = Long.parseLong(audienceIdentity.getId());
-    String content = getActivityBody(activity);
-    try {
-      content = HTMLSanitizer.sanitize(content);
-    } catch (Exception e) {
-      LOG.warn("Error while sanitizing activity content {}", content, e);
-    }
+    String content = StringEscapeUtils.unescapeHtml(getActivityBody(activity));
```
boubaker: we may need to unescape HTML in `tagService.detectTagNames` instead to centralize this behavior + a UT please

hakermi: done

boubaker (approved):

PR #1459, opened 2022-05-11
 Task-57040:  Still receiving notifications on old posted activities after leaving a space .

montassarhamadi:
Problem: user still receives the notification of the like or comment etc... and when clicking on it, it redirects to his activity.
 Fix: The user cannot receive any notification of like, comment, reply to comment, or like comment on activity. A check is added and sending of notifications depends on this last .but the problem of access to activity is still reproduced until now for the removed user.


boubaker:
This should be worked in a proper Task if accepted as improvement by POs, thus the PR is no needed anymore. I closed it until a new task will be added.

PR #1460, opened 2022-05-11
TASK-56521 Enhance Activity Metadata Association process to avoid duplication with related entities

boubaker:
Prior to this change, the favorite icon displayed in Activity stream was referencing the activity itself and doesn't allow to reference an attached entity. This modification allows to transfer all associated Metadata of an activity to an attached Object identified by new attributes introduced in ExoSocialActivity (metadataObjectType, metadataObjectId, metadataObjectParentId).
This will allow to transfer all Metadata associated to an activity to its attached entity.
This modification had allowed to display favorites in Activity stream while it's associated to attached Object instead of attaching them directly into activity. Thus, the lifecycle of Metadata will follow the object instead of following the activity lifecycle and it will allow to add more accurate information in generated indexes.


AzmiTouil (approved):
LGTM

PR #1472, opened 2022-05-13
Task-57018: Hidden space indication in an event for a member. (#1453)

montassarhamadi:
Problem: when opening the event1 it's well indicated that it's added under a hidden space instead of display name for validation and hidden spaces.
Fix: replace the empty value of parameter userId with the correct id of the current user  in the method buildEntityFromSpace () to check if this user is a member of the space or not.


boubaker:
please make sure to include only https://github.com/Meeds-io/social/pull/1472/commits/dad52546546ccac540c209f8fcb8c9dcc21ef2bf

boubaker (dismissed):

boubaker (approved):

PR #1477, opened 2022-05-16
Task-56986: Prevent sending notifications in case of any actions related to hidden activities

hchaabene:
Prior to this fix, when posting a hidden activity, a blue flag "new post" is displayed on the space activity stream.
After this fix, we avoid displaying the flag when posting a hidden activity


boubaker (approved):

azayati (approved):
LGTM

PR #1488, opened 2022-05-18
Task 57394: Fix user Disconnect dialog not working properly #1486

Amine0008:
ISSUE: when clicking the Disconnect button of a user and then confirm button in the confirmation dialog, the action of disconnecting doesn't work.
FIX: changed the dialog close method to set the dialog boolean variable to false and then call emitClosedEvent method to emit both the "dialog-closed" and the "modalClosed" events


boubaker (approved):
Please make sure to not include two PR IDs in title of the commit when merging (As changed in the PR title)

PR #1494, opened 2022-05-19
TASK-56781 Refactor User Popover to fix display of popover on activity and comment body

boubaker:
Prior tot his change, the suggested users inside activity and comment bodies doesn't display popover on it when the user hover the links. This change will introduce a new Vue directive to allow displaying popover on specific elements using a centralized application (without defining a menu/dialog per user popover) for better performances.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
@@ -35,6 +35,7 @@ import org.exoplatform.services.resources.LocaleContextInfo;
 import org.exoplatform.services.resources.LocalePolicy;
 import org.exoplatform.services.resources.ResourceBundleService;
 import org.exoplatform.social.core.identity.model.Identity;
+import org.exoplatform.social.core.identity.model.Profile;
 import org.exoplatform.social.core.identity.provider.OrganizationIdentityProvider;
 import org.exoplatform.social.core.identity.provider.SpaceIdentityProvider;
 import org.exoplatform.social.core.manager.IdentityManager;
@@ -47,6 +48,8 @@ import org.exoplatform.web.security.codec.CodecInitializer;
 import org.exoplatform.web.security.security.TokenServiceInitializationException;
 import org.exoplatform.web.url.navigation.NavigationResource;
 import org.exoplatform.web.url.navigation.NodeURL;
+
+import com.google.api.client.util.Objects;
 
 /**
  * Builds and provides default links and links of users, spaces and activities.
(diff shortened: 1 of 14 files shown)


component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
```diff
@@ -48,6 +49,8 @@
 import org.exoplatform.web.url.navigation.NavigationResource;
 import org.exoplatform.web.url.navigation.NodeURL;
 
+import com.google.api.client.util.Objects;
+
```
boubaker:
```suggestion
```

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/content/ActivityBody.vue
```diff
@@ -32,6 +32,11 @@ export default {
     body: null,
   }),
   computed: {
+    bodyElement() {
+      return {
+        template: `<div>${this.body}</div>` || '',
```
boubaker, line 37: Use the same function here to sanitize manually the body without directive: 

https://github.com/Meeds-io/commons/blob/9564910c74dae791d1f87c1ca5155ddb735451eb/commons-extension-webapp/src/main/webapp/WEB-INF/gatein-resources.xml#L428-L445

BTW, this function should be added inside a centralized function that is used here and that can be used elsewhere, such as here: 
https://github.com/Meeds-io/social/blob/6134dad651b2a4b119c14871a1d5fa4d36de5f08/webapp/portlet/src/main/webapp/vue-apps/common/js/Utils.js

webapp/portlet/src/main/webapp/vue-apps/popover/components/PopoverUser.vue
```diff
@@ -0,0 +1,182 @@
+<template>
+  <v-menu
+    v-model="menu"
+    rounded="rounded"
+    :close-on-content-click="false"
+    :left="popoverLeftPosition"
```
boubaker: the variable `popoverLeftPosition` seems not changed, why is it used here, is it for RTL ?

webapp/portlet/src/main/webapp/vue-apps/popover/components/PopoverUser.vue
```diff
@@ -0,0 +1,182 @@
...
+      // Close user popover when the user goes outside the hovered user popover after two seconds
+      this.closePopoverMenu = setTimeout(() => {
+        if (this.menu) {
+          if (!this.isMenuHovered) {
+            this.menu = false;
+          }
+        }
+      }, this.popoverDisplayDelay);
+      if (this.menu) {
+        this.externalExtensions.map((extension) => {
```
boubaker: do you mean `forEach` ? BTW, you can simply use this IMO : 
```javascript
        this.$nextTick().then(() => this.externalExtensions.forEach(this.initExtensionAction));
```

But, I can see errors and bugs coming if we don't make this : 
```javascript
      if (this.menu) {
        this.refreshExtensions();
        this.$nextTick().then(() => this.externalExtensions.forEach(this.initExtensionAction));
      } else {
        this.externalExtensions = [];
      }
```
This will make sure that actions are re-initialized each time to not keep the action button of previous popover display of a different user enabled.

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/content/ActivityBody.vue
boubaker: Ok, `HTMLSanitizer` is already used to purify HTML, thus I close this as resolved.

boubaker:
LGTM

PR #1501, opened 2022-05-23
TASK-36065: Second space manager is unable to correctly leave the space

hakermi:
Prior to this change, when a space manager attempts to leave the space, only member membershipType is handled in the leave rest endpoint, while the user will remain as a manager in the space which will cause two issues: button leave space is always displayed and you can't invite the user to the space again, also added `removeRedactor` because when the user joins the space again, he will be a redactor already.
This PR should remove the user as manager while leaving the space if it's already a manager

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/suggest/SpaceRestServices.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/suggest/SpaceRestServices.java
@@ -315,11 +315,9 @@ public class SpaceRestServices implements ResourceContainer {
         if (space == null) {
           return Response.status(HTTPStatus.NOT_FOUND).build();
         }
-  
         if (spaceService.isMember(space, userId)) {
           spaceService.removeMember(space, userId);
         }
-  
         return Response.ok().build();
       } catch (Exception e) {
         LOG.error("Error in space deny rest service: " + e.getMessage(), e);
--- a/component/core/src/main/java/org/exoplatform/social/core/space/SpaceUtils.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/space/SpaceUtils.java
@@ -1064,6 +1064,16 @@ public class SpaceUtils {
   }
 
   /**
+   * Removes the user from group with any(*) membership.
+   *
+   * @param remoteId target user remote id
+   * @param groupId group id
+   */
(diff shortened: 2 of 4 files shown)


ahamdi:
@hakermi I have a question : what about other memberships ? redactor for example ?

hakermi:
> @hakermi I have a question : what about other memberships ? redactor for example ?

I tested with redactors, the issue is only with manager membership

hakermi:
> @hakermi I have a question : what about other memberships ? redactor for example ?

@ahamdi I added `removeRedactor` too, because when the user joins the space again he will be a redactor already. because the membership wasn't already removed

component/service/src/main/java/org/exoplatform/social/rest/suggest/SpaceRestServices.java
```diff
@@ -315,11 +315,15 @@ public Response leave(@ApiParam(value = "Space technical identifier", required =
         if (space == null) {
           return Response.status(HTTPStatus.NOT_FOUND).build();
         }
-  
+        if (spaceService.isManager(space, userId)) {
+          spaceService.setManager(space, userId, false);
+        }
```
boubaker: IMO, it would be better to add this business logic in Service layer as recommended here : https://community.exoplatform.com/portal/g/:spaces:engineering/exo_architects/notes/13754

ahamdi:
> > @hakermi I have a question : what about other memberships ? redactor for example ?
> 
> @ahamdi I added `removeRedactor` too, because when the user joins the space again he will be a redactor already. because the membership wasn't already removed

ok, that is fine

component/service/src/main/java/org/exoplatform/social/rest/suggest/SpaceRestServices.java
hakermi: done

boubaker (commented):
I think that it would better to delete all memberships of user inside the space **systematically** when the user membership is removed. consequently, once the method `removeMember` is invoked, automatically we should revoke manager & redactor memberships + `*` if this membership type exist for current user (in OrganizationService). It's better to not add a dedicated method to this and centralize this behavior in removeMember.

boubaker (approved):

PR #1504, opened 2022-05-23
Task-56332: Error in log while typing characters to get suggested users

AzmiTouil:
Prior this change, an error appears in the log when trying to get suggested users

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/service/rest/PeopleRestService.java
+++ b/component/service/src/main/java/org/exoplatform/social/service/rest/PeopleRestService.java
@@ -537,14 +537,13 @@ public class PeopleRestService implements ResourceContainer{
                                                          boolean filterByName,
                                                          Locale locale) {
     Identity userIdentity;
-    boolean identityId = StringUtils.isNumeric(userId);
-    if (identityId) {
+    if (StringUtils.isNumeric(userId)) {
       userIdentity = getIdentityManager().getIdentity(userId, false);
     } else {
-      userIdentity = getIdentityManager().getOrCreateIdentity(OrganizationIdentityProvider.NAME, userId, false);
+      userIdentity = getIdentityManager().getOrCreateUserIdentity(userId);
     }
     if (userIdentity == null) {
-      LOG.warn("Cannot find user identity with {} = {}", identityId ? "id" : "remoteId", userId);
+      LOG.debug("Cannot find user identity by id nor remoteId {}", userId);
       return userInfos;
     }
     return addUserToInfosList(userIdentity, identityFilter, userInfos, currentUserId, filterByName, locale);
--- a/component/service/src/main/java/org/exoplatform/social/service/rest/PeopleRestService.java
+++ b/component/service/src/main/java/org/exoplatform/social/service/rest/PeopleRestService.java
@@ -530,13 +530,21 @@ public class PeopleRestService implements ResourceContainer{
     return userInfos;
   }
 
-  private LinkedHashSet<UserInfo> addUsernameToInfosList(String userId, ProfileFilter identityFilter, LinkedHashSet<UserInfo> userInfos, String currentUserId, boolean filterByName, Locale locale) {
-    Identity userIdentity = getIdentityManager().getIdentity(userId, false);
-    if (userIdentity == null) {
+  private LinkedHashSet<UserInfo> addUsernameToInfosList(String userId,
+                                                         ProfileFilter identityFilter,
+                                                         LinkedHashSet<UserInfo> userInfos,
+                                                         String currentUserId,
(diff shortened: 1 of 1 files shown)


boubaker:
It's better to not mention another addon in lower layers + describe the problem by the API problem + Unit Tests.
Besides, the problem is that Agenda Events uses identity ids instead of usernames which is a bug of Agenda itself IMO because the API here has to be coherent and not make two type of parameters for a single one (Identity technical identifier or User Name).

PS: it can also not being a bug in Agenda, if it's just a check about identity existence, thus the log shouldn't be displayed and has to be turned into debug mode. To check IMO.

AzmiTouil:
> It's better to not mention another addon in lower layers + describe the problem by the API problem + Unit Tests. Besides, the problem is that Agenda Events uses identity ids instead of usernames which is a bug of Agenda itself IMO because the API here has to be coherent and not make two type of parameters for a single one (Identity technical identifier or User Name).
> 
> PS: it can also not being a bug in Agenda, if it's just a check about identity existence, thus the log shouldn't be displayed and has to be turned into debug mode. To check IMO.

It's not a agenda bug, the events uses the username in this case, but the API tries to get the identity with the id first, even if the parameter is the username, and that's the cause of the log warn. IMO, we need to check the userId param (id or remoteID) to know which method will be used next.
@boubaker WDYT ?

component/service/src/main/java/org/exoplatform/social/service/rest/PeopleRestService.java
```diff
@@ -530,13 +530,21 @@ private LinkedHashSet<UserInfo> addUserToInfosList(Identity userIdentity, Profil
...
-    if (userIdentity == null) {
+  private LinkedHashSet<UserInfo> addUsernameToInfosList(String userId,
+                                                         ProfileFilter identityFilter,
+                                                         LinkedHashSet<UserInfo> userInfos,
+                                                         String currentUserId,
+                                                         boolean filterByName,
+                                                         Locale locale) {
+    Identity userIdentity;
+    boolean identityId = StringUtils.isNumeric(userId);
+    if (identityId) {
```
boubaker:
```suggestion
    if (StringUtils.isNumeric(userId)) {
```

boubaker:
```suggestion
    if (StringUtils.isNumeric(userId)) {
```

component/service/src/main/java/org/exoplatform/social/service/rest/PeopleRestService.java
```diff
@@ -530,13 +530,21 @@ private LinkedHashSet<UserInfo> addUserToInfosList(Identity userIdentity, Profil
...
+      userIdentity = getIdentityManager().getIdentity(userId, false);
+    } else {
       userIdentity = getIdentityManager().getOrCreateIdentity(OrganizationIdentityProvider.NAME, userId, false);
     }
     if (userIdentity == null) {
-      LOG.warn("Cannot find user identity with username = " + userId);
+      LOG.warn("Cannot find user identity with {} = {}", identityId ? "id" : "remoteId", userId);
```
boubaker:
```suggestion
      LOG.debug("Cannot find user identity by id nor remoteId {}", userId);
```

boubaker:
```suggestion
      LOG.debug("Cannot find user identity by id nor remoteId {}", userId);
```

component/service/src/main/java/org/exoplatform/social/service/rest/PeopleRestService.java
```diff
@@ -530,13 +530,21 @@ private LinkedHashSet<UserInfo> addUserToInfosList(Identity userIdentity, Profil
...
+                                                         LinkedHashSet<UserInfo> userInfos,
+                                                         String currentUserId,
+                                                         boolean filterByName,
+                                                         Locale locale) {
+    Identity userIdentity;
+    boolean identityId = StringUtils.isNumeric(userId);
+    if (identityId) {
+      userIdentity = getIdentityManager().getIdentity(userId, false);
+    } else {
       userIdentity = getIdentityManager().getOrCreateIdentity(OrganizationIdentityProvider.NAME, userId, false);
```
boubaker:
```suggestion
      userIdentity = getIdentityManager().getOrCreateUserIdentity(userId);
```

boubaker:
```suggestion
      userIdentity = getIdentityManager().getOrCreateUserIdentity(userId);
```

boubaker (approved):

PR #1526, opened 2022-05-26
task-57496: Fix loading page for room display space's default avatar

IlhemEssaadi:
Add a static attribute ROOM_DEFAULT_AVATAR_URL which contains the url of the room avatar to the LinkProvider class.


component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
```diff
@@ -59,6 +59,7 @@ public class LinkProvider {
 
   public static final String PROFILE_DEFAULT_AVATAR_URL = "/eXoSkin/skin/images/avatar/DefaultUserAvatar.png";
   public static final String SPACE_DEFAULT_AVATAR_URL = "/eXoSkin/skin/images/system/SpaceAvtDefault.png";
+  public static final String ROOM_DEFAULT_AVATAR_URL = "/chat/img/room-default.jpg";
```
boubaker, line 62: we can't add dependency in social about chat addon (dependency between bottom layer to top layer)

IlhemEssaadi: > we can't add dependency in social about chat addon (dependency between bottom layer to top layer)

Ok, i will close this PR !

PR #1529, opened 2022-05-26
TASK-57505 fix the problem of searching for activities that have the same keyword in search application

montassarhamadi:
Problem: nothing is displayed when the search term is a keyword of one or many activities.
Fix: The problem was in the query string used to search for activities. Therefore, it should use a fuzzy query with edit distance 1 to catch all activities with one changed character.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ActivitySearchConnector.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ActivitySearchConnector.java
@@ -50,8 +50,22 @@ public class ActivitySearchConnector {
       "    \"query\": \"@term@~\"," +
       "    \"fuzziness\": 1," +
       "    \"phrase_slop\": 1" +
-          "  }" +
+      "  }" +
       "},";
+
+  public static final String            SEARCH_QUERY_WITH_PHRASE     = "\"must\":{" +
+      "  \"query_string\":{" +
+      "    \"fields\": [\"body\", \"posterName\"]," +
+      "    \"default_operator\": \"AND\"," +
+      "    \"query\": \"(@term@) OR (\\\"@phrase@\\\"~)^5\"," +
+      "    \"fuzziness\": 1," +
+      "    \"phrase_slop\": 1" +
+      "  }" +
+      "},";
+
+  private static final String           TERM_REPLACEMENT             = "@term@";
+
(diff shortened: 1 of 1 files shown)


component/core/src/main/java/org/exoplatform/social/core/jpa/search/ActivitySearchConnector.java
```diff
@@ -47,7 +47,7 @@ public class ActivitySearchConnector {
       "  \"query_string\":{" +
       "    \"fields\": [\"body\", \"posterName\"]," +
       "    \"default_operator\": \"AND\"," +
-      "    \"query\": \"@term@\"" +
+      "    \"query\": \"@term@"+ "~1" + "\"" +
```
boubaker: What if there are multiple words, please use `fuzziness` field instead as suggested before

```suggestion
      "    \"query\": \"@term@\"," +
      "    \"fuzziness\": 1," +
      "    \"phrase_slop\": 1" +
```

See https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-query-string-query.html#query-string-top-level-params

component/core/src/main/java/org/exoplatform/social/core/jpa/search/ActivitySearchConnector.java
```diff
@@ -47,8 +47,10 @@ public class ActivitySearchConnector {
       "  \"query_string\":{" +
       "    \"fields\": [\"body\", \"posterName\"]," +
       "    \"default_operator\": \"AND\"," +
-      "    \"query\": \"@term@\"" +
-      "  }" +
+      "    \"query\": \"@term@~\"," +
```
boubaker, line 50:
```suggestion
      "    \"query\": \"(@term@) OR (\\\"@phrase@\\\")^5\"," +
```
we may need to add pertinence/boost information for complete phrase match as well, WDYT ?

montassarhamadi: > we may need to add pertinence/boost information for complete phrase match as well, WDYT ?

in that case, will phrase get the value of search term??

boubaker: ok, changed for full proposition as discussed, thanks.

boubaker (approved):

PR #1531, opened 2022-05-30
TASK-55619: Fix bad display of emoji on mail subject

IlhemEssaadi:
ISSUES : When we post or like a comment or an activity that contains an emoji, the email received incorrectly displays the emoji in the email subject.
FIX : The problem is that the email subject can not display HTML entities. To fix it, just convert it with unescapeHtml4 in case of like and putting the literal character in the subject as it is without modification in case of post.

### DIFF
--- a/component/notification/src/main/java/org/exoplatform/social/notification/channel/template/MailTemplateProvider.java
+++ b/component/notification/src/main/java/org/exoplatform/social/notification/channel/template/MailTemplateProvider.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
+import org.apache.commons.lang3.StringEscapeUtils;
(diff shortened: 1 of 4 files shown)


component/notification/src/main/java/org/exoplatform/social/notification/channel/template/MailTemplateProvider.java
```diff
@@ -25,6 +25,7 @@
 import java.util.Locale;
 import java.util.Map;
 
+import liquibase.repackaged.org.apache.commons.text.StringEscapeUtils;
```
boubaker:
```suggestion
import org.apache.commons.lang3.StringEscapeUtils;
```

boubaker (dismissed):

boubaker (approved):

PR #1548, opened 2022-06-06
task-57941: Fix the title in SpaceInfos Portlet is positioned on the center

IlhemEssaadi:
FIX : The problem is that the title is centered and is positioned by the padding of the CSS class spaceInfosApp, fixed by moving the title outside this class and adding  vuetify spacing Property.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/space-infos-app/components/ExoSpaceInfos.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/space-infos-app/components/ExoSpaceInfos.vue
@@ -1,9 +1,9 @@
 <template>
   <v-app>
-    <div class="body-1 text-uppercase text-sub-title">{{ $t("social.space.description.title") }}</div>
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/space-infos-app/components/ExoSpaceInfos.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <v-app>
-    <div class="body-1 text-uppercase text-sub-title">{{ $t("social.space.description.title") }}</div>
+    <div class="body-1 text-uppercase text-sub-title mx-n1">{{ $t("social.space.description.title") }}</div>
```
ahamdi: Could you check where the margin is added and fix it from there instead of adding a negative margin ?

ahamdi (changes requested):

ahamdi (approved):

PR #1549, opened 2022-06-07
Task-161: Fix redirect to the space when deleting an activity

SaraBoutej:
Prior to this change when deleting an activity from the page activity details a new message with "Activity not found" is displayed.
After this change when deleting an activity, it will be redirect to to the space's home page if it is a space activity, else it will be redirect to the default home link.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/list/ActivityStreamList.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/list/ActivityStreamList.vue
@@ -127,7 +127,9 @@ export default {
   created() {
     document.addEventListener('activity-deleted', event => {
       const activityId = event && event.detail;
-      this.isDeleted = true;
+      if (this.isActivityDetail) {
+        this.isDeleted = true;
+      }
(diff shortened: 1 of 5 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/list/ActivityStreamList.vue
```diff
@@ -126,6 +127,7 @@ export default {
   created() {
     document.addEventListener('activity-deleted', event => {
       const activityId = event && event.detail;
+      this.isDeleted = true;
```
boubaker:
```suggestion
  if (this.isActivityDetail) {
    this.isDeleted = true;
  }
```
If you have one single activity inside the stream of a space and then delete it, we should display `activity-stream-empty-message-space` and not an empty content. consequently, we need to change this variable only when we are in a context of single activity display.

SaraBoutej: done!!

AzmiTouil (approved):
LGTM

PR #1550, opened 2022-06-08
Task-176: Review the activity three dots menu UI

SaraBoutej:
Prior to this change, the UI of the three dots menu in the activity menu has not the same UI as  other three dots menu in the platform. This change add a new font awesome icon for the activity menu and edit the color as the others menu.


AzmiTouil (approved):
LGTM

PR #1551, opened 2022-06-08
Task-178: kudos counter (mobile view)

AzmiTouil:
Prior to this change, the kudos count in the mobile view was displayed right next to the kudos icon.
This PR will update the display of the number of kudos, so it will have the same behavior as the comments and the likes in the mobile view.


webapp/portlet/src/main/webapp/vue-apps/activity-reactions/components/ActivityReactions.vue
```diff
@@ -81,6 +80,10 @@ export default {
       type: Number,
       default: 0
     },
+    kudosNumber: {
```
boubaker: In social layer, we don't have kudos, you will have to add an `<extension-registry-components` to allow its extensibility

boubaker (approved):

PR #1553, opened 2022-06-08
BUILDERS-185: Edit favorite icon display when not bookmarked

SaraBoutej:
In this PR we change the favorite icon when not bookmarked fas fa-star by far fa-star


boubaker (approved):

SaraBoutej:
In this PR we change the favorite icon when not bookmarked fas fa-star by far fa-star

PR #1558, opened 2022-06-09
Task-57653: Fix copy paste url of search  from the address bar

sofyenne:
ISSUE : when the user copy the Url of a search result and paste it in a new tab no result displayed , the search term is null 
FIX : extract the search term  from the pasted url  .


ahamdi (approved):

Jihed525:
@sofyenne branch should be moved to patch/6.2.5(based on 6.2.5) as we can't get the dep needed on stable/6.2.x
would it be ok @ahamdi ?

PR #1576, opened 2022-06-15
Task-58389: Button send request is not updated

hchaabene:
Problem: when connect to user in a space, so button is not updated to 'cancel request' , the problem is cahce control http 
Fix: i added 'cache-control' in headers of services .

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/common/js/SpaceService.js
+++ b/webapp/portlet/src/main/webapp/vue-apps/common/js/SpaceService.js
@@ -14,9 +14,13 @@ export function getSpaceTemplates() {
 }
 
 export function getSpaceMembers(query, offset, limit, expand, role, spaceId) {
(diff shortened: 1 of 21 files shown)


webapp/portlet/src/main/webapp/vue-apps/common/js/SpaceService.js
```diff
@@ -14,9 +14,13 @@ export function getSpaceTemplates() {
 }
 
 export function getSpaceMembers(query, offset, limit, expand, role, spaceId) {
+  const headers = new Headers();
+  headers.append('pragma', 'no-cache');
+  headers.append('cache-control', 'public, max-age=31536000, immutable, no-store');
```
boubaker, line 19: Why disabling cache management from UI ? If the cache lifecycle isn't well managed in Backend, better fix it in backend instead of doing this workaround.

PR #1579, opened 2022-06-15
MEED-265 : add a new prop to the activity-rich-editor allowing the use of extended plugins 

GouadriaHanen:
Before this change, only CKEditor of Type 'ActivityContent' or its type starts with 'comment_' can extend additional plugins. 
after this change, a dedicated parameter has been defined to allow reusing the extended plugins of Activity Rich Editor.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/content/ActivityCommentRichText.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/content/ActivityCommentRichText.vue
@@ -18,7 +18,7 @@
           :activity-id="activityId"
           :template-params="templateParams"
           suggestor-type-of-relation="mention_comment"
-          add-extra-plugin
+          use-extra-plugins
           autofocus
           @ready="handleEditorReady" />
         <extension-registry-components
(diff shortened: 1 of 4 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/content/ActivityCommentRichText.vue
```diff
@@ -18,6 +18,7 @@
           :activity-id="activityId"
           :template-params="templateParams"
           suggestor-type-of-relation="mention_comment"
+          add-extra-plugin
```
boubaker:
```suggestion
          use-extra-plugins
```
plural

boubaker (commented):
Please consider using a meaningful title

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/content/ActivityCommentRichText.vue
GouadriaHanen: done!

boubaker:
Description to change IMO

> Before this change, only CKEditor of Type 'ActivityContent' or its type starts with 'comment_' can extend additional plugins.
> after this change, any component user adds a "useExtraPlugins", which is false by default, as a prop to allow him to use additional plugins

...
after this change, a dedicated parameter has been defined to allow reusing the extended plugins of Activity Rich Editor.

boubaker (approved):

PR #1581, opened 2022-06-16
MEED-256: "New activities" information display and behavior

AzmiTouil:
Prior to this change, when a new activity is added to the space stream, the new activity information appears, but I am not sent to the top of the stream portlet when clicking on button

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/list/ActivityStreamList.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/list/ActivityStreamList.vue
@@ -229,10 +229,10 @@ export default {
       this.limit += newActivitiesCount;
       this.loadActivityIds().catch(() => window.location.reload());
       this.$nextTick().then(() => {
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/list/ActivityStreamList.vue
```diff
@@ -228,6 +228,17 @@ export default {
     loadActivities(newActivitiesCount) {
       this.limit += newActivitiesCount;
       this.loadActivityIds().catch(() => window.location.reload());
+      this.$nextTick().then(() => {
+        const $streamPageContainer = $('#ActivityStream')[0];
```
boubaker:
```suggestion
        const streamPageContainerElement = document.getElementById('ActivityStream');
```
it's better to not use jQuery anymore, we want to deprecate it.

AzmiTouil: Updated!

boubaker (approved):

PR #1583, opened 2022-06-17
Task-58389: Button send request is not updated

hchaabene:
Problem: when connect to user in a space, so button is not updated to 'cancel request' , the problem is cahce control http
Fix: i added 'cache-control' in headers of services .

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
@@ -722,36 +722,36 @@ public class SpaceRestResourcesV1 implements SpaceRestResources {
     }
 
     long cacheTime = space.getCacheTime();
-      if (StringUtils.isBlank(role)) {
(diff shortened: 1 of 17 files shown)


component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
```diff
@@ -755,11 +750,14 @@ public Response getSpaceMembers(@Context UriInfo uriInfo,
...
-      builder.expires(new Date(cacheTime));
-    }
+      EntityTag eTag  = new EntityTag(String.valueOf(collectionUser.hashCode()), true);
+      Response.ResponseBuilder builder = request.evaluatePreconditions(eTag);
+      if (builder == null) {
+        builder = Response.ok(collectionUser, MediaType.APPLICATION_JSON);
+        builder.tag(eTag);
+        builder.lastModified(new Date(cacheTime));
+        builder.expires(new Date(cacheTime));
+      }
```
boubaker: code format

ahamdi (approved):

PR #1589, opened 2022-06-20
TASK-58523: Fix bookmark an item from the unified search (#1588)

Jihed525:
Prior to this change, when sending an empty spaceId, a parsing error happens in REST endpoint that leads to an HTTP 404 (This may happen on items added outside Spaces).
This fix will ensure to send a 0 in case of empty spaceId to avoid such an error.

(cherry picked from commit ef8db07892d3b4388ae9f6c0dee605b557e405da)


ahamdi:
@Jihed525 We do not need a PR for Maintenance as the issue will be fixed by Meeds team

Jihed525:
> @Jihed525 We do not need a PR for Maintenance as the issue will be fixed by Meeds team

you are absolutely right @ahamdi , I should have backported it to stable/6.3.x

PR #1603, opened 2022-06-21
task-48748: pass user enabled/disabled/deleted status to extension registry components

hthabti:
Prior to this change it was possible to perform actions with deleted/disabled users
fix:
 - We will pass the user status enabled/disabled/deleted to the other components injected with the extension registry to allow components to set specific behavior depending on the user status.


boubaker:
we don't have kudos in social project, please change the title to mention what behavior you fixed in social.

webapp/portlet/src/main/webapp/vue-apps/common/components/ExoUserAvatar.vue
```diff
@@ -267,12 +270,15 @@ export default {
       return {
```
ahamdi, line 270: Where do we use this computed property ?

ahamdi, line 273: Could you update the title to reflect the change ? 
Suggestion : 
Pass user enabled/disabled status to extension registry components

ahamdi (approved):

PR #1612, opened 2022-06-22
MEED-217: Access to comments (Link)

AzmiTouil:
Prior change, user cannot share comment link.
Update : Add browser option for comment timestamp to copy link and open in new tab

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/footer/ActivityCommentTime.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/footer/ActivityCommentTime.vue
@@ -10,6 +10,7 @@
           link
           text
           plain
+          @click="openCommentsDrawer"
           v-bind="attrs"
           v-on="on">
           <relative-date-format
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/footer/ActivityCommentTime.vue
```diff
@@ -4,12 +4,12 @@
...
         <v-btn
           :height="20"
+          :href="commentLink"
           class="hover-underline width-auto text-capitalize-first-letter d-inline px-0"
           x-small
           link
           text
           plain
-          @click="openCommentsDrawer"
           v-bind="attrs"
```
boubaker: In fact, you can keep this method to not refresh the page when clicking on it, by doing this:

```suggestion
          @click="openCommentsDrawer"
          v-bind="attrs"
```
and making the method like this:

```javascript
    openCommentsDrawer(event) {
      if (event) {
         event.preventDefault();
         event.stopPropagation();
      }
      document.dispatchEvent(new CustomEvent('activity-comments-display', {detail: {
        activity: this.activity,
        selectedCommentId: this.commentId,
        selectedActivityId: this.activity.id,
        offset: 0,
        limit: 200, // To display all
      }}));
  }
```

This way, you will benefit from two options:
1. Make the comment access link (that can be copied and highlighted)
2. When clicking on it, the page isn't refreshed and the drawer is opened with highlight to comment

AzmiTouil: Done!

boubaker (approved):

PR #1613, opened 2022-06-23
Task-58137: Add space and groups suggester for how can make a request…

lbengayed:
… process

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/common/components/ExoIdentitySuggester.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/common/components/ExoIdentitySuggester.vue
@@ -115,6 +115,12 @@ export default {
         return false;
       },
     },
+    includeGroups: {
+      type: Boolean,
+      default: function() {
+        return false;
+      },
+    },
     noRedactorSpace: {
       type: Boolean,
       default: function() {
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/common/components/ExoIdentitySuggester.vue
```diff
@@ -239,12 +245,13 @@ export default {
...
             this.loadingSuggestions = 0;
             this.items = [];
-            this.$suggesterService.searchSpacesOrUsers(value,
+            this.$suggesterService.searchSpacesOrUsersOrGroups(value,
               this.items,
               this.typeOfRelations,
               this.searchOptions,
               this.includeUsers,
               this.includeSpaces,
+              this.includeGroups,
```
boubaker: It's better to not break the API by adding an additional parameter in the middle. Besides, you may need to add a new method inside `suggesterService.js` with a config object (instead of chained parameters with all options listed there) : 

this.$suggesterService.search({
  term: String,
  typeOfRelations: String,
  searchOptions: Object,
  includeUsers: Boolean,
  includeSpaces: Boolean,
  includeGroups: Boolean,
  onlyRedactor: Boolean,
  onlyManager: Boolean,
  loadingCallback: Function,
  successCallback: Function,
  errorCallback: Function,
})

mkrout: +1

mkrout (approved):
LGTM

PR #1614, opened 2022-06-23
TASK-58025: Remove all related anti-malware code

hakermi:
Prior to this change, The anti-malware base code was implemented inside the social module,
This PR should remove all its related code after extracating it into an addon


azayati (approved):
LGTM

PR #1623, opened 2022-06-27
Task-54820: unify the Title Style of the External spaces list portlet

wassimmhamdi:
Problem: The external spaces portlet title style is different than the style of the other widgets in the snapshot page
Fix: Unify the title style of external spaces portlet to be coherent with the other portlet's styles


ahamdi:
@wassimmhamdi What isue should be fixed here ? the styles added are already set on the parent ? what style is still missing ?

ahamdi (dismissed):

ahamdi (approved):

PR #1627, opened 2022-06-27
MEED-217: Access to comments (Link)

AzmiTouil:
Prior change, user cannot share comment link.
Update : Add browser option for comment timestamp to copy link and open in new tab

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/list/ActivityCommentsDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/list/ActivityCommentsDrawer.vue
@@ -99,6 +99,20 @@ export default {
     },
   },
   created() {
+    const search = document.location.search.substring(1);
+    if (search) {
+      const parameters = JSON.parse(
+        `{"${decodeURI(search)
+          .replace(/"/g, '\\"')
+          .replace(/&/g, '","')
+          .replace(/=/g, '":"')}"}`
+      );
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/list/ActivityCommentsDrawer.vue
```diff
@@ -195,15 +215,25 @@ export default {
...
-              if (this.$refs.activityCommentsDrawer) {
-                this.$refs.activityCommentsDrawer.open();
-              }
-            });
+            this.$nextTick().then(() => this.openDrawer());
           }
         });
       }
     },
+    displaySelectedActivityComment(commentId, activityId) {
```
boubaker: comment highlighting already work (you can try [this link](https://community.exoplatform.com/portal/dw/activity?id=397094#comment-comment397184) by example)
Why adding a new method to highlight it and display comments drawer ?

AzmiTouil: It is not for highlight comment, it is just to open comment drawer from url, since we need to get activity by Id

AzmiTouil: Updated!

boubaker (approved):

PR #1649, opened 2022-07-06
TASK-58701: remove padding from mobile drawer

Amine0008:
ISSUE: the rounded top right corner isn't visible
remove unnecessary padding for mobile drawer


mkrout (approved):
LGTM

PR #1677, opened 2022-07-18
MEED-175: Filter the stream display (no result)

AzmiTouil:
This PR will improve the display of the stream filter when no result found


SaraBoutej (approved):
LGTM

PR #1679, opened 2022-07-19
MEED-384: People suggester UX

AzmiTouil:
Improve people suggester service by list result only once the user have finished typing letters


SaraBoutej (approved):
LGTM

PR #1683, opened 2022-07-20
MEED-394: People suggestor in invitation when members can't be found

AzmiTouil:
Inform user when user can't be found when using people suggestor


SaraBoutej (approved):
LGTM

PR #1692, opened 2022-07-26
TASK-59079: Hide the delete icon when the default banner is set

akhanfir:
Prior to this change, When the default banner is displayed on a profile, the delete icon display.
To fix this,add the method update , assign the value of isDefaultBanner.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/profile-header/components/ProfileHeaderBannerButton.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/profile-header/components/ProfileHeaderBannerButton.vue
@@ -56,6 +56,9 @@ export default {
     sendingImage: false,
     isDefaultBanner: false,
   }),
+  updated() {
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/profile-header/components/ProfileHeaderBannerButton.vue
```diff
@@ -56,6 +56,9 @@ export default {
     sendingImage: false,
     isDefaultBanner: false,
   }),
+  updated() {
+    this.isDefaultBanner=this.user.banner.startsWith('/portal/rest/v1/social/users/default-image/');
```
ahamdi, line 60: this.isDefaultBanner = this.user.banner.startsWith('/portal/rest/v1/social/users/default-image/');

ahamdi (approved):

PR #1693, opened 2022-07-26
MEED-383: Stream filter is saved automatically

AzmiTouil:
This PR will ensure that the stream filter applied by the user will be automatically saved in storage.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/list/ActivityStreamList.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/list/ActivityStreamList.vue
@@ -113,7 +113,7 @@ export default {
   created() {
     this.streamFilter = localStorage.getItem('activity-stream-stored-filter');
     if (this.streamFilter) {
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/list/ActivityStreamList.vue
```diff
@@ -111,6 +111,10 @@ export default {
     },
   },
   created() {
+    this.streamFilter = localStorage.getItem('activity-stream-stored-filter');
+    if (this.streamFilter) {
+      document.dispatchEvent(new CustomEvent('activity-stream-stored-filter-applied', {detail: this.streamFilter}));
```
boubaker: We use `document.addEventListener` to allow dispatching events outside the application. When the event is about an internale mechanism of the application, it's better to use `this.$root.$emit` and `this.$root.$on` to simplify event management.

AzmiTouil: Done!

boubaker (approved):

PR #1715, opened 2022-08-03
TASK-59071: Empty Email notification info area , after sharing a post/article to another space

hchaabene:


component/api/src/main/java/org/exoplatform/social/core/manager/ActivityManager.java
```diff
@@ -46,16 +46,20 @@ public interface ActivityManager {
...
+   * +   * set.
+   * +   *
+   * +   * @param streamOwner The activity stream owner.
+   * +   * @param activity The activity to be saved.
+   * +   * @param broadcast if true broadcast the event.
+   * +   * @LevelAPI Platform
+   * +   * @since 6.4.0-GA
+   * +
+   */
+  default void saveActivityNoReturn(Identity streamOwner, ExoSocialActivity activity, boolean broadcast) {
```
boubaker, line 60: No API change please. You must define a new method allowing to choose whether to broadcast events or not.
In addition, the Javadoc format isn't good.

PR #1721, opened 2022-08-08
MEED-363-421: Identity popover actions data collection

AzmiTouil:
This PR will assure collecting data for access to profile action from user and space popover


webapp/portlet/src/main/webapp/vue-apps/popover/components/SpacePopoverContent.vue
```diff
@@ -143,6 +144,14 @@ export default {
         }
       }
     },
+    spaceAccess() {
+      document.dispatchEvent(new CustomEvent('identity-profile-access', {
```
boubaker: should use URLs to access pages instead of user actions

SaraBoutej (approved):

PR #1727, opened 2022-08-10
Task-59071: Empty Email notification info area , after sharing a post/article to another space

hchaabene:
Prior this change, when sharing post from space to other space , template mail not displayed the title or summary of post where shared , so the template email is applied to add a short message or an article ,
Add new notification shared for Activity (short message | Article) and implement the basic plugins for shared activity.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/manager/ActivityManagerImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/manager/ActivityManagerImpl.java
@@ -251,7 +251,7 @@ public class ActivityManagerImpl implements ActivityManager {
     }
     checkCanShareActivityToSpaces(spaceIds, viewer);
     activityShareAction.setUserIdentityId(Long.parseLong(viewerIdentity.getId()));
-    List<ExoSocialActivity> sharedActivities = createShareActivities(activityTemplate, activityShareAction, viewerIdentity.getId());
+    List<ExoSocialActivity> sharedActivities = createShareActivities(activityTemplate, activityShareAction, viewerIdentity.getId(),activity);
     Set<Long> sharedActivityIds = sharedActivities.stream()
                                                    .map(tmpActivity -> Long.parseLong(tmpActivity.getId()))
                                                    .collect(Collectors.toSet());
@@ -853,7 +853,8 @@ public class ActivityManagerImpl implements ActivityManager {
 
   private List<ExoSocialActivity> createShareActivities(ExoSocialActivity activityTemplate,
                                                         ActivityShareAction activityShareAction,
-                                                        String viewerIdentityId) {
+                                                        String viewerIdentityId,
+                                                        ExoSocialActivity activity) {
     String title = activityTemplate == null || activityTemplate.getTitle() == null ? "" : activityTemplate.getTitle();
     String type = activityTemplate == null ? null : activityTemplate.getType();
     Map<String, String> templateParams = activityTemplate == null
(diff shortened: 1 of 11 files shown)


component/api/src/main/java/org/exoplatform/social/core/activity/model/ExoSocialActivity.java
```diff
@@ -446,4 +446,7 @@ public interface ExoSocialActivity extends CacheEntry {
    * @return true if Activity has a specific {@link MetadataObject} type
    */
   boolean hasSpecificMetadataObject();
+
+  List<String> getTargetSpaces();
+  void setTargetSpaces(List<String> targetSpaces);
```
ahamdi: There is no need to add this properties in the activity model IMO

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
```diff
@@ -689,7 +689,7 @@ public Response shareActivity(
                                 @QueryParam("expand")
                                 String expand,
                                 @RequestBody(description = "Share target spaces", required = true)
-                                ActivityEntity model) {
+                                ExoSocialActivityImpl model) {
```
ahamdi: Could you use ActivityEntity and add needed property there if necessary.
What is the difference between summary and body of an activity ?

webapp/portlet/src/main/webapp/vue-apps/common/components/ActivityShareDrawer.vue
```diff
@@ -112,9 +114,16 @@ export default {
       this.description = '';
       this.sharing = false;
     },
-    open(activityId, currentApp) {
-      this.activityId = activityId;
+    open(activity, currentApp) {
+      this.activityId = activity.id;
       this.currentApp = currentApp;
+      if (typeof(activity.news)  !== 'undefined') {
```
ahamdi: we can not mention News here , as it is an external addon. we need to manage to publish activities based on their content and unify how they will render depending on their contents.

webapp/portlet/src/main/webapp/vue-apps/common/js/ActivityService.js
```diff
@@ -91,6 +91,8 @@ export function shareActivity(activityId, message, templateParams, spaces) {
     method: 'POST',
     body: JSON.stringify({
       title: message,
+      body: title,
+      summary: summary,
```
ahamdi: we may send the ID of the original activity and then get its title & summary from backend with no need to send them with the shared activity inofmration

ahamdi (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
hchaabene: in news entity has a title, summary and body 
in ExoSocialActivityImpl i can get all information about shared activity  but ActivityEntity not contain summary for shared activity to passed in email template

webapp/portlet/src/main/webapp/vue-apps/common/js/ActivityService.js
hchaabene: this two information title and summary of activity shared is not recovred in https://github.com/Meeds-io/social/blob/012a8090b96985d9b6663aaca9749dd27336dc2e/component/service/src/main/java/org/exoplatform/social/rest/entity/ActivityEntity.java

hchaabene: title i can get it from model but summary is not stored in ActivityEntity , i need to add in Model!

ahamdi (approved):

PR #1733, opened 2022-08-15
MEED-339: Display the placeholder in the activity and comment composer

SaraBoutej:
Prior to this change, the placeholder in the activity editor is not visible because we change the ckeditor enter mode to DIV,

This change allows to add a new div contains the placeholder text ("Share your thoughts here"), and it will be shown only when the edited text is empty.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/content/ActivityCommentRichText.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/content/ActivityCommentRichText.vue
@@ -14,7 +14,7 @@
           v-model="message"
           :ck-editor-type="ckEditorId"
           :max-length="$activityConstants.COMMENT_MAX_LENGTH"
-          :placeholder="$t('activity.composer.placeholder', {0: $activityConstants.COMMENT_MAX_LENGTH})"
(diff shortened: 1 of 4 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/comment/content/ActivityCommentRichText.vue
```diff
@@ -14,7 +14,7 @@
           v-model="message"
           :ck-editor-type="ckEditorId"
           :max-length="$activityConstants.COMMENT_MAX_LENGTH"
-          :placeholder="$t('activity.composer.placeholder', {0: $activityConstants.COMMENT_MAX_LENGTH})"
+          :placeholder="$t('activity.newComposer.placeholder')"
```
boubaker: It's not new anymore IMO. It would be better to avoid adding `new` inside labels

```suggestion
          :placeholder="$t('activity.composer.placeholder')"
```

SaraBoutej: Done!

boubaker (approved):

PR #1740, opened 2022-08-17
MEED-368: Review activity composer editor UX when clicking outside

SaraBoutej:


webapp/portlet/src/main/webapp/vue-apps/common/components/ExoActivityRichEditor.vue
```diff
@@ -257,6 +257,9 @@ export default {
           change: function (evt) {
             const newData = evt.editor.getData();
             self.inputVal = newData;
+            if ( !self.activityId ) {
```
boubaker:
```suggestion
            if (!self.activityId) {
```

boubaker (approved):

PR #1752, opened 2022-08-23
MEED-455 Shared activity display when hidden space

AzmiTouil:
Prior this change, two labels "Hidden Space" and "Space name" appear on the activity when a platfrom admin post a activity in a hidden space(not a membre of).
Regression due to space popover refactoring.


sarahboutej (approved):
LGTM

SaraBoutej (approved):

PR #1756, opened 2022-08-25
MEED-456 Bug: Set a space not a member of as a favorite shouldn't be possible

AzmiTouil:
Prior this change, not a member user can mark space a favorite by executing a fetch command
Fix : Add restriction to API


SaraBoutej (approved):
LGTM

PR #1760, opened 2022-08-30
MEED-445 Refresh stream when an activity is unbookmarked

AzmiTouil:
This PR allows dynamic updating of the stream when an activity is unbookmarked


SaraBoutej (approved):
LGTM

PR #1780, opened 2022-09-07
TASK-58326: Can't accept user invitation from All notifications page

akhanfir:
Prior to this change, when click on the button to accept user invitation from All notifications page . To fix this, change the click method to pass URL and Id of the invitation.

### DIFF
--- a/webapp/portlet/src/main/webapp/js/eXo/social/notification/WebNotification.js
+++ b/webapp/portlet/src/main/webapp/js/eXo/social/notification/WebNotification.js
@@ -19,10 +19,7 @@
       appendMessage : function(message) {
         WebNotification.evalMethod('appendMessage', message);
       },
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/js/eXo/social/notification/WebNotification.js
```diff
@@ -19,8 +19,15 @@
       appendMessage : function(message) {
         WebNotification.evalMethod('appendMessage', message);
       },
-      doAction : function(elm) {
-        WebNotification.evalMethod('doAction', elm);
+      doAction : function(id, restURL) {
```
ahamdi, line 22: What is the difference between doAction and doCancelAction ?

ahamdi (approved):

PR #1784, opened 2022-09-07
MEED-457 Space host or redactor can pin an activity

AzmiTouil:
This PR will allow to space host/redactor to pin or unpin activity(ies).


SaraBoutej (approved):
LGTM

PR #1787, opened 2022-09-09
Task-48748: Pass user enabled/disabled/deleted status to extension registry components

Jihed525:
Prior to this change it was possible to perform actions with deleted/disabled users fix:

We will pass the user status enabled/disabled/deleted to the other components injected with the extension registry to allow components to set specific behavior depending on the user status.

(cherry picked from commit 48e16f481e63e61eef9697b78a8c0416dd16f246)


azayati (approved):
LGTM

PR #1790, opened 2022-09-12
TASK-58326: Can't accept user invitation from All notifications page

akhanfir:
Prior to this change, we can not accept user invitation from All notifications page.  To fix that, we have changed the click method to pass the URL and Id of the invitation.


azayati (approved):
LGTM

PR #1826, opened 2022-09-22
fix: Space administrator still a space member when he deletes his own space membership from group management - EXO-58785 - Meeds-io/meeds#310

azayati:
Prior to this change, a spaces administrator still a space member even if deletes the corresponding space membership from group management. This is caused by SpaceUtils.isUserHasMembershipTypesInGroup method used to check if a user has a membership in a group which considers any spaces administrator a member of any space. After this commit, we will instead perform the check with Identity.isMemberOf method which will enhance performances.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/listeners/SocialMembershipListenerImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/listeners/SocialMembershipListenerImpl.java
@@ -16,22 +16,22 @@
  */
 package org.exoplatform.social.core.listeners;
 
+import java.util.List;
+
 import org.exoplatform.commons.utils.CommonsUtils;
 import org.exoplatform.container.ExoContainer;
 import org.exoplatform.container.ExoContainerContext;
 import org.exoplatform.portal.config.UserACL;
-import org.exoplatform.services.organization.*;
+import org.exoplatform.services.organization.Membership;
+import org.exoplatform.services.organization.MembershipEventListener;
+import org.exoplatform.services.organization.MembershipTypeHandler;
+import org.exoplatform.services.organization.OrganizationService;
+import org.exoplatform.services.organization.User;
 import org.exoplatform.services.security.ConversationState;
-import org.exoplatform.social.core.identity.model.Identity;
-import org.exoplatform.social.core.identity.model.Profile;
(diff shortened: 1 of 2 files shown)


component/core/src/main/java/org/exoplatform/social/core/listeners/SocialMembershipListenerImpl.java
```diff
@@ -65,10 +65,9 @@ public void postDelete(Membership m) throws Exception {
...
           space.setEditor(state.getIdentity().getUserId());
         }
-        boolean hasAllMembership = SpaceUtils.isUserHasMembershipTypesInGroup(m.getUserName(), m.getGroupId(), MembershipTypeHandler.ANY_MEMBERSHIP_TYPE);
-        boolean hasManagerMembership = hasAllMembership || SpaceUtils.isUserHasMembershipTypesInGroup(m.getUserName(), m.getGroupId(), acl.getAdminMSType());
-        boolean hasMemberMembership = hasManagerMembership || SpaceUtils.isUserHasMembershipTypesInGroup(m.getUserName(), m.getGroupId(), SpaceUtils.MEMBER);
-
+        boolean hasAllMembership = orgService.getMembershipHandler().findMembershipByUserGroupAndType(m.getUserName(), m.getGroupId(), MembershipTypeHandler.ANY_MEMBERSHIP_TYPE) != null;
+        boolean hasManagerMembership = hasAllMembership || orgService.getMembershipHandler().findMembershipByUserGroupAndType(m.getUserName(), m.getGroupId(), acl.getAdminMSType()) != null;
+        boolean hasMemberMembership = hasManagerMembership || orgService.getMembershipHandler().findMembershipByUserGroupAndType(m.getUserName(), m.getGroupId(), SpaceUtils.MEMBER) != null;
```
boubaker: This will cause a **big performance degradation**, please consider adding a dedicated listener when a , instead to update user identity inside current ConversationState using IdendityRegistry, which is already made by the listener

https://github.com/Meeds-io/gatein-portal/blob/d8e2afb2d40f5456d19f5e5528dd311a664cadfc/web/portal/src/main/webapp/WEB-INF/conf/organization/organization-configuration.xml#L348-L352

The fix will be to ensure that this Listener do the Job of updating ACL Identity of user after deleting him from space.

boubaker (approved):

PR #1837, opened 2022-09-26
feat: Pinned activities filter - MEED-606

GouadriaHanen:
after this change in the general stream, a new filter is added "Pinned activities"  which allows filtering of the activities pinned in the spaces of which I am member.


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityStreamFilter.vue
```diff
@@ -85,6 +85,10 @@ export default {
       return [{
         text: eXo.env.portal.spaceId && this.$t('activity.filter.anyActivity') || this.$t('activity.filter.all'),
         value: 'all_stream',
+      },{
+        text: this.$t('activity.filter.pinnedActivities'),
+        value: 'user_pin_stream',
+        enable: !eXo.env.portal.spaceId,
```
AzmiTouil: should not be enabled when pin feature is disabled

AzmiTouil (changes requested):

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityStreamFilter.vue
GouadriaHanen: done!

AzmiTouil (approved):

PR #1840, opened 2022-09-27
feat: left menu review Update the icons for space apps - MEED-559 - Meeds-io/MIPs#16

SaraBoutej:
This change allows to edit the icons in the space bottom navigation menu on mobile devises from PLF icons to font-awesome icons.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/space-menu/components/SpaceMenu.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/space-menu/components/SpaceMenu.vue
@@ -11,15 +11,10 @@
           background-color="transparent"
           class="spaceButtomNavigationParent"
           flat>
-          <v-btn
+          <space-menu-item
             v-for="nav in navigations"
             :key="nav.id"
-            :value="nav.uri"
-            :href="nav.uri"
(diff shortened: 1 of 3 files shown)


component/core/src/main/java/org/exoplatform/social/core/space/SpaceUtils.java
```diff
@@ -1937,10 +1937,9 @@ public static String getIconClass(UserNode node) {
       return null;
     }
     if (isHomeNavigation(node)) {
-      return "uiIconAppSpaceHomePage uiIconDefaultApp";
+      return "fa-stream";
     }
-    String appName = node.getPageRef().getName();
-    return "uiIconApp" + node.getName() + " uiIconApp" + appName + " uiIconDefaultApp";
+    return "fa-icon-" + node.getName();
```
boubaker: No need to make it variablized, but you can simply make a computed value to get the icon to use for each page or, you can also inherit CSS class in less instead of copying the properties value.

SaraBoutej: done

AzmiTouil (approved):

PR #1858, opened 2022-10-06
 fix: Hide delete icon and add icon when banner is not hovered over - EXO-60111 - Meeds-io/meeds#328

akhanfir:
Prior to this change, delete and edit icons are always displayed if mouse isn't over.  After this change, we ensure that delete and edit icons shouldn't be displayed if mouse isn't over.


azayati (approved):
LGTM

PR #1862, opened 2022-10-07
feat: Pin an activity Space host or redactor can pin/unpin an activity to/from the space stream - MEED-457 - Meeds-io/meeds#14

### LINKED ISSUE #14: Pin an activity
## Rationale

Details of the GAP ( problem(s) to address ) expressed by few words using some screenshots.

Some of the space hosts need to hilight activity for their space. The problem is that it can be done currently and important message cannot be easily identified.
While it needs to be for example read by all the members, etc.
Same thing when a new comer to the space discovers the community, it could be nice to check what's to know first.

So the idea here is to propose to space host or redactor when identified to pin an activity to the top of the space stream.

## 1. Functional requirements
- A space host or a redactor of a given space can pin/unpin an activity to make it displayed on top of the space activity stream
- The button pin/unpin will be displayed in activity menu for a manager/redactor of a space only
- A pinned activity might be easily identified so users can see it right away
- More than one activity can be pinned in a space

Design for information: https://xd.adobe.com/view/66d1a30d-2e41-4557-b4a0-dd1464e13fd7-3382/screen/5fce51b6-df31-4a7a-b17f-b6761bf78495/

- This feature has to be disabled by default and enabled through a property (Feature flag): Set to false by default - exo.feature.PinActivity.enabled

- In the future, it can be useful to propose a pinned activities list in the general stream so it is easy to check what's in the spaces I am member of. Here are the informations to be displayed in this list:
-- Activity / activity-type (poll, challenge, note, activity, etc.)
-- Space where this has been posted initially
-- Author of the activity to know who has posted this initially
-- Date of the activity
This list should be sorted by the time it has been pinned. Meaning that if I pin an activity today, on September, 5th while the activity has been posted two months ago, this activity will be displayed in the first position in the list. That way, hosts and redactors will have easy ability to order the pinned activities list.

For now, activityAuthor will be displayed in the pinned activities to hilight contributions from any members. 

In the future obviously, after some usage, it could be interesting to add settings for this portlet so the admin can decide to change the sort of the pinned activities (pinDate, activityDate) and to change the display of authors for communication policies (display the pinAuthor instead of the activityAuthor). For example, when the author is not active anymore in the platform neither in the organization.

For your informations, notifications won't be sent when activity is pinned, but it would be interesting to have this information in addition to the pin author so the notification receiver understand who has done this action.

pinned activities in the stream

### 2.1. Performance requirement
The stream rendering has to be as performant as current rendering performances. This means, we shouldn't change the number of RDBMS queries nor computing in Service layer.
### 2.2. Security requirement
Only space redactor or manager can pin or unpin an activity. If an activity is pinned by a manager/redactor, any other manager/redactor can unpin it or replace it by another pinned activity. The fine security check will be made on Service layer only with a 403 HTTP response code when user is not allowed.
### 2.3. Extensibility requirements
The display of the pinned activity(ies) has to be made in a dedicated Vue component that has to be extensible (without having to redefine the whole activity stream component).
### 2.4 Configurability requirements
We have to allow to pin multiple activities through a system (or exo.properties) property. The default value of allowed activities to pin will be 1.

### 2.5 Feature Flag (property name, default status, target audience, functional behaviour)
A feature flag is applied for the feature. And FYI, as soon as the feature is available in develop, then a flag will be set to true by default (`exo.feature.PinActivity.enabled`):
- meeds package: set to true by default
- others: set to true by default
That way, if other (i.e eXo) wants to disable the feature by default, that won't change the meedslab way to deliver features.

This feature includes:
- ability for hosts and redactors to pin activities in their space
- ability for users to filter their general stream by pinned activities
- ability for analytics admin to know the usage of pinning an activity

## 3. Impacts
- #### Documentation to update
- #### Feature flag to false by default

### 4.1. Security
The REST endpoints to access activities will remains as it is and doesn't need to be changed with an additional security check.
The new endpoints that will allow to pin/unpin activities has to consider user role inside the space to make a security check in Service layer.. The REST endpoints has to be accessible to all internal users (no external user is allowed).
### 4.2. REST API
Two endpoints has to be developed for pin & unpin operations.

The URL of REST endpoints are :

- **POST** /portal/rest/v1/social/activities/{activityId}/pins : to pin an activity
- **DELETE** /portal/rest/v1/social/activities/{activityId}/pins : to unpin an activity

The Activity DTO has to be modified to add:

- `pinned` boolean attribute to indicated whether the activity is pinned in activity stream of space or not
- `pinDate` String attribute with ISO8601 Date format for date of pin

When the user pins an activity on a space that have already reached the maximum pinned activities, a **HTTP 400** response code with text error message (`maximum_pinned_activities_reached`) will be used and the user has to add a query param `replaceOlder=true` to confirm replacement.

### 4.3. Services & processing
The statistics collection has to be made on pin/unpin activities with the author identity id and date of pin.

#### 4.4.1. Macro data model
The following fields will be added in existing Activities Table :
image

- `PINNED`: a **Boolean** attribute to indicate whether the activity is pinned or not
- `PIN_DATE`: a **Timestamp** of pin operation (should be null when unpinned)
- `PIN_AUTHOR_ID`: a **Long** attribute for the AUTHOR identity identifier (should be null when unpinned)

#### 4.4.2. Cache management
The activity and space stream activity ids has to be purged when an activity gets pinned or unpinned.
#### 4.4.3. Persistence layer
The Query of retrieving the activities of a space (get activities by owner) has to be modified in order to retrieve the list of activities by the following orders :

1. `ORDER BY activity.pinned DESC`
2. `ORDER BY activity.pinDate DESC`
3. `ORDER BY activity.updateDate DESC`

This will ensure to retrieve the pinned activities in top of retrieved activities list inside space activity. (that will can be used as feature inside the main activity stream)

AzmiTouil:
This PR will :
* allow to space host/redactor to pin/unpin activities to/from the space stream
* Add a new filter "Pinned activities" which allows filtering of the activities pinned in the spaces of which I am member.


component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
```diff
@@ -855,6 +855,65 @@ public Response deleteLike(
...
+    if (StringUtils.isBlank(activityId)) {
+      return Response.status(Response.Status.BAD_REQUEST).build();
+    }
+    org.exoplatform.services.security.Identity authenticatedUserIdentity = ConversationState.getCurrent().getIdentity();
+    String authenticatedUser = authenticatedUserIdentity.getUserId();
+    Identity currentUser = identityManager.getOrCreateIdentity(OrganizationIdentityProvider.NAME, authenticatedUser);
+
+    ExoSocialActivity activity = activityManager.getActivity(activityId);
+
+    if (!activityManager.canPinActivity(activity, currentUser)) {
```
rdenarie, line 882: What if activity is null (non existing id) ?

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
```diff
@@ -855,6 +855,65 @@ public Response deleteLike(
...
```
rdenarie, line 915: What if activity is null (activity not founded) ?

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
rdenarie: It seems to generate a NPE.

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
rdenarie: It seems to generate a NPE.

rdenarie (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
rdenarie: Example on builders : 

> 2022-10-18 08:28:24,913 | WARN  | Uncaught REST Service Exception [o.e.s.rest.impl.DefaultExceptionMapper<http-nio-0.0.0.0-8080-exec-144>] 
> java.lang.NullPointerException: null
> 	at org.exoplatform.social.core.manager.ActivityManagerImpl.canPinActivity(ActivityManagerImpl.java:569)
> 	at org.exoplatform.social.rest.impl.activity.ActivityRestResourcesV1.pinActivity(ActivityRestResourcesV1.java:879)
> 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> 	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
> 	at org.exoplatform.services.rest.impl.method.DefaultMethodInvoker$1.run(DefaultMethodInvoker.java:192)
> 	at org.exoplatform.commons.utils.SecurityHelper.doPrivilegedExceptionAction(SecurityHelper.java:311)
> 	at org.exoplatform.services.rest.impl.method.DefaultMethodInvoker.invokeMethod(DefaultMethodInvoker.java:188)
> 	at org.exoplatform.services.rest.impl.method.DefaultMethodInvoker.invokeMethod(DefaultMethodInvoker.java:181)
> 	at org.exoplatform.services.rest.impl.RequestDispatcher.invokeSubResourceMethod(RequestDispatcher.java:657)
> 	at org.exoplatform.services.rest.impl.RequestDispatcher.dispatch(RequestDispatcher.java:602)
> 	at org.exoplatform.services.rest.impl.RequestDispatcher.dispatch(RequestDispatcher.java:507)
> 	at org.exoplatform.services.rest.impl.RequestHandlerImpl.handleRequest(RequestHandlerImpl.java:143)
> 	at org.exoplatform.services.rest.servlet.RestServlet.onService(RestServlet.java:100)
> 	at org.exoplatform.container.web.AbstractHttpServlet.service(AbstractHttpServlet.java:133)
> 	at javax.servlet.http.HttpServlet.service(HttpServlet.java:764)

I used pins endpoint, with activityId = 100000 : 
`romain@romain-XPS-ubuntu:~$ curl -X POST -b cookies https://builders.meeds.io/portal/rest/v1/social/activities/100000/pins
java.lang.NullPointerException`

rdenarie: I have a security question here : 
When I call this endpoint with an activity id : 
- if it answer 404, i know that the activity id do not exists
- if it answer 403, i know that activity id exists and that I'm not authorized to pin it.
So we disclose information about the existence of activity id.

In function getActivityById, we answer 404 in case activity not exists, or in case activity is not viewable by me.

We have two options : 
- if activity id not exists and if i'm not authorized to pin it, answer 404
- if activity id not exists, and if i'm not authorized to view it, answer 404, if i'm authorized to see it, but not to pin it, answer 403.
Second case is perhaps more complex to do.

Same remark for unpin function.

WDYT @boubaker ?

rdenarie (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
boubaker: > if activity id not exists, and if i'm not authorized to view it, answer 404, if i'm authorized to see it, but not to pin it, answer 403.

I agree

boubaker: > Second case is perhaps more complex to do.

Not too much in fact, it will be just same condition as getActivityById

```java
if (activity == null || !activityManager.isActivityViewable(activity, authenticatedUserIdentity)) {
  throw new WebApplicationException(Response.Status.NOT_FOUND);
}

```

rdenarie: :+1:

AzmiTouil: Done!

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
AzmiTouil: Done!

rdenarie (dismissed):
LGTM

boubaker (approved):

PR #1911, opened 2022-10-24
feat: Resize user avatars depends on the needed display size - EXO-60290 - Meeds-io/MIPs#25

hakermi:
Prior to this change, avatars are displayed and fetched with their original size, which affects the perf and leading to a critical loading time in the stream page because of the number of the enormous loaded images, This PR use the ResizeImageService and applying a resize step on avatars on server side before rendering it on the page which enhance the perf on the stream page and the images load time.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/manager/IdentityManagerImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/manager/IdentityManagerImpl.java
@@ -22,6 +22,7 @@ import java.util.*;
 
 import org.apache.commons.lang3.StringUtils;
 
+import org.exoplatform.commons.file.model.FileItem;
 import org.exoplatform.commons.utils.ListAccess;
 import org.exoplatform.container.PortalContainer;
 import org.exoplatform.container.xml.InitParams;
@@ -210,13 +211,24 @@ public class IdentityManagerImpl implements IdentityManager {
   /**
    * {@inheritDoc}
    */
+  @Override
+  public FileItem getAvatarFile(Identity identity) {
+    if (identity == null) {
+      return null;
+    }
+    return identityStorage.getAvatarFile(identity);
+  }
+
+  /**
+   * {@inheritDoc}
+   */
   public InputStream getBannerInputStream(Identity identity) throws IOException {
     if (identity == null) {
       return null;
     }
     return identityStorage.getBannerInputStreamById(identity);
   }
-  
+
   /**
    * {@inheritDoc}
    */
--- a/component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
@@ -1,0 +1,110 @@
+package org.exoplatform.social.core.metadata.thumbnail;
+
+import org.apache.commons.io.IOUtils;
+import org.exoplatform.commons.file.model.FileInfo;
+import org.exoplatform.commons.file.model.FileItem;
+import org.exoplatform.commons.file.services.FileService;
+import org.exoplatform.commons.file.services.FileStorageException;
+import org.exoplatform.services.log.ExoLogger;
+import org.exoplatform.services.log.Log;
+import org.exoplatform.services.thumbnail.ImageResizeService;
+import org.exoplatform.social.core.identity.model.Identity;
+import org.exoplatform.social.metadata.MetadataService;
+import org.exoplatform.social.metadata.model.MetadataItem;
+import org.exoplatform.social.metadata.model.MetadataKey;
+import org.exoplatform.social.metadata.model.MetadataType;
+import org.exoplatform.social.metadata.thumbnail.ImageThumbnailService;
+import org.exoplatform.social.metadata.thumbnail.model.ThumbnailObject;
+
+import java.io.ByteArrayInputStream;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class ImageThumbnailServiceImpl implements ImageThumbnailService {
+
+  private static final Log          LOG                       = ExoLogger.getExoLogger(ImageThumbnailServiceImpl.class);
+
+  private static final MetadataType THUMBNAIL_METADATA_TYPE   = new MetadataType(5, "thumbnail");
+
+  private static final String       SOCIAL_NAME_SPACE         = "social";
+
+  private static final String       THUMBNAIL_OBJECT_TYPE     = "file";
+
+  private static final String       THUMBNAIL_WIDTH_PROPERTY  = "width";
+
+  private static final String       THUMBNAIL_HEIGHT_PROPERTY = "height";
+
+  private final MetadataService     metadataService;
+
+  private final FileService         fileService;
+
+  private final ImageResizeService  imageResizeService;
+
+  public ImageThumbnailServiceImpl(MetadataService metadataService,
+                                   FileService fileService,
+                                   ImageResizeService imageResizeService) {
+    this.metadataService = metadataService;
+    this.fileService = fileService;
+    this.imageResizeService = imageResizeService;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public FileItem getOrCreateThumbnail(FileItem file, Identity identity, int width, int height) throws Exception {
+    if (file == null) {
+      throw new IllegalArgumentException("file argument is mandatory");
+    }
+    if (width == 0 && height == 0) {
+      return file;
+    }
+    FileInfo fileInfo = file.getFileInfo();
+    ThumbnailObject thumbnailObject = new ThumbnailObject(THUMBNAIL_OBJECT_TYPE, Long.toString(fileInfo.getId()));
+    List<MetadataItem> metadataItemList = metadataService.getMetadataItemsByMetadataTypeAndObject(THUMBNAIL_METADATA_TYPE.getName(),
+                                                                                                thumbnailObject);
+    List<MetadataItem> items = metadataItemList.stream()
+                                               .filter(metadataItem -> metadataItem.getProperties() != null
+                                                   && metadataItem.getProperties()
+                                                                  .get(THUMBNAIL_WIDTH_PROPERTY)
+                                                                  .equals(String.valueOf(width))
+                                                   && metadataItem.getProperties()
+                                                                  .get(THUMBNAIL_HEIGHT_PROPERTY)
+                                                                  .equals(String.valueOf(height)))
+                                               .collect(Collectors.toList());
+    if (!items.isEmpty()) {
+      long fileId = Long.parseLong(items.get(0).getParentObjectId());
+      try {
+        return fileService.getFile(fileId);
+      } catch (FileStorageException e) {
+        LOG.error("Error while getting thumbnail, original Image will be returned", e);
+        return file;
+      }
+    } else {
+      byte[] imageContent = imageResizeService.scaleImage(IOUtils.toByteArray(file.getAsStream()), width, height, false, false);
+      FileItem thumbnail = new FileItem(null,
+                                        fileInfo.getName(),
+                                        fileInfo.getMimetype(),
+                                        SOCIAL_NAME_SPACE,
+                                        imageContent.length,
+                                        new Date(),
+                                        fileInfo.getUpdater(),
+                                        false,
+                                        new ByteArrayInputStream(imageContent));
+      FileItem thumbnailFileItem = fileService.writeFile(thumbnail);
+      FileInfo thumbnailFileInfo = thumbnailFileItem.getFileInfo();
+      ThumbnailObject thumbnailMetadataObject = new ThumbnailObject(THUMBNAIL_OBJECT_TYPE,
+                                                                    Long.toString(fileInfo.getId()),
+                                                                    Long.toString(thumbnailFileInfo.getId()));
+      MetadataKey metadataKey = new MetadataKey(THUMBNAIL_METADATA_TYPE.getName(), THUMBNAIL_METADATA_TYPE.getName(), 0);
+      Map<String, String> properties = new HashMap<>();
+      properties.put(THUMBNAIL_WIDTH_PROPERTY, String.valueOf(width));
+      properties.put(THUMBNAIL_HEIGHT_PROPERTY, String.valueOf(height));
+      metadataService.createMetadataItem(thumbnailMetadataObject, metadataKey, properties, Long.parseLong(identity.getId()));
+      return thumbnailFileItem;
+    }
+  }
+}
--- a/component/service/pom.xml
+++ b/component/service/pom.xml
@@ -207,6 +207,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.imgscalr</groupId>
+      <artifactId>imgscalr-lib</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.exoplatform.commons</groupId>
       <artifactId>commons-testing</artifactId>
       <scope>test</scope>
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
@@ -45,6 +45,7 @@ import org.exoplatform.portal.mop.user.UserNode;
 import org.exoplatform.services.log.ExoLogger;
 import org.exoplatform.services.log.Log;
 import org.exoplatform.services.security.ConversationState;
+import org.exoplatform.social.common.Utils;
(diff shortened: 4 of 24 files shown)


component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -533,14 +540,26 @@ public Response getUserAvatarById(@Context UriInfo uriInfo,
...
+            */
             builder = Response.ok(stream, "image/png");
+            String[] dimension = size.split("x");
+            try {
+              avatarContent = imageResizeService.scaleImage(IOUtils.toByteArray(stream),
+                                                            Integer.parseInt(dimension[0]),
+                                                            Integer.parseInt(dimension[1]),
+                                                            false,
+                                                            false);
+              builder = Response.ok(avatarContent, "image/webp");
```
rdenarie: Image is not a webp format. Please keep the correct mimetype

component/service/pom.xml
```diff
@@ -201,6 +201,12 @@
       <artifactId>javax.servlet-api</artifactId>
       <scope>provided</scope>
     </dependency>
+    <!-- Test -->
+    <dependency>
```
rdenarie: Why this lib is needed for test ?

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -533,14 +540,26 @@ public Response getUserAvatarById(@Context UriInfo uriInfo,
...
-             * set the avatar content-type to "image/png". So, its data would be
-             * recognized as "image" by the user-agent
-             */
+            * As recommended in the the RFC1341
+            * (https://www.w3.org/Protocols/rfc1341/4_Content-Type.html), we
+            * set the avatar content-type to "image/png". So, its data would be
+            * recognized as "image" by the user-agent
+            */
             builder = Response.ok(stream, "image/png");
+            String[] dimension = size.split("x");
```
rdenarie: what if size is not splittable ? 
Inspire you of function thumbnailService.parseImageSize

rdenarie (changes requested):

rdenarie (changes requested):
You must change also for space avatars

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
hakermi: It will be always splittable, the URls will specified by developers and not regular users. it should be in a specific format.

component/service/pom.xml
hakermi: for integration social tests

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
hakermi: why this is a problem for you?

hakermi:
> You must change also for space avatars

yes, will do it

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
rdenarie: What if a dishonest user call directly the avatar rest service with a non parsable string ?

component/service/pom.xml
rdenarie: Why the dependency on commons do not pull this dependency ?

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
rdenarie: the webp format is not just a mime type. It is also a format. 
For make a parallel, you will never provide a xml file, with a word mimetype ? the browser, the software to use it ... will not be able to read correctly. 
As the image is a png or a jpeg, we must provide the correct mime type.

component/service/pom.xml
hakermi: the dependency added in commons is in provided scope, this is needed for useRestResourcesTest when testing resize avatars it will need this dependency in test scope or we will have a MethodDefNotFoundException when it calls scaleImage in testes

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
hakermi: done

hakermi:
> You must change also for space avatars

added

component/core/src/main/java/org/exoplatform/social/core/manager/IdentityManagerImpl.java
```diff
@@ -216,7 +228,7 @@ public InputStream getBannerInputStream(Identity identity) throws IOException {
     }
     return identityStorage.getBannerInputStreamById(identity);
   }
-  
+
```
rdenarie, line 231: Unecessary change

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+   * {@inheritDoc}
+   */
+  @Override
+  public FileItem getOrCreateThumbnail(FileItem file, Identity identity, int width, int height) throws Exception {
+    if (file == null) {
+      throw new IllegalArgumentException("file argument is mandatory");
+    }
+    if (identity == null) {
+      throw new IllegalArgumentException("User identity argument is mandatory");
+    }
```
rdenarie: Why do we need identity ?
We only need the original file, the wanted width and heigh. 
The identity should not be stored in the metadata

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    if (width == 0 && height == 0) {
+      return file;
+    }
```
rdenarie, line 64: Can there is a case where we provide only width or height ?

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+                                                   && metadataItem.getProperties()
+                                                                  .get(THUMBNAIL_HEIGHT_PROPERTY)
+                                                                  .equals(String.valueOf(height)))
+                                               .collect(Collectors.toList());
+    if (!items.isEmpty()) {
+      long fileId = Long.parseLong(items.get(0).getParentObjectId());
+      try {
+        return fileService.getFile(fileId);
+      } catch (FileStorageException e) {
+        return null;
```
rdenarie: Should we return oringial file and log an error ?

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+      }
+    } else {
+      byte[] imageContent = imageResizeService.scaleImage(IOUtils.toByteArray(file.getAsStream()), width, height, false, false);
+      FileItem thumbnail = new FileItem(null,
+                                        fileInfo.getName(),
+                                        fileInfo.getMimetype(),
+                                        SOCIAL_NAME_SPACE,
+                                        imageContent.length,
+                                        new Date(),
+                                        identity.getRemoteId(),
```
rdenarie: Can we say that the updater for the thumbnail is the same as the updater of the original file ?

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+      FileItem thumbnailFileItem = fileService.writeFile(thumbnail);
+      FileInfo thumbnailFileInfo = thumbnailFileItem.getFileInfo();
+      ThumbnailObject thumbnailMetadataObject = new ThumbnailObject(THUMBNAIL_OBJECT_TYPE,
+                                                                    Long.toString(fileInfo.getId()),
+                                                                    Long.toString(thumbnailFileInfo.getId()));
+      MetadataKey metadataKey = new MetadataKey(THUMBNAIL_METADATA_TYPE.getName(), THUMBNAIL_METADATA_TYPE.getName(), 0);
+      Map<String, String> properties = new HashMap<>();
+      properties.put(THUMBNAIL_WIDTH_PROPERTY, String.valueOf(width));
+      properties.put(THUMBNAIL_HEIGHT_PROPERTY, String.valueOf(height));
+      metadataService.createMetadataItem(thumbnailMetadataObject, metadataKey, properties, Long.parseLong(identity.getId()));
```
rdenarie, line 106:
```suggestion
      metadataService.createMetadataItem(thumbnailMetadataObject, metadataKey, properties, 0);
```

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
```diff
@@ -502,6 +509,19 @@ public Response getSpaceAvatarById(@Context UriInfo uriInfo,
         InputStream stream = identityManager.getAvatarInputStream(identity);
         if (stream != null) {
           builder = Response.ok(stream, "image/png");
+          int[] dimension = Utils.parseDimension(size);
+          try {
+            byte[] avatarContent = imageThumbnailService.getOrCreateThumbnail(identityManager.getAvatarFile(identity),
+                            identity,
+                            dimension[0],
+                            dimension[1])
```
rdenarie, line 517: by using Utils.parseDimension, dimension[0] and dimension[1] can be empty

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
```diff
@@ -502,6 +509,19 @@ public Response getSpaceAvatarById(@Context UriInfo uriInfo,
...
+                    .getAsByte();
+            if (avatarContent != null) {
+              builder = Response.ok(avatarContent, "image/png");
+            }
+          } catch (Exception e) {
+            LOG.warn("Error while resizing avatar, original Image will be returned", e);
```
rdenarie:
```suggestion
            LOG.error("Error while resizing avatar, original Image will be returned", e);
```

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -533,14 +540,28 @@ public Response getUserAvatarById(@Context UriInfo uriInfo,
...
+            * set the avatar content-type to "image/png". So, its data would be
+            * recognized as "image" by the user-agent
+            */
             builder = Response.ok(stream, "image/png");
+            int[] dimension = Utils.parseDimension(size);
+            try {
+              avatarContent = imageThumbnailService.getOrCreateThumbnail(identityManager.getAvatarFile(identity),
+                                                                         identity,
+                                                                         dimension[0],
+                                                                         dimension[1])
```
rdenarie, line 557: same remark as for spaces

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -533,14 +540,28 @@ public Response getUserAvatarById(@Context UriInfo uriInfo,
...
+                                                   .getAsByte();
+              if (avatarContent != null) {
+                builder = Response.ok(avatarContent, "image/png");
+              }
+            } catch (Exception e) {
+              LOG.warn("Error while resizing avatar, original Image will be returned", e);
```
```suggestion
              LOG.error("Error while resizing avatar, original Image will be returned", e);
```

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -1752,4 +1773,5 @@ private StringBuilder getUrl(HttpServletRequest request) {
     }
     return url;
   }
+
```
rdenarie: unecessary change

rdenarie (changes requested):
Did you resize logo in Whosonline application ? 
Should we modify ExoUserAvatar component ? ExoSpaceAvatar component ?

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
hakermi: OK

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
hakermi: OK

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRestResourcesV1.java
hakermi: Yes it can be 0 height and 0 width and this case is treated in the imageResizeService and the ImageThumbnailService

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
hakermi: Yes, there is no a problem in  that, it's treated in the ImageResizeService. but in this case of (0,0) the ImageREsizeService will return the original image, so no need to create a metadata for it too.

component/core/src/main/java/org/exoplatform/social/core/manager/IdentityManagerImpl.java
hakermi: > Did you resize logo in Whosonline application ? Should we modify ExoUserAvatar component ? ExoSpaceAvatar component ?

They will get the default 45x45 thumbnail

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
hakermi: yes

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
hakermi: will throw an exception:  https://github.com/Meeds-io/social/blob/develop/component/core/src/main/java/org/exoplatform/social/core/metadata/MetadataServiceImpl.java#L183

hakermi: i need the used id here, so i add the identity as a parameter in the getOrCreateThumbnail? or i add only user identity id?

component/core/src/main/java/org/exoplatform/social/core/manager/IdentityManagerImpl.java
```diff
@@ -216,7 +228,6 @@ public InputStream getBannerInputStream(Identity identity) throws IOException {
     }
     return identityStorage.getBannerInputStreamById(identity);
   }
-
```
rdenarie, line 219: Unecessary change

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
rdenarie: As the thumbnail is related to the original file, should we use the originalFile updater as thumbnail creator ?

rdenarie (changes requested):

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
rdenarie: ok to keep identity as it is the avatar owner identity

component/core/src/main/java/org/exoplatform/social/core/manager/IdentityManagerImpl.java
hakermi: i undo the change, it's just bad detected in github, as you can see it's a simple line between functions
Screenshot from 2022-11-16 14-35-10

rdenarie (changes requested):
Could you add the mip issue link in the commit message.

hakermi:
> Could you add the mip issue link in the commit message.

done

rdenarie (approved):

PR #1916, opened 2022-10-25
fix: Notes_Version history_bug: History version drawer is empty - EXO-60449

mkrout:
Prior to this fix the version history drawer for notes is empty


hakermi (commented):
This is will cause an issue in documents app, it would be better to add an id for note version model

PR #1939, opened 2022-11-01

sofyenne:
Prior to this change, we can not edit an activity with body containing ` % ` character and link with preview, the problem is that the activity body to be edited is a malformed URI which can not be decoded. After this change, we will encode the activity body to be edited.


rdenarie (changes requested):
Can you modify your pr title ?
You wrote meedsq-io/meeds #374 instead of meedsq-io/meeds#374 so the link is not auto generated. Please update so that the link to the issue is correctly generated.

sofyenne:
> Can you modify your pr title ? You wrote meedsq-io/meeds #374 instead of meedsq-io/meeds#374 so the link is not auto generated. Please update so that the link to the issue is correctly generated.

done!

rdenarie (approved):

PR #1942, opened 2022-11-07

sofyenne:
Prior to this change, when we add a new post with text containing "%" character and a link with preview, it is not possible to edit it later, we have tow problem :
- The activity body to be edited contains a mal formed URI so we can't decode it .
- The generated oembed tag is encoded twice.

After this change, we will decode the oembed tag then encode the whole activity body to be edited in order to ensure the decode later.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
@@ -162,7 +162,7 @@ export default {
         this.originalBody = params.activityBody;
         this.activityId = params.activityId;
         this.spaceId = params.spaceId;
-        this.templateParams = params.activityParams || {};
+        this.templateParams = params.activityParams || params.templateParams || {};
         this.files = params.files || [];
         this.activityType = params.activityType;
       } else {
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/extensions.js
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/extensions.js
@@ -81,7 +81,13 @@ const defaultActivityOptions = {
            || '');
   },
   getBodyToEdit: activity => {
-    const templateParams = encodeURIComponent(activity.templateParams);
+    let templateParams = activity.templateParams; 
+    if (templateParams.default_title && templateParams.default_title.includes('<oembed>') && templateParams.link){
+      const url = window.encodeURIComponent(templateParams.link);
+      templateParams.default_title = templateParams.default_title.replace(`<oembed>${url}</oembed>`, `<oembed>${templateParams.link}</oembed>`);
+      activity.title = activity.title.replace(`<oembed>${url}</oembed>`, `<oembed>${templateParams.link}</oembed>`);
+    }
+    templateParams = encodeURIComponent(activity.templateParams);
     return Vue.prototype.$utils.trim(window.decodeURIComponent(templateParams
       && templateParams.default_title
       && templateParams.default_title


component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSActivityStorageImpl.java
```diff
@@ -1397,9 +1397,13 @@ public void updateActivity(ExoSocialActivity existingActivity) {
...
-        Map<String, String> updatedActivityTemplateParams = updatedActivity.getTemplateParams();
-        updatedActivityTemplateParams.putAll(existingActivity.getTemplateParams());
-        updatedActivity.setTemplateParams(updatedActivityTemplateParams);
+        if (existingActivity.getTemplateParams() != null && updatedActivity.getTemplateParams().containsKey("default_title") && !updatedActivity.getTemplateParams().get("default_title").equals(existingActivity.getTitle())){
+          updatedActivity.setTemplateParams(existingActivity.getTemplateParams());
+        }else {
+          Map<String, String> updatedActivityTemplateParams = updatedActivity.getTemplateParams();
+          updatedActivityTemplateParams.putAll(existingActivity.getTemplateParams());
+          updatedActivity.setTemplateParams(updatedActivityTemplateParams);
+        }
```
azayati: not needed IMO

webapp/portlet/src/main/webapp/vue-apps/activity-stream/extensions.js
```diff
@@ -81,13 +81,17 @@ const defaultActivityOptions = {
...
+    const templateParams = activity.templateParams;
+    return Vue.prototype.$utils.trim(templateParams
       && templateParams.default_title
       && templateParams.default_title
-      || (activity.title && activity.title.replaceAll('%', '%25'))
-      || (activity.body && activity.title.replaceAll('%', '%25'))
-      || ''));
+      || (activity.title && activity.title)
+      || (activity.body && activity.title)
+      || '');
```
azayati: it is another problem that should be treated separately in another US

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
```diff
@@ -162,7 +162,7 @@ export default {
         this.originalBody = params.activityBody;
         this.activityId = params.activityId;
         this.spaceId = params.spaceId;
-        this.templateParams = params.activityParams || {};
+        this.templateParams = params.activityParams || params.templateParams || {};
```
azayati, line 165: we should see why it is needed with this [modification](https://github.com/Meeds-io/social/pull/1939/files#diff-e4d30f4c5a0b8396f260601b333ccf48a8c144a751a2ec440e412f6bf2a9f1eeR84)

azayati (dismissed):
see review

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSActivityStorageImpl.java
sofyenne: PR updated !

webapp/portlet/src/main/webapp/vue-apps/activity-stream/extensions.js
sofyenne: PR updated

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
sofyenne: en mode edit we need this modification because if we edited activity before the loading of the preview link the activity can't updated so we need to load the existing template params 
we can skip this modification but we should save the edited text after loading the preview link

azayati (approved):
LGTM

PR #1954, opened 2022-11-10

sofyenne:
Prior to this change, when we add a new post with text containing "%" character and a link with preview, it is not possible to edit it later, we have tow problem :

- The activity body to be edited contains a mal formed URI so we can't decode it .
- The generated oembed tag is encoded twice.

After this change, we will decode the oembed tag then encode the whole activity body to be edited in order to ensure the decode later.

(cherry picked from commit https://github.com/Meeds-io/social/commit/8db8e38d56576ad725c316be15f1a1121f7a49b7)


azayati (approved):
LGTM

PR #1960, opened 2022-11-11
feat: Stream activities when mobile MIP#21

### LINKED ISSUE #21: Stream activities - Mobile view of activities
## Rationale

We have identified improvements to make the stream activities UI up to date regarding the mobile usage we can provide.
Indeed, when using mobile, the activities are displayed the same way it is when using desktop. The problem is that when long name for user or space, the UI is KO: we cannot see the whole user name and more often, the audience (which is the space name cannot be seen)

So we suggest to make Activities UI more responsive to our use case.

## 1. Functional requirements
- Activities UI when using mobile
When using the mobile view and browsing the general stream, I should have a topbar activity UI that way (similar to other social networks)
image

When browsing the space stream, no need to remind the audience (space) since we are browsing the space stream
image

## 2. Non Functional requirements - Feature Flags

 >**Note**:  No feature flag needed

ArfaouiMohamedAmine:
This change is going to change to space and user avatar display when using mobile through:

-space avatar behind the user avatar
-space name above the user name
-When browsing a space activity stream The activities should be adjusted so we only see the username. No need to have the space name reminded since we are browsing the space itself
-the alignement of the name and the timestamp in the activity stream are adjusted .
-timestamp is displayed next to the username

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityMobileHead.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityMobileHead.vue
@@ -1,3 +1,19 @@
+<!--
+  This file is part of the Meeds project (https://meeds.io/).
+  Copyright (C) 2022 Meeds Association
+  contact@meeds.io
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
+  version 3 of the License, or (at your option) any later version.
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  Lesser General Public License for more details.
+  You should have received a copy of the GNU Lesser General Public License
+  along with this program; if not, write to the Free Software Foundation,
+  Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+-->
 <template>
   <v-list-item v-if="displayUserAvatar" class="text-truncate">
     <exo-user-avatar
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityMobileHead.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityMobileHead.vue
@@ -45,19 +45,20 @@
           link-style
           username-class />
       </v-list-item-title>
-      <v-list-item-subtitle class="d-flex flex-row align-center flex-nowrap">
-        <exo-user-avatar
-          :identity="posterIdentity"
-          extra-class="text-truncate ms-2 me-1"
-          fullname
-          link-style
-          small-font-size
-          username-class />
-        <activity-head-time
-          :activity="activity"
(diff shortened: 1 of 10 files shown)


webapp/portlet/src/main/webapp/vue-apps/common/components/ExoSpaceAvatar.vue
```diff
@@ -9,16 +9,17 @@
       class="flex-nowrap flex-shrink-0 d-flex spaceAvatar not-clickable-link">
       <v-avatar
         :size="size"
+        :class="pullLeft"
```
rdenarie, line 12: Does this will change behaviour from ExoSpaceAvatar when used with mobile in another component than ActivityHead ?

webapp/portlet/src/main/webapp/vue-apps/common/components/ExoUserAvatar.vue
```diff
@@ -12,7 +12,8 @@
       :class="avatarClass">
       <v-avatar
         :size="size"
-        class="ma-0 pull-left">
+        :class="pullLeft"
```
rdenarie, line 15: Does this will change behaviour from ExoUserAvatar when used with mobile in another component than ActivityHead ?

ArfaouiMohamedAmine: pull left won't be applied when 'this.isMobile' 
`    pullLeft() {
      return this.isMobile && ' ' || 'pull-left';
    },
`

rdenarie: Before your change, pull-left is always applied. 
After your change, it will not be applied when in mobile.

Does it will change something when the component ExoUserAvatar is used in another component then ActivityHead ?

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityMobileHead.vue
```diff
@@ -0,0 +1,94 @@
+<template>
```
rdenarie, line 17: Please add the header.

rdenarie (changes requested):

webapp/portlet/src/main/webapp/vue-apps/common/components/ExoUserAvatar.vue
ArfaouiMohamedAmine: thank you for the clarification,
no the exoUserAvatar remained with same behaviour, it wont be affected with change .

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityMobileHead.vue
ArfaouiMohamedAmine: header added !

rdenarie:
Can you rebase the branch on develop ?

rdenarie (changes requested):
There is problem with pom version which are related to meed. It should not.

@boubaker can you check the new header on the vue file ? The sonar rule do not pass, but I dont find why.

rdenarie:
Can you rebase the PR ?

rdenarie (dismissed):

rdenarie (approved):

rdenarie (approved):

PR #1973, opened 2022-11-17
feat: Delete related metadata when updating user avatar - EXO-60290 - Meeds-io/MIPs#25

hakermi:
Prior to this change, when updating user avatar, the thumbnail remains the same without getting updated too, which cause the display of the old avatar always.
This PR should make sure to drop related metadata when the avatar was updated

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/listeners/FileActionListener.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/listeners/FileActionListener.java
@@ -1,0 +1,49 @@
+/*
+ * This file is part of the Meeds project (https://meeds.io/).
+ * Copyright (C) 2022 Meeds Association contact@meeds.io
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package org.exoplatform.social.core.listeners;
+
+import org.exoplatform.commons.file.model.FileInfo;
+import org.exoplatform.services.listener.Event;
+import org.exoplatform.services.listener.Listener;
+import org.exoplatform.social.metadata.MetadataService;
+import org.exoplatform.social.metadata.thumbnail.model.ThumbnailObject;
+
+public class FileActionListener extends Listener<FileInfo, Object> {
+
+  private static final String   FILE_UPDATED_EVENT      = "file.updated";
+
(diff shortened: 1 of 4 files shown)


component/core/src/main/java/org/exoplatform/social/core/listeners/FileActionListener.java
```diff
@@ -0,0 +1,34 @@
+package org.exoplatform.social.core.listeners;
```
rdenarie, line 16: Licence header is missing

component/core/src/main/java/org/exoplatform/social/core/listeners/FileActionListener.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public FileActionListener(MetadataService metadataService) {
+    this.metadataService = metadataService;
+  }
+
+  @Override
+  public void onEvent(Event<FileInfo, Object> event) throws Exception {
+    String eventName = event.getEventName();
+    FileInfo fileInfo = event.getSource();
+    if (eventName.equals(FILE_UPDATED_EVENT) && fileInfo.getNameSpace().equals(SOCIAL_NAME_SPACE)) {
```
rdenarie: add delete event remove second test

rdenarie (changes requested):

component/core/src/main/java/org/exoplatform/social/core/listeners/FileActionListener.java
hakermi: done

rdenarie (approved):

component/core/src/test/java/org/exoplatform/social/core/listeners/FileActionListenerTest.java
```diff
@@ -0,0 +1,92 @@
...
+    metadataService = getContainer().getComponentInstanceOfType(MetadataService.class);
+    identityManager = getContainer().getComponentInstanceOfType(IdentityManager.class);
+    userIdentity = identityManager.getOrCreateIdentity(OrganizationIdentityProvider.NAME, "john");
+    fileService = getContainer().getComponentInstanceOfType(FileService.class);
+
+    fileActionListener = new FileActionListener(metadataService);
+  }
+
+  public void testOnEvent() throws Exception {
+    ThumbnailObject thumbnailObject = new ThumbnailObject("file", "1");
```
rdenarie, line 49: Your test here is not a unit test, it is integration test.
You are verifying that, when creating the file by file service, the event is correctly launched. So your are testing the ListenerService.

If you want to unit test the FileActionServiceListener, you should do something like that :
FileActionListener f = new fileActionListener();

Event<FileInfo, Object> event = new Event ....
event.fileInfo.setEventName=FILE_UPDATED
f.onEvent(event);
check that deleteMetadataItemsByMetadataTypeAndObject was called

same with eventName = FILE_DELETED

rdenarie (changes requested):

component/core/src/test/java/org/exoplatform/social/core/listeners/FileActionListenerTest.java
```diff
@@ -0,0 +1,66 @@
+package org.exoplatform.social.core.listeners;
```
rdenarie, line 16: Missing header

rdenarie (changes requested):

component/core/src/test/java/org/exoplatform/social/core/listeners/FileActionListenerTest.java
hakermi: added

rdenarie (approved):

PR #1977, opened 2022-11-17
MEED-511: Delete activity draft when updating a message previoly posted

AzmiTouil:
Prior to this change, when updating an existing activity message from the activity composer drawer and the click to add a new activity message the last activity updated message remains in the text editor area. This change, allow to delete the message from the local storage after updating an existing activity message.


SaraBoutej (approved):

SaraBoutej (approved):
LGTM

PR #1981, opened 2022-11-17
fix: Fix edited activity link and its preview still displayed after removing the link - EXO-60760 - Meeds-io/meeds#417

IlhemEssaadi:
Prior to this change, when we modify an activity containing a link with its corresponding preview by deleting the link and the corresponding preview and adding a simple text, the link and the corresponding preview are still displayed. The problem is that the updated activity templateParams related to the link will be deleted, so when we put the template params of the updated activity to those of the old one, the deleted link template params of the updated activity will persist with the old activity ones. After this modification, we will ensure to persist the link template params of the updated activity in order to update the activity without link and corresponding preview.

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
@@ -764,7 +764,7 @@ public class EntityBuilder {
     while (entries.hasNext()) {
       Map.Entry<String, String> entry = entries.next();
       if (entry != null && (StringUtils.isBlank(entry.getValue()) || StringUtils.equals(entry.getValue(), "-"))) {
-        entries.remove();
+        entry.setValue("");
       }
     }
     activity.setTemplateParams(currentTemplateParams);
--- a/component/service/src/test/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesTest.java
+++ b/component/service/src/test/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesTest.java
@@ -590,6 +590,35 @@ public class ActivityRestResourcesTest extends AbstractResourceTest {
     assertEquals(4, result.getTemplateParams().size());
   }
 
+  public void testUpdateLinkActivityById() throws Exception {
+    startSessionAs("root");
+
+    ExoSocialActivity rootActivity = new ExoSocialActivityImpl();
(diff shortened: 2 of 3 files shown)


component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSActivityStorageImpl.java
```diff
@@ -1397,9 +1397,7 @@ public void updateActivity(ExoSocialActivity existingActivity) {
       if (existingActivity.getPermaLink() != null)
         updatedActivity.setPermaLink(existingActivity.getPermaLink());
       if (existingActivity.getTemplateParams() != null) {
-        Map<String, String> updatedActivityTemplateParams = updatedActivity.getTemplateParams();
-        updatedActivityTemplateParams.putAll(existingActivity.getTemplateParams());
-        updatedActivity.setTemplateParams(updatedActivityTemplateParams);
+        updatedActivity.setTemplateParams(existingActivity.getTemplateParams());
```
azayati: normally we should keep the default code, otherwise we will have regressions according to [this commit](https://github.com/Meeds-io/social/commit/adc9f7e190fe938b3714201b18a42b2ec56beed1)

azayati (changes requested):
see review

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSActivityStorageImpl.java
plamarque: normally we should have a test case and a comment that explains why we do such odd things...

azayati: we haven't implemented odd things, it is well explained in the commit why it is modified. The test case for update activity already exists but not for this specific case maybe.

plamarque: my point is exactly that. This regression should not have been possible if we had had a unit test that would protect this case. It should be added because this code is not self explanatory and we coud easily think it does not do much.

azayati (dismissed):
LGTM

ahamdi (approved):

azayati (approved):
LGTM

PR #1982, opened 2022-11-18
fix : Fix duplicated encoded link in activity - EXO-60750 - Meeds-io/meeds#398

sofyenne:
Prior to this change when we post an activity with a copied link from Brave browser without "www" generating a preview, the posted activity is displayed with a duplicated encoded link . The problem is caused by missing `www` in the copied link when we copy from brave browser.
After this change, we will update the installOmbed method in order to fix this problem.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/common/components/ExoActivityRichEditor.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/common/components/ExoActivityRichEditor.vue
@@ -342,7 +342,7 @@ export default {
           && this.templateParams.link
           && this.templateParams.default_title !== message) {
         this.templateParams.default_title = message;
-        const codedMessage = window.encodeURIComponent(message.substring(0,message.indexOf('<oembed>')));
+        const codedMessage = window.encodeURIComponent(message.substring(0, message.indexOf('<oembed>')));
         this.templateParams.comment = window.decodeURIComponent(codedMessage);
       }
     },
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/common/components/ExoActivityRichEditor.vue
```diff
@@ -342,8 +342,7 @@ export default {
           && this.templateParams.link
           && this.templateParams.default_title !== message) {
         this.templateParams.default_title = message;
-        const url = window.encodeURIComponent(this.templateParams.link);
-        const codedMessage = window.encodeURIComponent(message.replace(`<oembed>${url}</oembed>`, ''));
+        const codedMessage = window.encodeURIComponent(message.substring(0,message.indexOf('<oembed>')));
```
azayati:
```suggestion
        const codedMessage = window.encodeURIComponent(message.substring(0, message.indexOf('<oembed>')));
```

azayati (changes requested):
verify that no regressions on previous fixes are caused by this change

webapp/portlet/src/main/webapp/vue-apps/common/components/ExoActivityRichEditor.vue
sofyenne: done !

azayati (approved):
LGTM

PR #1984, opened 2022-11-21
fix: Fix duplicated encoded link in activity - EXO-60750 - Meeds-io/meeds#398 

sofyenne:
Prior to this change when we post an activity with a copied link from Brave browser without "www" generating a preview, the posted activity is displayed with a duplicated encoded link . The problem is caused by missing `www` in the copied link when we copy from brave browser. After this change, we will update the installOmbed method in order to fix this problem.


ahamdi (approved):

azayati (approved):
LGTM

PR #1986, opened 2022-11-23
feat: Resize avatars in space card and people card - EXO-60290 - Meeds-io/MIPs#25

hakermi:
Prior to this change used dimension for avatars in space card is 75x75 and for people card is 65x65 but we request a default 45x45 avatar. This PR adds the size flag for avatar rest endpoint to get the appropriate dimension

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/content/ActivityLink.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/content/ActivityLink.vue
@@ -60,7 +60,7 @@
         tile>
         <img
           v-if="thumbnail"
-          :src="`${thumbnail}&size=250x150`"
+          :src="thumbnail"
           :alt="title"
           class="object-fit-cover my-auto"
           loading="lazy"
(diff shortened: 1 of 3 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/content/ActivityLink.vue
```diff
@@ -60,7 +60,7 @@
         tile>
         <img
           v-if="thumbnail"
-          :src="`${thumbnail}&size=250x150`"
```
rdenarie, line 63: Why do you remove the size here ?

rdenarie (changes requested):

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/content/ActivityLink.vue
hakermi: Because it's set here now https://github.com/exoplatform/news/pull/675/files#diff-9823e5377655181c88db5d8bd4e7e988795970046ef7d49627a36ca782d4305cR31

rdenarie: As you moved that in news, does it impacts images added in activity (not a news activity) ?

hakermi: Added images in activity are attachments, and they are managed in ecms with thumbnail service. But this change was for news illustrations before

rdenarie (approved):

PR #1991, opened 2022-11-25
fix: Fix edited activity link and its preview still displayed after removing the link - EXO-60760 - Meeds-io/meeds#417

IlhemEssaadi:
Prior to this change, when we modify an activity containing a link with its corresponding preview by deleting the link and the corresponding preview and adding a simple text, the link and the corresponding preview are still displayed. The problem is that the updated activity templateParams related to the link will be deleted, so when we put the template params of the updated activity to those of the old one, the deleted link template params of the updated activity will persist with the old activity ones. After this modification, we will ensure to persist the link template params of the updated activity in order to update the activity without link and corresponding preview.


azayati (approved):
LGTM

PR #1998, opened 2022-11-28
feat: Resize avatars depends on the needed display size - EXO-60290 - Meeds-io/MIPs#25

hakermi:
Prior to this change, avatars are displayed and fetched with their original size, which affects the perf and leading to a critical loading time in the stream page because of the number of the enormous loaded images.
This PR:
1 - Use the ResizeImageService and applying a resize step on avatars (users and spaces) on server side before rendering it on the page which enhance the perf on the stream page and the images load time. The resized image in store by using Metadata Service
2 - Make sure to drop related metadata when the avatar was updated to recreate it when resize the new uploaded image.


component/core/src/main/java/org/exoplatform/social/core/listeners/FileActionListener.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package org.exoplatform.social.core.listeners;
+
+import org.exoplatform.commons.file.model.FileInfo;
+import org.exoplatform.services.listener.Event;
+import org.exoplatform.services.listener.Listener;
+import org.exoplatform.social.metadata.MetadataService;
+import org.exoplatform.social.metadata.thumbnail.model.ThumbnailObject;
+
+public class FileActionListener extends Listener<FileInfo, Object> {
```
boubaker: Could you please rename this class to a more specific name, knowing that it's about cleaning & updating thumbnails only. It would be `ThumbnailFileListener`

component/core/src/main/java/org/exoplatform/social/core/listeners/FileActionListener.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Override
+  public void onEvent(Event<FileInfo, Object> event) throws Exception {
+    String eventName = event.getEventName();
+    FileInfo fileInfo = event.getSource();
+    if (eventName.equals(FILE_UPDATED_EVENT) || eventName.equals(FILE_DELETED_EVENT)) {
+      ThumbnailObject thumbnailObject = new ThumbnailObject(THUMBNAIL_OBJECT_TYPE, Long.toString(fileInfo.getId()));
+      metadataService.deleteMetadataItemsByMetadataTypeAndObject(THUMBNAIL_METADATA_NAME, thumbnailObject);
+    }
```
boubaker: The main idea about using APIs is to not define the business rules outside it. In addition, the used storage for API implementation should be gathered by the Implementation only without exposing it outside. Thus here, you need to define a specific method inside `ImageThumbnailService` to delete thumbnails of an associated files and avoid using directly the MetadataService, which is the storage layer for thumbnails as defined in current implementation of  `ImageThumbnailService`. In other terms, this call should be :

```suggestion
  imageThumbnailService.deleteThumbnails(fileInfo.getId())
```

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -533,14 +540,28 @@ public Response getUserAvatarById(@Context UriInfo uriInfo,
...
             builder = Response.ok(stream, "image/png");
+            int[] dimension = Utils.parseDimension(size);
+            try {
+              avatarContent = imageThumbnailService.getOrCreateThumbnail(identityManager.getAvatarFile(identity),
+                                                                         identity,
+                                                                         dimension[0],
+                                                                         dimension[1])
+                                                   .getAsByte();
+              if (avatarContent != null) {
+                builder = Response.ok(avatarContent, "image/png");
```
boubaker: Why making two calls of `Response.ok(stream, "image/png")` ? IMO, it should be managed by an `if...else` to ensure to either return a Thumbnail or the original image, without the return instruction to make sure to benifit from cache management (HTTP 304 when not modified). Besides, the eTag value must change depending on sent sizes to make sure that the HTTP 304 works for each size on its own.

```suggestion
            builder = Response.ok(stream, "image/png");
            int[] dimension = Utils.parseDimension(size);
            try {
              avatarContent = imageThumbnailService.getOrCreateThumbnail(identityManager.getAvatarFile(identity),
                                                                         identity,
                                                                         dimension[0],
                                                                         dimension[1])
                                                   .getAsByte();
              if (avatarContent != null) {
                builder = Response.ok(avatarContent, "image/png");
```

component/core/src/main/java/org/exoplatform/social/core/listeners/FileActionListener.java
rdenarie: Done

component/core/src/main/java/org/exoplatform/social/core/listeners/FileActionListener.java
rdenarie: Done

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
rdenarie: Done

boubaker: We may avoid retrieving the avatar stream twice IMO, by testing on existence of `AvatarLastUpdated` field in both REST endpoints (`UserRestResourcesV1` & `SpaceRestResourcesV1`), by example:
```java
/* eTag Check - Start */
      builder = (eTag == null ? null : request.evaluatePreconditions(eTag));
      if (builder == null) {
/* eTag Check - Not matching Start */

// Retrieve one Stream content either from thumbnail, or original stream (when no size defined) or default avatar when the user/space doesn't have an avatar yet
if (lastUpdated == null) {// lastUpdated = identity.getAvatarLastUpdated() is already retrieved
  builder = getDefaultAvatarBuilder(); // Existing method From SpaceRestResourcesV1
} else {
  if (StringUtils.isNotBlank(size)) {
       int[] dimension = Utils.parseDimension(size);
      ...
     builder = Response.ok(avatarContent, "image/png");
  } else {
    InputStream stream = identityManager.getAvatarInputStream(identity);
     builder = Response.ok(stream, "image/png");
  }
}

/* eTag Check - Not matching End */

     }
/* eTag Check - End */

```

Besides, I think it would be better to not have a default size value in REST endpoint to be able to retrieve the original avatar image.

rdenarie: We decide to put a default size at 45x45 because it is the major size required everywhere we use avatar : 
Stream, activity, whoonline, task ...
So, by using a default size, we ensure to treat all place the avatar is used.

But you're right, there is no way to get the default avatar. 
So I propose to have a special size, "original", which return the original avatar if needed.
WDYT ?

About the other part "avoid retrieving the avatar stream twice IMO", I'm ok to change that when we agree on the original size

boubaker: Yep, totally agree, thanks !

hakermi: original avatar will be returned with `&size=0x0`

rdenarie: Thanks for the precision @hakermi 
Is it sufficient @boubaker ?

boubaker: Yep, for the default value retrieved by `size=0x0` instead of using `original`, but we may need to make this as default value in REST endpoint instead of `45x45` (no need if it's too much modifications) & explain in doc that using 0x0 will retrieve the default value + a Unit Test to prevent it from regressing in future evolution.

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+                                                                                                thumbnailObject);
+    List<MetadataItem> items = metadataItemList.stream()
+                                               .filter(metadataItem -> metadataItem.getProperties() != null
+                                                   && metadataItem.getProperties()
+                                                                  .get(THUMBNAIL_WIDTH_PROPERTY)
+                                                                  .equals(String.valueOf(width))
+                                                   && metadataItem.getProperties()
+                                                                  .get(THUMBNAIL_HEIGHT_PROPERTY)
+                                                                  .equals(String.valueOf(height)))
+                                               .collect(Collectors.toList());
```
boubaker:
```suggestion
                                               .toList();
```

boubaker:
The remark https://github.com/Meeds-io/social/pull/1998#discussion_r1050597360 is still not resolved yet

rdenarie:
I'm on it. I let you know when finished.
Thanks

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
rdenarie: Done : add information on how to get original size, add unit test to be sure to prevent regression
Reorganize code to be more efficient when reading input stream : do not read default image if not needed, do not read original image if not needed.

component/core/src/main/java/org/exoplatform/social/core/metadata/thumbnail/ImageThumbnailServiceImpl.java
rdenarie: done

rdenarie:
@boubaker modifications applied, you can review

boubaker (approved):
LGTM

PR #2018, opened 2022-12-05
fix: getting notifications while not member anymore in a space - EXO-61097

GouadriaHanen:
before this change, users continue to receive notifications of activities they have already commented on or liked even though they are no longer members.
after this change, notification receivers are filtered by whether they are members of the space or not.

### DIFF
--- a/component/notification/src/main/java/org/exoplatform/social/notification/Utils.java
+++ b/component/notification/src/main/java/org/exoplatform/social/notification/Utils.java
@@ -146,9 +146,9 @@ public class Utils {
    * Checks if a notification message is sent to an activity poster when a new comment is created.
    * @param receivers The list of users receiving the notification message.
    * @param activityPosterId Id of the activity poster.
-   * @param posteId Id of the user who has commented.
+   * @param posterId Id of the user who has commented.
    */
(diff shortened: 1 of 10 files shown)


ahamdi (commented):
@GouadriaHanen Could we add this check globally ? Otherwise we will have to look for all Notification plugins in all other modules to make them aware of this restriction.

component/notification/src/main/java/org/exoplatform/social/notification/Utils.java
```diff
@@ -148,15 +148,22 @@ public static void sendToStreamOwner(Set<String> receivers, String streamOwner,
    * @param activityPosterId Id of the activity poster.
    * @param posteId Id of the user who has commented.
    */
-  public static void sendToActivityPoster(Set<String> receivers, String activityPosterId, String posteId) {
+  public static void sendToActivityPoster(Set<String> receivers, String activityPosterId, String posteId , String spaceId) {
```
ahamdi: Could fix th etypo of the variable posteId ?
```suggestion
  public static void sendToActivityPoster(Set<String> receivers, String activityPosterId, String posterId , String spaceId) {
```

ahamdi (changes requested):
Could you add dedicated tests for this change ?

ahamdi (approved):

PR #2025, opened 2022-12-07
fix: Post Activity: Same activity was posted multiple times - EXO-61200 - Meeds-io/meeds#454

akhanfir:
Prior to this change, in the activity drawer when add content and click publish as much as possible, this activity is published multiple times. After this change, this activity posted only once.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
@@ -94,7 +94,8 @@ export default {
       activityBodyEdited: false,
       originalBody: '',
       messageEdited: false,
-      activityType: null
+      activityType: null,
+      loading: false
     };
   },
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
```diff
@@ -192,6 +192,9 @@ export default {
     postMessage() {
       // Using a ref to the editor component and the getMessage method is mandatory to
       // be sure to get the most up to date value of the message
+      if (this.loading) {
```
ahamdi: There is already the computed property postDisabled that should disable the button once this.loading is true. It should be sufficient to prevent sending the message multiple times
maybe we need just to change the position of the instruction ```this.loading = true``` to be in first lines of this function ?  WDYT ?

ahamdi (changes requested):

ahamdi (approved):
LGTM

PR #2041, opened 2022-12-14
feat: Merge Feature Publisher Role - Meeds-io/MIPs#35

### LINKED ISSUE #35: Publisher Role
## Rationale

The current permission management in a space is too restrictive. We have the need to promote the user as a new role to delegate, within a space, specific permissions  on the subject of information diffusion. 

## 1. Functional Requirements

- Be able for a host to promote users as "Publisher" in the space.
- This role gives permission to these users to achieve new actions.
- Be able to filter by the publisher in the member list in the space.
- We are using the same UI and UX in mobile as now => bottom drawer.

FT News - Publisher role

### Feature Flags

- **Functional behavior:** Hide (false) or display (true) the option "promote as publisher" in the 3 dots option in the member tab of a space
- **Property name:** exo.feature.publisherRolePromotion.enabled
- **Default status:** false
- **Target audience:** all users
- **Remove:** Keep the flag at the end of the iteration but switch the default value to true in eXo Package and keep it false in the Meeds package.

### Access 

Space members portlet should be updated in order to add:

- A new "profile-extension" action which allows to "Set user as publisher"
- A new "profile-extension" action to "remove publisher"

### Services & processing

- "ProfileEntity" should be updated by adding "isSpacePublisher" property.
- The existing "publisher" organization role will be used.
- The following rest POST service "/v1/groups/memberships" should be called in order to set/remove "publisher" membership on the space

azayati:


boubaker (dismissed):

rdenarie (approved):
Approved after Rebase

boubaker (approved):

PR #2042, opened 2022-12-14
fix: missing activity types values in analytics - EXO-60792

GouadriaHanen:
prior to this change when creating file or link activity its type is not set resulting not indexing the activity type
after this change, the type is set for created activity and are well indexed


GouadriaHanen:
to be treated in analytics

PR #2073, opened 2022-12-22
fix: Hide expand icon when using mobile - MEED-1376

MayTekayaa:
This change will hide the expand icon used in drawer when using mobile.


ArfaouiMohamedAmine (commented):
LGTM

boubaker (approved):

PR #2085, opened 2022-12-27
feat: Move OAuth Modules from GateIn to centralize them in Social - MEED-1308

boubaker:
Prior to this change, the oAuth Modules are defined in three artifacts and was managed by overrides in order to access users Social Identity. This change will allow to centralize oAuth Module in one single place to avoid overrides and complicated maintenance & evolutions. In addition, the Unit Tests has been configured to make it possible to call PortalContainer.getInstance() in addons and upper layes without exception when having the good Maven Modules added into it using <type>test-jar</type> for test artifacts.


exo-swf:

PR #2112, opened 2023-01-06
fix: favIcon do not appear in some space Popover - EXO-61430

GouadriaHanen:
prior to this change, favIcon does not appear in some space Popover since its component is not imported.
after this change, the component is imported and the favicon appears


ahamdi (commented):
How was it working before ? isn't it added using the extension registry ?

ahamdi (approved):

PR #2161, opened 2023-01-16
fix: Group Suggestor : double suggestion for same space - EXO-61750 (#2152)

GouadriaHanen:
prior to this change, space and group space are suggested together since the group of space is not excluded because of the wrong type name. after this change, group space fixed to excluded not needed groups


Jihed525 (approved):

ahamdi (approved):

webapp/portlet/src/main/webapp/vue-apps/common/js/SuggesterService.js
```diff
@@ -63,7 +63,7 @@ function searchGroups(filter, groupMember, groupType, items, allGroupsForAdmin,
   formData.append('groupMember', groupMember);
   formData.append('groupType', groupType);
   formData.append('allGroupsForAdmin', allGroupsForAdmin);
-  formData.append('excludeParentGroup', '/space');
+  formData.append('excludeParentGroup', '/spaces');
```
azayati, line 66: This is strange, with the old code and without this modification, it works well on tribe but not on ACC cc @ahamdi

PR #2185, opened 2023-01-18
feat: Impacts of Profile settings on user Profile  - EXO-61485 - Meed…

mkrout:
…s-io/MIPs#38


hakermi (approved):

component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
```diff
@@ -349,13 +349,24 @@ public static List <ProfilePropertySettingEntity> buildProperties(Profile profil
     List<ProfilePropertySetting> settings = profilePropertySettingsService.getPropertySettings();
     List<ProfilePropertySetting> subProperties = new ArrayList<>();
     List<Long> parents = new ArrayList<>();
+    boolean internal = false;
+    try {
+      OrganizationService organizationService = getOrganizationService();
```
ahamdi, line 354: Calls to OrganizationService are expensive, isn't it better to add the property directly in the social profile when loading it once if the property is not yet added to it ?

mkrout: yep, but IMO this should be done in all rest calls, not only this one, It should be treated in a separate task/pr to improve getting this property

PR #2201, opened 2023-01-20
feat: Change the space menu options in the spaces page - MEED-1491 - Meeds-io/meeds#549

MayTekayaa:
This change will fix the favorite title after the add and delete of the book.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/spaces-list/components/ExoSpaceCardFront.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/spaces-list/components/ExoSpaceCardFront.vue
@@ -367,13 +367,13 @@ export default {
         }, this.waitTimeUntilCloseMenu);
       }
     });
-    document.addEventListener('favorite-added', event => {
-      if (event && event.detail) {
+    document.addEventListener('space-favorite-added', event => {
+      if (event && event.detail === this.space.id) {
         this.space.isFavorite = 'true';
       }
     });
-    document.addEventListener('favorite-removed', event => {
-      if (event && event.detail) {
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/spaces-list/components/ExoSpaceCardFront.vue
```diff
@@ -367,6 +367,16 @@ export default {
         }, this.waitTimeUntilCloseMenu);
       }
     });
+    document.addEventListener('favorite-added', event => {
+      if (event && event.detail) {
+        this.space.isFavorite = 'true';
+      }
+    });
```
AzmiTouil:
```suggestion
    document.addEventListener('space-favorite-added', event => {
      if (event && event.detail === this.space.id) {
        this.space.isFavorite = 'true';
      }
    });
```

AzmiTouil (changes requested):

webapp/portlet/src/main/webapp/vue-apps/spaces-list/components/ExoSpaceCardFront.vue
MayTekayaa: Done for the verification space, but for the event name, it is already created and used with this title.
Do I have to change it for every use of the event? WDYT @AzmiTouil ?

MayTekayaa: Done!

AzmiTouil (approved):

PR #2212, opened 2023-01-23
fix: New activity toast appeared despite applying the activity stream  filter - EXO-61600

GouadriaHanen:
prior to this change,  the "New posts" chips  is displayed with all stream filter case.
after this change, the "New posts" chips should only be displayed if the filter applied is to "Any Activity"
If filtered to any other filter, then no chips at all

### DIFF
--- a/component/api/pom.xml
+++ b/component/api/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <artifactId>social-component</artifactId>
     <groupId>org.exoplatform.social</groupId>
-    <version>6.4.x-exo-SNAPSHOT</version>
+    <version>6.4.x-maintenance-SNAPSHOT</version>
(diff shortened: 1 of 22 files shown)


ahamdi (commented):
I am not sure but when we select spaces I host or my favorite spaces, shouldn't we get a notification for that ?

GouadriaHanen:
> I am not sure but when we select spaces I host or my favorite spaces, shouldn't we get a notification for that ?
The problem here is related to the space where the activity had been posted.
It may not part of my fav space or of my spaces that I host

ahamdi:
Will it be hard to get the information of the source of the activity (which space) and what if the activity belongs to it ?

ahamdi (approved):

PR #2223, opened 2023-01-24
feat : Apply quick search rules into space members App - EXO-61920 -Meeds-io/MIPs#39

sofyenne:
This change is going to set the end of typing timeout to 300 ms for searching into space members app .


azayati (approved):
LGTM

PR #2229, opened 2023-01-24
fix: fix user full name and popover after updating user contact in kudos activity - EXO-59990

GouadriaHanen:
prior to this change, when the user had updated his contact info (his name or job title), the user popover in the kudos activity still displayed old user info, since the activity is cached
prior to this change, after updated user contact, activity cache is cleared


ahamdi (commented):
@GouadriaHanen  We have to close this PR right ?

GouadriaHanen:
> @GouadriaHanen We have to close this PR right ?

to be close yep!

PR #2279, opened 2023-01-31
fix: Label is not I18Ned in Add and Edit attribute drawers - EXO-6204…

mkrout:


azayati (approved):
PR title to be changed for the commit message

PR #2295, opened 2023-02-08
fix: Resolve the PostgreSQL error during server startup by adding missing sequences - EXO-62184

sofyenne:
This change is going to resolve the PostgresSQL error during the server startup by adding the messing sequences.

### DIFF
--- a/component/core/src/main/resources/db/changelog/social-rdbms.db.changelog-1.0.0.xml
+++ b/component/core/src/main/resources/db/changelog/social-rdbms.db.changelog-1.0.0.xml
@@ -893,5 +893,9 @@
             ALTER TABLE SOC_ACTIVITY_SHARE_ACTIONS MODIFY COLUMN TITLE longtext CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
         </sql>
     </changeSet>
+    <changeSet author="social" id="1.0.0-90" dbms="postgresql">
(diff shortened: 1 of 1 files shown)


component/core/src/main/resources/db/changelog/social-rdbms.db.changelog-1.0.0.xml
```diff
@@ -888,5 +888,9 @@
     <changeSet author="social" id="1.0.0-88" dbms="hsqldb">
         <createSequence sequenceName="SEQ_SOC_LABELS_ID" startValue="1"/>
     </changeSet>
+    <changeSet author="social" id="1.0.0-89" dbms="hsqldb,oracle,postgresql">
```
ahamdi: No need to add hsqldb as sequences are already added for that DB, and we do no more support Oracle

ahamdi (changes requested):

component/core/src/main/resources/db/changelog/social-rdbms.db.changelog-1.0.0.xml
sofyenne: fixed

ahamdi:
@sofyenne could you fix the conflict ?

sofyenne:
> 

@ahamdi  resolved !

ahamdi (approved):

PR #2297, opened 2023-02-09
fix: Space is not correctly retrieved for space popover actions - EXO-62231

IlhemEssaadi:
Priori to this fix, popover menu actions does not get correctly the space if opened in a page outside a space since it relies on the global variable **exo.env.portal.spaceId** that is present only when a space is opened.
The fix changes the identity id with the local property spaceId instead of the global variable exo.env.portal.spaceId


boubaker:
Please change title and description to reflect the fix in context of Meeds only without eXo addons context and names.

ahamdi (dismissed):

ahamdi (changes requested):
@IlhemEssaadi Could you update title & description ? this is a Meeds project and we should not mention consequent issues in eXo addons, but just current issues on Meeds !

ahamdi (approved):

PR #2304, opened 2023-02-13
fix: Profile admin modified contact field is displayed empty on edit mode - EXO-62451

GouadriaHanen:
prior to this change, after modifying the contact field and then canceling, the modification is displayed instead of the original contact info since the setting object is linked to the original setting object resulting that the changes will be bidirectional.
After this change, the new modification is displayed only after saving the modifications by creating a copy of the setting object with no reference to objects linked to the original one

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/menu/ProfileSettingsActionMenu.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/menu/ProfileSettingsActionMenu.vue
@@ -91,7 +91,7 @@ export default {
   },
   methods: {
     edit(){
-      this.$root.$emit('open-settings-edit-drawer', this.setting);
+      this.$root.$emit('open-settings-edit-drawer', JSON.parse(JSON.stringify( this.setting)));
     },
     moveUp(){
       this.$root.$emit('move-up-setting', this.setting);


webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/menu/ProfileSettingsActionMenu.vue
```diff
@@ -91,7 +91,7 @@ export default {
   },
   methods: {
     edit(){
-      this.$root.$emit('open-settings-edit-drawer', this.setting);
+      this.$root.$emit('open-settings-edit-drawer', JSON.parse(JSON.stringify( this.setting)));
```
ahamdi, line 94: Could you give more details why we need to strigify then parse the JS Object ?

GouadriaHanen: to avoid using the same object (the changes will be bidirectional) we must parse it to a new JSON object

ahamdi: @GouadriaHanen I see , then you just wanted to create a copy of the setting object with no reference to objects linked to the original one. It would be helpful if you put more details about your solution and the cause in the PR description and

GouadriaHanen: done!

ahamdi (approved):

PR #2323, opened 2023-02-15
fix: Profile admin modified contact field is displayed empty on edit mode - EXO-62451 (#2304)

GouadriaHanen:
prior to this change, after modifying the contact field and then canceling, the modification is displayed instead of the original contact info since the setting object is linked to the original setting object resulting that the changes will be bidirectional. After this change, the new modification is displayed only after saving the modifications by creating a copy of the setting object with no reference to objects linked to the original one


azayati (approved):
LGTM

PR #2328, opened 2023-02-15
fix: Add IDs to Contact information Blocks - MEED-1790

boubaker:
1/ Prior to this change, the Meeds-io/MIPs#38 has removed the ids in profile contact information section. This change will rework components to allow having using identifier for each displayed profile property.

2/ In addition, this change will apply `npm run lint` formatting on newly introduced blocks.

3/ Prior to this change, the profile update triggering was changed to not send the modified user information in event anymore. this leaded to bugs in Profile Header and Profile displayed in Left Menu. This change will improve updating the profile strategy in both portlets to refresh right away the user profile from backend once updated. In addition, this will delete outdated usage of v-cacheable directive.

( Will be available for **Builders** only for now )


boubaker:
Will avoid changing the behavior worked in MIP 38, @rdenarie I let you take care of issues created for this. Thanks.

PR #2333, opened 2023-02-17
Revert "fix: Fix npm run lint error on webapp/portlet module - Meeds-io/meeds#693"

boubaker:
This reverts commit 84683edcb72303720bc67f531fd94894b9f847a2 which has been pushed without dedicated PR.

### DIFF
--- a/webapp/portlet/package.json
+++ b/webapp/portlet/package.json
@@ -8,8 +8,8 @@
   },
   "scripts": {
     "watch": "webpack --config webpack.dev.js --progress --color --watch",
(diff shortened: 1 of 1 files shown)


boubaker:
This PR reverts a commit that has been pushed directly on develop without PR and which makes a quality check regression. In fact the new pattern doesn't cover all files for ESLint check.

boubaker:
@rdenarie I let you rework the pattern to make one that works on Windows OS as well as the current one that works on Linux OS.

mkrout (approved):

PR #2334, opened 2023-02-17
fix: Fix npm run lint error on webapp/portlet module with windows - Meeds-io/meeds#693

azayati:
Prior to this fix, npm run lint on webapp/portlet module is working on for linux and not for windows since the pattern used for lint script is not suitable for windows.
After this change, we ensure to use the pattern \\"./src/main/webapp/vue-apps/**\\" instead of './src/main/webapp/vue-apps/**' which will be suitable for both windows and linux.

### DIFF
--- a/webapp/portlet/package.json
+++ b/webapp/portlet/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "watch": "webpack --config webpack.dev.js --progress --color --watch",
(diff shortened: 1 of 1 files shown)


azayati:
To be verified on linux @boubaker @rdenarie

rdenarie:
coonfirmed on my side, it is working

boubaker (approved):

mkrout (approved):

PR #2336, opened 2023-02-20
fix: fix space and user popover position - EXO-62433 - Meeds-io/meeds#706

IlhemEssaadi:
Prior to  this change,When I put the cursor on the space/user popup then scroll the popover is still displayed.After this change, we force to close the user popover when scrolling.


ahamdi (commented):
@IlhemEssaadi Could you fix the conflict ?

IlhemEssaadi:
> @IlhemEssaadi Could you fix the conflict ?

Done!

ahamdi (approved):

PR #2340, opened 2023-02-20
fix: No alert when adding an invalid URL to profile - EXO-62604

mkrout:
Prior to this change when the user edits his profile information and add or edit a URL field with less than 10 chars, nothing happens, the profile is not updated since the URL should be more than 10 chars, and no alert is displayed.
After this modification the limit of 10 chars is removed, and the url should be valid, if the text added is not a valid url, the user well get an info alert telling that the URL won't be considered as is as it is not an actual link
In addition, this change will also apply npm run lint formatting not formatted vue files.


ahamdi (approved):

rdenarie:
Meeds issue reference is missing. Please add before pushing to develop

PR #2342, opened 2023-02-20
fix: change waitForEndTyping by  waitForInitializing - EXO-61790

GouadriaHanen:
prior to this change, a js error "  this.waitForEndTyping is not a function" when searching users since waitForEndTyping no more exists replaced by waitForInitializing
after this change, searching users works perfectly


ahamdi (approved):

rdenarie:
@GouadriaHanen When you will backport this fix in develop, can you ensure to change the task id which seems not the good one ?
In addition, the meeds issue is missing in the ocmmit message

PR #2353, opened 2023-02-22
fix: Profile - username and Position not updated directly in the profile's header after add/edit - EXO-62664 - Meeds-io/meeds#756 

GouadriaHanen:
prior to this change, after saving/editing the username or user position are not updated directly in the profile's header 
since the data from the event 'userModified' in this case is null
after this change, the modification is well displaying by refreshing the user properties


rdenarie:
Should we passe the object user in the detail of the event ?

webapp/portlet/src/main/webapp/vue-apps/profile-header/components/ProfileHeader.vue
```diff
@@ -146,8 +146,10 @@ export default {
     document.addEventListener('userModified', event => {
       if (event && event.detail && event.detail !== this.user) {
         this.user = Object.assign({}, this.user, event.detail);
-        this.$nextTick().then(() => this.$root.$emit('application-loaded'));
+      } else  {
```
ahamdi: This is a forcing to reload that is not needed. Could you check why the userModified is null ?

GouadriaHanen: the event is dispatched by ProfileContactInformationDrawer which has only the user's properties to modify. 
https://github.com/Meeds-io/social/blob/7f85b69e9c7edcbbfed6485087f05eee94c865f7/webapp/portlet/src/main/webapp/vue-apps/profile-contact-information/components/ProfileContactInformationDrawer.vue#L205

GouadriaHanen: FYI @rdenarie

ahamdi: @GouadriaHanen the check inside the event listener should be updated without the need to call refresh explicitly which will reload data from server.
you can convert the array of properties in https://github.com/Meeds-io/social/blob/7f85b69e9c7edcbbfed6485087f05eee94c865f7/webapp/portlet/src/main/webapp/vue-apps/profile-contact-information/components/ProfileContactInformationDrawer.vue#L205 to a user object  then send it in the event data . this way you won't need to change in [ProfileHeader.vue](https://github.com/Meeds-io/social/pull/2353/files#diff-8a7df92ac03c601f3d8d0748d9960ec49280076846328494cb5c16bfe7b57256)

GouadriaHanen: done!

ahamdi (dismissed):

ahamdi (approved):

PR #2354, opened 2023-02-22
fix: Search result are replaced by original request because it is tool ong to execute - EXO-61970 - Meeds-io/meeds#758 (#2296)

GouadriaHanen:
prior to this change, the Search result of users is replaced by the original request because it takes too long to retrieve the result after this change when searching people the first fetch is aborted

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleCardList.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleCardList.vue
@@ -103,6 +103,7 @@ export default {
     users: [],
     limitToFetch: 0,
     originalLimitToFetch: 0,
+    abortController: null,
   }),
   computed: {
     profileActionExtensions() {
@@ -131,7 +132,7 @@ export default {
         this.searchPeople();
         return;
       }
-      this.waitForInitializing();
+      this.searchPeople();
 
     },
     limitToFetch() {
@@ -140,7 +141,7 @@ export default {
     filter() {
       this.searchPeople();
     },
-  },
+  }, 
   created() {
     this.originalLimitToFetch = this.limitToFetch = this.limit;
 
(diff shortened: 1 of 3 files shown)


ahamdi (dismissed):

webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleCardList.vue
```diff
@@ -251,6 +251,15 @@ export default {
...
 
+    },
+    waitForInitializing() {
+      window.setTimeout(() => {
+        if (this.initialized) {
+          this.searchPeople();
+        } else {
+          this.waitForInitializing();
+        }
+      }, 50);
```
boubaker:
```suggestion
```

Add a watch on `initialized` instead :

```javascript
watch: {
  initialized() {
    if (this.initialized && this.keyword?.length) {
          this.searchPeople();
    }
  }
},
```

boubaker (changes requested):

webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleCardList.vue
boubaker: BTW, it would be more elegant IMO to abort first fetch and execute right a way the user query by using :
https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort

```javascript
  if (this.controller) {
    this.controller.abort();
  }
  this.controller = new AbortController();,
  ...
  // Pseudo algorithm, this will allow to have `this.signal` variable in `getUsers` method without adding signal a argument which will break the API signature
  this.$userService.getUsers.apply({singal: this.controller.signal}, this.keyword, this.offset...)
  .finally(() => {
...
    this.controller = null;
  });
....
```

boubaker (approved):

PR #2355, opened 2023-02-22
fix: Search result are replaced by original request because it is too long to execute - EXO-61970 - Meeds-io/meeds#758   (#2296)

GouadriaHanen:
prior to this change, the Search result of users is replaced by the original request because it takes too long to retrieve the result after this change, the search request starts after initializing the page after retrieving the first request of user


ahamdi (approved):

rdenarie:
cc @ahamdi if the pr on develop modified, we should revert this commit in order to have the correct fix

PR #2360, opened 2023-02-23
fix: avoid displaying and indexing user technical properties - EXO-62724 - meeds-io/meeds#762

### LINKED ISSUE #762: Technical properties are displayed in user profile
**Steps to reproduce**
Steps:
 - Configure openid on exo plf
 - launch server
 - user1 on borwser 1 connect through openid
 - on browser 2 connect with root

**Current behaviour**
user1 open his own profile : social contact info displayed
root searches for user1 in people page: user1 isn't displayed in search results

**Expected behavior**
user1 open his own profile : social contact info shouldn't be added in contact info 
root searches for user1 in people page: user1 profile should be indexed thus  displayed in search results

ahamdi:
When login using openID provider, the related user informations are saved in the user profile an then indexed in ElasticSearch which causes an indexing error. Those properties are also displayed in the user profile while they are technical information that should not be viewed.
The fix sets user properties to inactive (to not display them in the user profile) and do not index them until they are activated by the administrator from the Profile properties settings.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/listeners/SocialUserProfileEventListenerImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/listeners/SocialUserProfileEventListenerImpl.java
@@ -109,9 +109,9 @@ public class SocialUserProfileEventListenerImpl extends UserProfileEventListener
       ProfilePropertySetting profilePropertySetting = new ProfilePropertySetting();
       profilePropertySetting.setPropertyName(propertyName);
       profilePropertySetting.setMultiValued(false);
-      profilePropertySetting.setActive(true);
+      profilePropertySetting.setActive(false);
       profilePropertySetting.setEditable(false);
-      profilePropertySetting.setVisible(true);
+      profilePropertySetting.setVisible(false);
       profilePropertySetting.setParentId(null);
       try {
         profilePropertyService.createPropertySetting(profilePropertySetting);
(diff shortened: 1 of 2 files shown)


component/core/src/main/java/org/exoplatform/social/core/listeners/SocialUserProfileEventListenerImpl.java
```diff
@@ -109,9 +109,9 @@ private void updateProfilePropertySettings(String propertyName, ProfilePropertyS
       ProfilePropertySetting profilePropertySetting = new ProfilePropertySetting();
       profilePropertySetting.setPropertyName(propertyName);
       profilePropertySetting.setMultiValued(false);
-      profilePropertySetting.setActive(true);
+      profilePropertySetting.setActive(false);
       profilePropertySetting.setEditable(false);
-      profilePropertySetting.setVisible(true);
+      profilePropertySetting.setVisible(false);
```
rdenarie, line 114: Does this modification will create initial properties not active and not visible ?

Or only added properties (added in interface or programmatically like OIDC properties) ?

ahamdi: @rdenarie This will set just added properties as inactive and invisible, to let the administrator decide whether he should activate them or not.
initial properties are activated and set visible using the XML configuration.

rdenarie (approved):

PR #2364, opened 2023-02-24
fix :fix save button not disabled in profile information drawer when there is no change - EXO-62627 - Meeds-io/meeds#763

IlhemEssaadi:
Prior to this change, when we access to edit the profile contact informations without changing anything, the save button is proposed not disabled and not clickable, so the edit drawer cannot be closed. After this fix, the save button is disabled when we open the drawer and no change is detected, and it is enabled when we detect a change in any of the profile information.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/profile-contact-information/components/ProfileContactEditMultiFieldSelect.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/profile-contact-information/components/ProfileContactEditMultiFieldSelect.vue
@@ -20,7 +20,8 @@
       class="ignore-vuetify-classes align-end flex-grow-1"
       maxlength="2000"
       ref="multiInput"
-      @change="$emit('propertyUpdated')">
+      @change="$emit('propertyUpdated')"
+      @input="$emit('propertyUpdated')">
     <v-icon
       small
       class="removeMultiFieldValue error--text"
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/profile-contact-information/components/ProfileContactEditMultiField.vue
```diff
@@ -50,6 +52,9 @@ export default {
     propertyUpdated(){
       this.$emit('propertyUpdated',this.property);
     },
+    propertyChanged(){
+      this.$emit('propertyChanged',this.property);
```
ahamdi: why should we have two different functions for property changes ?

ahamdi (changes requested):

webapp/portlet/src/main/webapp/vue-apps/profile-contact-information/components/ProfileContactEditMultiField.vue
IlhemEssaadi: > why should we have two different functions for property changes ?

The propertyUpdated function is used to detect the @change event that occurs when the element loses focus, after the content has been changed. But in my case, I need to detect the change of the property value immediately, so I added the propertyChanged function to detect the @input event that occurs immediately after the property value has changed.

ahamdi: @IlhemEssaadi Yes but we can use the same function for both events.

IlhemEssaadi: > @IlhemEssaadi Yes but we can use the same function for both events.

Yes, Done!

ahamdi (approved):

PR #2365, opened 2023-02-27
Cleanup css files

boubaker:


boubaker:
Worked in Meedsio/MIPs#52 PRs

PR #2366, opened 2023-02-27
fix: ignore openId properties while synchronize user profile - EXO-62724 - meeds-io/meeds#762

ahamdi:
Some technical properties saved in Portal user profile are used for technical purposes and not intended to be shared with the social user profile. Those properties contain DOT  '.' character which may cause problems when indexing those properties.
The fix adds the possibility to define properties to ignore when synchronizing both profiles and replaces dot '.' characters with underscore '_' for indexed filed names .


mkrout (dismissed):
LGTM

rdenarie:
Can you fix the code smell ?

mkrout (approved):

PR #2388, opened 2023-03-07
fix: stabilize social build for SocialUserProfileEventListenerImplTest - EXO-63065

ahamdi:
Unit test `SocialUserProfileEventListenerImplTest#testIgnoreExcludedProfileProperties` fails sometimes randomly, the JPM system properties are sometimes added after initiation of the listener causing failure for asserted expression.
The fix includes the testsing system property in the surefire test configuration instead of the kernel internal mechanism using using ExtendedPropertyConfigurator

### DIFF
--- a/component/api/pom.xml
+++ b/component/api/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <artifactId>social-component</artifactId>
     <groupId>org.exoplatform.social</groupId>
-    <version>6.5.x-exo-SNAPSHOT</version>
+    <version>6.5.x-maintenance-SNAPSHOT</version>
   </parent>
   <artifactId>social-component-api</artifactId>
   <name>eXo PLF:: Social API</name>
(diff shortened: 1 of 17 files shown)


rdenarie:
@ahamdi 
We are currently merging meeds-io/MIPs#42
As I understood, the modification done by @boubaker should fix this problem. I think at least you should wait that the new code propagate on feature/maintenance before continuiing this modification.

ahamdi:
> @ahamdi We are currently merging [Meeds-io/MIPs#42](https://github.com/Meeds-io/MIPs/issues/42) As I understood, the modification done by @boubaker should fix this problem. I think at least you should wait that the new code propagate on feature/maintenance before continuiing this modification.

@rdenarie Ok I will check it

boubaker:
Yep, in fact the property definition has been moved to RootContainer to ensure that it's loaded before the Service.

PR #2390, opened 2023-03-07
fix: Change in profile management field is not applied to profile until the cache is cleared - EXO-62960 -  Meeds-io/meeds#786

IlhemEssaadi:
Before this fix, after changing the value of a field's switch option in profile management, the option was not applied in the user's profile information until the cache was cleared.After this fix, we use the list of property settings in order to calculate the Etag, so that each change in property value force to takes the new value of user's profile information.


rdenarie:
In fact, it should not be the front which say "I dont want the cache". But the back should say : "hey : the answer is not modified, use your cache" OR "the answer is modified since your last request, I give you the new answer"

The modification you propose is not performant as it completly by pass the cache

ahamdi (changes requested):
Could you resolve the following points : 
 - When a field is changed, we should calculate correctly Etag to push a new one to help the browser detect that and ask for a new response
 - Description should be added accordingly

ahamdi (approved):

PR #2392, opened 2023-03-08
fix: TOP bar menu isn't displayed - EXO-62924 -  Meeds-io/meeds#787

GouadriaHanen:
refactor get navigations methods so the siteType  param will be passed as path param, not as QueryParam


ahamdi (changes requested):
We need a Meeds issue here

ahamdi (approved):

PR #2404, opened 2023-03-13
fix : Can't save edited profile attributes without additional actions - Meeds-io/meeds#795 - EXO-63095

sofyenne:
Before to this change we were unable to save any edited property of a profile attribute due of the required text filed label.  
I don't think there is a need to display an empty text filed when there is no label property value to edit on the edit profile attribute drawer.
After this change we will display the text filed label property when we have a value to edit , if that is'n the case we display only the add property label button to use it if we want to add an new property during the editing of an attribute .

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/drawers/ProfilePropertyLabels.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/drawers/ProfilePropertyLabels.vue
@@ -71,7 +71,7 @@ export default {
 
   methods: {
     addNewLabel() {
-      this.propertylabels.push( {language: '', label: '', objectId: this.id, objectType: this.labelsObjectType});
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/drawers/ProfilePropertyLabels.vue
```diff
@@ -71,7 +71,7 @@ export default {
 
   methods: {
     addNewLabel() {
-      this.propertylabels.push( {language: '', label: '', objectId: this.id, objectType: this.labelsObjectType});
+      this.propertylabels.push( {language: 'en', label: '', objectId: this.id, objectType: this.labelsObjectType});
```
ahamdi: Wouldn't it be better to select the default language of the user with the property **eXo.env.portal.language** ?

sofyenne: done !

ahamdi (approved):

PR #2410, opened 2023-03-14
fix: Profile new multi field sub attributes are not saved in profile - Meeds-io/meeds#801 - EXO-63188

GouadriaHanen:
prior to this change, after editing  a newly added attribute profile with sub-attributes, the changes are not displayed since the parent was saved as a simple attribute ( their children have null value)
after this change, the newly added attribute is treated as a parent attribute with children, and the edit and displayed are well saved


ahamdi (commented):
Coumd you add a UT and update the description of the PR ?

ahamdi (approved):

PR #2426, opened 2023-03-16
fix: Erroneous number of search results when using advanced people search - EXO-63148 -  Meeds-io/meeds#818

akhanfir:
Prior to this change, when Go to People page and make an advanced search on a field and confirm, The number of results displayed on the top of the page is lesser than the real number of cards displayed by 1 card. After this changes,  the number of results displayed have the real number of cards in the results.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/EntityConverterUtils.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/EntityConverterUtils.java
@@ -152,25 +152,32 @@ public class EntityConverterUtils {
       for (Entry<String, String> entry : properties.entrySet()) {
         String key = entry.getKey();
         String value = entry.getValue();
-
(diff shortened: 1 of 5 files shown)


webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleCardList.vue
```diff
@@ -241,9 +241,6 @@ export default {
         users = users.slice(0, this.limitToFetch);
         this.users = users;
         this.peopleCount = data && data.size && data.size || 0;
-        if (this.peopleCount > 0 && this.filter === 'all' && !this.keyword) {
```
ahamdi: Why was this ```-1``` present ? was it to not count root user ?

rdenarie (changes requested):
Can you explain
- WHY the number is not correct ?
- HOW you fix it ?

Theses information should be in the commit message
Thanks

akhanfir:
This pr is closed by an incorrect command line and I have replaced it with this pr :  https://github.com/Meeds-io/social/pull/2458

PR #2428, opened 2023-03-16
feat: Open Left menu when hovering the hamburger icon - MEED-1807 - Meeds-io/meeds#755

SaraBoutej:
This changes allows to add: 
- A ripple effect when hovering the hamburger menu
- Add Timeout of 5 ms before opening or closing the left menu with mouseover and mouseleave events

### DIFF
--- a/webapp/portlet/src/main/webapp/skin/less/portlet/HamburgerMenu/Style.less
+++ b/webapp/portlet/src/main/webapp/skin/less/portlet/HamburgerMenu/Style.less
@@ -90,6 +90,40 @@
       width: 16px;
       height: 16px;
     }
+    .ripple {
+      &:before {
+        content: "";
+        position: absolute;
+        background-color: rgba(0,0,0,.1);
+        width: 10%;
+        height: 25%;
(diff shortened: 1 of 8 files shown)


webapp/portlet/src/main/webapp/skin/less/portlet/HamburgerMenu/Style.less
```diff
@@ -90,6 +90,40 @@
       width: 16px;
       height: 16px;
     }
+    .ripple {
```
boubaker: please use ripple effect of button in Vuetify

SaraBoutej: I tried v-ripple directive, but it's applied only with click event not with hover event
cc @boubaker

boubaker: ok, we will have to check it together.

boubaker (approved):

PR #2436, opened 2023-03-17
fix: activity params keys case insensitivity block activity updates -EXO-62802 - Meeds-io/meeds#815

### LINKED ISSUE #815: Activity params keys case insensitivity block activity updates
Environment: 6.4.0-RC02 (Only on Mysql)

Step to reproduce:

Edit a document
On Only office editor, create a version with a comment. A post is generated in the stream
Like the post
Current behavior

- We can't like the post.

Expected behavior

-We can like the post.

hakermi:
Prior to this change, on mysql a the column of activity template params case insensitive by default, this have created an issue when attempting to like a comment activity and an error of duplicated entry exception is thrown.
This PR modifies the affected column by adding the binary keyword to make the column case sensitive

### DIFF
--- a/component/core/src/main/resources/db/changelog/social-rdbms.db.changelog-1.0.0.xml
+++ b/component/core/src/main/resources/db/changelog/social-rdbms.db.changelog-1.0.0.xml
@@ -900,4 +900,10 @@
         <createSequence sequenceName="SEQ_SOC_LABELS_ID" startValue="1"/>
     </changeSet>
 
+    <changeSet author="social" id="1.0.0-91" dbms="mysql">
+        <sql>
+            ALTER TABLE SOC_ACTIVITY_TEMPLATE_PARAMS MODIFY COLUMN TEMPLATE_PARAM_KEY NVARCHAR(255) BINARY;
+        </sql>
+    </changeSet>
+
 </databaseChangeLog>


component/core/src/main/resources/db/changelog/social-rdbms.db.changelog-1.0.0.xml
```diff
@@ -900,4 +900,10 @@
         <createSequence sequenceName="SEQ_SOC_LABELS_ID" startValue="1"/>
     </changeSet>
 
+    <changeSet author="social" id="1.0.0-91" dbms="mysql">
+        <sql>
+            ALTER TABLE SOC_ACTIVITY_TEMPLATE_PARAMS MODIFY COLUMN TEMPLATE_PARAM_KEY NVARCHAR(255) BINARY;
```
boubaker, line 905: what about emojis in TEMPLATE Params like kudos message .... does it work on MySQL with this change without `utf8mb4_unicode_ci` ? similar to 

https://github.com/Meeds-io/social/blob/c5734174023b5a0fb18e9297afd9a540601abd0c/component/core/src/main/resources/db/changelog/social-rdbms.db.changelog-1.0.0.xml#L563

hakermi: @boubaker But this changes only the param **key** column (It's value always in English as its the name of the param and not the param value isn't it ?) + it juts changes the collation to a binary one but not the collation type due to the `BINARY` keyword.

boubaker: ok, I see, thanks for explanation, ok for me.

boubaker (approved):

PR #2441, opened 2023-03-17
fix: Profile - Attribute position change not taken into account - EXO-63150

GouadriaHanen:
prior to this change, the added profiles properties from the XML file have the wrong order which lead to having duplication of order for some newly added properties
after this change, the properties in the XML file are fixed and distinct 
a change set is was added to fix the already saved ones

### DIFF
--- a/component/core/src/main/resources/db/changelog/social-rdbms.db.changelog-1.0.0.xml
+++ b/component/core/src/main/resources/db/changelog/social-rdbms.db.changelog-1.0.0.xml
@@ -904,5 +904,10 @@
             ALTER TABLE SOC_ACTIVITY_TEMPLATE_PARAMS MODIFY COLUMN TEMPLATE_PARAM_KEY NVARCHAR(255) BINARY;
         </sql>
     </changeSet>
+    <changeSet author="social" id="1.0.0-92">
+        <sql>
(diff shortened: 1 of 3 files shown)


rdenarie (changes requested):
I'm not confident with 2 things :

- upgrade-plugin : we should not have an UP in social, we have a module for that
- all the stuff around maxOrder

It is not possible to delete a settings so using id for order when creating the property settings seems ok for me.
The problem comes from the fact that in configuration, a order number is missing (11). 
So you fix the configuration, this is ok. 

You try to fix existing datas with an UP. It is big weapon for a little problem. As we are only 2 customer to use the feature for the moment (tribe and builders), could use a simple liquibase query which copy id column in order column once. It will, in the same reset and fix the missing order number. 

I think it is not problematic to loose the current order on tribe and builders. 
cc @ahamdi 

So the fix should contains : 

- xml configuration file fix
- liquibase instruction

ahamdi:
@rdenarie @GouadriaHanen Using Liquibase sounds good, especially that the issue does not concern many customers (just internal ones) Thanks

exo-swf:

rdenarie (approved):

PR #2444, opened 2023-03-21
fix: Deactivating sub attributes of IMs does not work int edit drawer - Meeds-io/meeds#797 - EXO-63174 (#2427)

Jihed525:
prior to this change, deactivated and invisible IMs are displayed in the user profile's edit contact info drawer since they are not restricted from being displayed after this change, deactivated and invisible IMs are not displayed anymore

(cherry picked from commit 1f47b9bdb6fb4ad329969f120bf6ab9e4df462bf)


rdenarie (changes requested):
This fix contains 2 commits in the task, I think one is missing

rdenarie (approved):

PR #2458, opened 2023-03-21
fix: Erroneous number of search results when using advanced people search - EXO-63148 - Meeds-io/meeds#818

akhanfir:
Prior to this change, when go to People page and make an advanced search on a field and confirm, The number of results displayed on the top of the page is less than the real number of cards displayed by 1 card. This problem occurs when the current user is not included in this list. To fix that, it is necessary to calculate the correct size of the list of people in backend by adding the current user in excludedIdentityList of filter.
After this changes, the number of results displayed is the size of the list of all the people who are involved in this search except the current user who should not be applied in this search.

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/api/UserRestResources.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/api/UserRestResources.java
@@ -49,7 +49,8 @@ public interface UserRestResources extends SocialRest {
                                     @QueryParam("offset") int offset,
                                     @QueryParam("limit") int limit,
                                     @QueryParam("returnSize") boolean returnSize,
-                                    @QueryParam("expand") String expand) throws Exception;
+                                    @QueryParam("expand") String expand,
+                                    @QueryParam("excludeCurrentUser") boolean excludeCurrentUser) throws Exception;
 
   /**
    * Creates an user
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
@@ -263,7 +263,8 @@ public class UserRestResourcesV1 implements UserRestResources, Startable {
           description = "Using the query param \"q\" to filter the target users, ex: \"q=jo*\" returns all the users beginning by \"jo\"."
                 + "Using the query param \"status\" to filter the target users, ex: \"status=online*\" returns the visible online users."
                 + "Using the query params \"status\" and \"spaceId\" together to filter the target users, ex: \"status=online*\" and \"spaceId=1*\" returns the visible online users who are member of space with id=1."
-                + "The params \"status\" and \"spaceId\" cannot be used with \"q\" param since it will falsify the \"limit\" param which is 20 by default. If these 3 parameters are used together, the parameter \"q\" will be ignored")
+                + "The params \"status\" and \"spaceId\" cannot be used with \"q\" param since it will falsify the \"limit\" param which is 20 by default. If these 3 parameters are used together, the parameter \"q\" will be ignored,"
+                + "the current user \"excludeCurrentUser\" will be excluded")
   @ApiResponses(value = {
     @ApiResponse(responseCode = "200", description = "Request fulfilled"),
     @ApiResponse (responseCode = "404", description = "Resource not found"),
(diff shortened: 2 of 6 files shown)


component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -454,6 +454,11 @@ public Response getUsersOrConnectionsByAdvancedFilter(@Context UriInfo uriInfo,
           filter.setUserType(userType);
         }
       }
+      if (settings != null) {
+        List<Identity> excludedIdentityList = new ArrayList();
+        excludedIdentityList.add(target);
+        filter.setExcludedIdentityList(excludedIdentityList);
```
ahamdi: Could this be improved like this, WDYT?
```suggestion
        filter.setExcludedIdentityList(Collections.singletonList(target));
```

rdenarie (changes requested):
Can you rebase the PR and fix conflict

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -315,9 +315,13 @@ public Response getUsers(@Context UriInfo uriInfo,
...
       }
     } else {
+      Identity target = identityManager.getOrCreateIdentity(OrganizationIdentityProvider.NAME, userId);
       ProfileFilter filter = new ProfileFilter();
       filter.setName(q == null || q.isEmpty() ? "" : q);
       filter.setEnabled(!isDisabled);
+      if (target != null && filter.getName().equals(q)) {
+        filter.setExcludedIdentityList(Collections.singletonList(target));
+      }
       if (!isDisabled) {
```
rdenarie: @ahamdi The getUsers function (and the restService in general) provide the list of users which match all parameters. There is no information about excluding the current user.
I'm afraid that excluding the current user in the rest service can lead to problems in other parts of the platform, where the current user is needed.

If this task is too complex to fix, we can postpone it to 6.4.1.

I think we should take reflexion time here

rdenarie (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
ahamdi: @rdenarie ok I see the point
we can add a new optional parameter to filter or not the connected user WDYT ?

rdenarie: Yes I think it is the better solution.
But is it possible for 6.4.0 ? or is it safer to do it properly in 6.4.1 ?

ahamdi: @rdenarie Yes very possible for 6.4.0

rdenarie: ok, so go with a new parameter in the function. 

I keep the possibility to push back in 6.4.1 if not fully finished at the end of the week

exo-swf:

component/service/src/main/java/org/exoplatform/social/rest/api/UserRestResources.java
```diff
@@ -49,7 +49,8 @@ public abstract Response getUsers(@Context UriInfo uriInfo,
                                     @QueryParam("offset") int offset,
                                     @QueryParam("limit") int limit,
                                     @QueryParam("returnSize") boolean returnSize,
-                                    @QueryParam("expand") String expand) throws Exception;
+                                    @QueryParam("expand") String expand,
+                                    @QueryParam("excludedIdentity") boolean excludedIdentity) throws Exception;
```
rdenarie:
```suggestion
                                    @QueryParam("excludeCurrentUser") boolean excludeCurrentUser) throws Exception;
```

rdenarie (changes requested):

rdenarie (dismissed):

rdenarie:
@akhanfir there is a build problem. Can you check ? 
I think there is a comment to update in PeopleRestService

rdenarie (approved):

PR #2460, opened 2023-03-22
feat: People page enhancement - MEED-1838 - Meeds-io/Meeds#746

SaraBoutej:
This change allows to: 
 - Reduce the filter dropdown width to 100px max
 - Remove the filter label in advanced filters and Keep the number of filters applied if applied

### DIFF
--- a/webapp/portlet/src/main/webapp/skin/less/common/PeopleList/Style-mobile.less
+++ b/webapp/portlet/src/main/webapp/skin/less/common/PeopleList/Style-mobile.less
@@ -5,6 +5,9 @@
   .inputPeopleFilter {
     max-width: none;
   }
+  .selectPeopleFilter {
+    width: 100px;
+  }
 }
 
 .inputPeopleFilter {
--- a/webapp/portlet/src/main/webapp/skin/less/common/PeopleList/Style-mobile.less
+++ b/webapp/portlet/src/main/webapp/skin/less/common/PeopleList/Style-mobile.less
@@ -1,23 +1,35 @@
-#peopleListToolbar {
-  .addNewPeopleButton {
-    padding: 0 !important;
-  }
-  .inputPeopleFilter {
-    max-width: none;
-  }
-  .selectPeopleFilter {
-    width: 100px;
+@media (max-width: @maxMobileWidth) {
+  #peopleListApplication {
+    #peopleListToolbar {
+      .addNewPeopleButton {
+        padding: 0 !important;
+      }
+      .inputPeopleFilter {
+        max-width: none;
+      }
+    }
+    
+    .inputPeopleFilter {
+      .v-input__prepend-inner {
+        display: none;
+      }
+    
+      input{
+        padding-top: 0;
+      }
+    }
   }
 }
 
-.inputPeopleFilter {
-  .v-input__prepend-inner {
-    display: none;
-  }
-
-  input{
-    padding-top: 0;
+@media (max-width: @maxLargeTabletWidth) {
+  #peopleListApplication {
+    #peopleListToolbar {
+      .selectPeopleFilter {
+        width: 100px;
+      }
+    }
   }
 }
+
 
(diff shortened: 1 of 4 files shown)


webapp/portlet/src/main/webapp/skin/less/common/PeopleList/Style-mobile.less
```diff
@@ -5,6 +5,9 @@
   .inputPeopleFilter {
     max-width: none;
   }
+  .selectPeopleFilter {
+    width: 100px;
```
boubaker: Why `100px` and why adding anew file for tablet ? please reduce the CSS by merging both tablet & mobile styles.
Will we apply this same UX in other filters in other pages, if yes, we should add a CSS helper to define the filter field style for all applications. By example, instead of using `width: 100px` we can use `max-width: 100px` in new CSS helper and let the container decide which width to add for the input.

webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleToolbar.vue
```diff
@@ -34,13 +34,13 @@
...
         class="btn px-2 btn-primary"
+        :min-width="iconWidth"
         outlined
         @click="openPeopleAdvancedFilterDrawer()">
-        <v-icon small class="primary--text mr-1">fa-sliders-h</v-icon>
-        <span class="d-none font-weight-regular caption d-sm-inline mr-1">
-          {{ $t('profile.label.search.openSearch') }}
-          {{ advancedFilterCountDisplay }}
-        </span>
+        <v-icon small class="primary--text mr-lg-1">fa-sliders-h</v-icon>
```
boubaker:
```suggestion
        <v-icon small class="primary--text me-lg-1">fa-sliders-h</v-icon>
```
RTL, remember we don't use `mr` nor `ml` anymore.
Besides, why adding margin to an icon, by experience, it's not a good behavior.

webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleToolbar.vue
```diff
@@ -34,13 +34,13 @@
...
+        <span class="d-none font-weight-regular caption d-lg-inline mr-1">
+          {{ $t('profile.label.search.openSearch') }} </span>
+        <span class="font-weight-regular caption ml-1"> {{ advancedFilterCountDisplay }} </span>
```
boubaker:
```suggestion
        <span class="font-weight-regular caption ms-1"> {{ advancedFilterCountDisplay }} </span>
```
same

webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleToolbar.vue
```diff
@@ -100,7 +101,7 @@ export default {
       return this.advancedFilterCount > 0 ? `(${this.advancedFilterCount})`:'';
     },
     isMobile() {
-      return this.$vuetify.breakpoint.width < 768;
+      return this.$vuetify.breakpoint.name === 'sm';
```
boubaker:
```suggestion
      return this.$vuetify.breakpoint.width < 768;
```
nope, the `xs` breakpoint is a Mobile breakpoint too, please use CSS classes instead to display or hide elements with `d-none d-md-flex` by example.

webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleToolbar.vue
SaraBoutej: done!

webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleToolbar.vue
SaraBoutej: done!

boubaker (approved):

PR #2480, opened 2023-03-28
fix: Unable to invite user in a room, or in a space - EXO-63472

rdenarie:
Before the fix, the es query built to search a user have a syntax error, in some case, a ',' is added and should not This fix add tests to add the ',' only when necessary


ahamdi (approved):

boubaker:
We may need contextualize messages outside chat context

PR #2482, opened 2023-03-28
fix: Replace the error alert with a global notification alert when creating a space with an existing name - Meeds-io/meeds#859 - EXO-62856

sofyenne:
Before to this change when we try to create a new space with an existing space name an error alert was displayed and doesn't disappear , this change is going to replace this error alert with the global notification alert   .

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/spaces-list/components/ExoSpaceFormDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/spaces-list/components/ExoSpaceFormDrawer.vue
@@ -455,7 +455,7 @@ export default {
             // eslint-disable-next-line no-console
             console.warn('Error creating space ', this.space, e);
             if (String(e).indexOf('SPACE_ALREADY_EXIST') >= 0) {
-              this.error = this.$t('spacesList.error.spaceWithSameNameExists');
+              this.displayAlert(this.$t('spacesList.error.spaceWithSameNameExists'), 'error');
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/spaces-list/components/ExoSpaceFormDrawer.vue
```diff
@@ -455,7 +455,7 @@ export default {
             // eslint-disable-next-line no-console
             console.warn('Error creating space ', this.space, e);
             if (String(e).indexOf('SPACE_ALREADY_EXIST') >= 0) {
-              this.error = this.$t('spacesList.error.spaceWithSameNameExists');
+              this.displayAlert(this.$t('spacesList.error.spaceWithSameNameExists'), 'error');
             } else if (String(e).indexOf('INVALID_SPACE_NAME') >= 0) {
               this.error = this.$t('spacesList.error.InvalidSpaceName');
```
ahamdi: Could you refactor the whole file and use the global alert instead of the variable `alert` ? and remove the error variable if no more needed

ahamdi (changes requested):

webapp/portlet/src/main/webapp/vue-apps/spaces-list/components/ExoSpaceFormDrawer.vue
sofyenne: Done !

ahamdi (approved):

PR #2485, opened 2023-03-29
fix: Refresh cache is needed after renaming profile attribute - Meeds-io/meeds#867 - EXO-62875

sofyenne:
Prior to this change , when we retrieved the user by id with the expended settings , we used the ProfilePropertySetting list for the Etag value , because of this we needed to refresh the cache after updating a profile label value as the ProfilePropertySetting did not contain the list of ProfileLabels. After this change, we will use the ProfilePropertySettingEntity for the Etag value, as it contains the ProfileLabel list.

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
@@ -560,7 +560,7 @@ public class UserRestResourcesV1 implements UserRestResources, Startable {
 
     String expandedSettings = expand;
     if (expand != null && expand.contains("settings")) {
-      expandedSettings = String.valueOf(Objects.hash(profilePropertyService.getPropertySettings()));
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/ProfileSettings.vue
```diff
@@ -105,6 +105,7 @@ export default {
           this.getSettings();
         }
         this.displayMessage({type: 'success', message: this.$t('profileSettings.update.success.message')});
+        document.dispatchEvent(new CustomEvent('userPropertiesModified'));
```
ahamdi: I am not sure how this event from the profile Settings will be triggered to the user profile which is in another page. Could you recheck ?

sofyenne: done !

ahamdi (approved):

PR #2496, opened 2023-04-10
fix: New multi valued profile attribute not saved - Meeds-io/meeds#889 - EXO-63578

sofyenne:
Prior to this change, newly added multi-valued profile attributes were not being saved. The issue was that the UserRestResourceV1 was unable to save a property without a name. However, in the case of multi-valued properties, there is no need for the children property to have a name because they will be displayed under the parent property's name. The same applies to the visibility check. There is no need to check if the value is visible when we have a visible attribute.

Finally, in the case of multi-valued profile properties, the child value property is considered as the value of the parent property. Therefore, there is no need to check the name and visibility of the children since these are already checked for the parent.

After this change, multi-valued profile properties can be saved and edited from the profile contact information drawer."

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/profile-contact-information/components/ProfileContactEditMultiField.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/profile-contact-information/components/ProfileContactEditMultiField.vue
@@ -20,7 +20,7 @@
     </div>
     <v-flex v-for="(childProperty, i) in property.children" :key="i">
       <profile-contact-edit-multi-field-select
-        v-if="childProperty.isNew || childProperty.visible && childProperty.active && childProperty.value"
+        v-if="childProperty.isNew || (childProperty.visible && childProperty.active && childProperty.value) || (property.multiValued && property.active && property.visible && childProperty.value)"
         :property="childProperty"
         :properties="property.children"
         :multi-valued="property.multiValued"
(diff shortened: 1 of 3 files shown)


webapp/portlet/src/main/webapp/vue-apps/profile-contact-information/components/ProfileContactEditMultiField.vue
```diff
@@ -20,7 +20,7 @@
     </div>
     <v-flex v-for="(childProperty, i) in property.children" :key="i">
       <profile-contact-edit-multi-field-select
-        v-if="childProperty.isNew || childProperty.visible && childProperty.active && childProperty.value"
+        v-if="childProperty.isNew || (childProperty.visible && childProperty.active && childProperty.value) || (property.multiValued && property.active && property.visible && childProperty.value)"
```
ahamdi, line 23: `(childProperty.visible && childProperty.active && childProperty.value) || (property.multiValued && property.active && property.visible && childProperty.value)` could be simplified with `childProperty.visible && childProperty.active && childProperty.value` as the value of `property.multiValued` is ignored

ahamdi (changes requested):
Could you provide a clearer description on the cause of the issue? what was exactly causing the issue ?

webapp/portlet/src/main/webapp/vue-apps/profile-contact-information/components/ProfileContactEditMultiField.vue
sofyenne: In this case, the child property has only a value, and all other fields take the default value, so 'visible' and 'active' are set to false by default !

sofyenne:
> Could you provide a clearer description on the cause of the issue? what was exactly causing the issue ?

@ahamdi  description updated !

ahamdi (approved):

PR #2497, opened 2023-04-10
Fix : Technical label displayed instead of label on people advanced filter button - Meeds-io/meeds#890 - EXO-63060

sofyenne:
Prior to this change, it took a few seconds for the advanced filter button label to appear. The problem was that it was using an existing technical label from the gamification resource addon. With this change, a new technical label will be added to the people list application resource and displayed through a computed value.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleToolbar.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleToolbar.vue
@@ -41,7 +41,7 @@
           @click="openPeopleAdvancedFilterDrawer()">
           <v-icon small class="primary--text me-lg-1">fa-sliders-h</v-icon>
           <span class="d-none font-weight-regular caption d-lg-inline me-1">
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/people-list/components/PeopleToolbar.vue
```diff
@@ -103,6 +103,9 @@ export default {
     advancedFilterCountDisplay() {
       return this.advancedFilterCount > 0 ? `(${this.advancedFilterCount})`:'';
     },
+    advancedFilterButtonLabel() {
```
ahamdi: Why do we need to have this computed property ?

sofyenne: There's no need for a computed property when we have a static label. It can be displayed using interpolation! PR updated @ahamdi

ahamdi (approved):

PR #2519, opened 2023-04-28
feat: Display scheduled navigation nodes in top bar navigation menu and  hamburger menu navigation - EXO-63240 - Meeds-io/MIPs#51

IlhemEssaadi:
Prior to this change, top bar navigation menu and hamburger menu navigation display only visible navigation nodes. After this change, we will allow them to display also scheduled navigation nodes (with temporal visibility) between the start scheduling and end scheduling dates.


azayati (approved):
LGTM

PR #2530, opened 2023-05-11
feat: Make translation drawer component a common reusable component - EXO-63408 - Meeds-io/MIPs#51

IlhemEssaadi:
Currently the translation drawer is only usable by translation component field of settings page, we need to use it outside. So after this change, we will make it as common reusable component.


azayati (approved):
LGTM

PR #2531, opened 2023-05-12
feat: Attach an image to an activity MEED-2030- Meeds-io/MIPs#53

SaraBoutej:
fix attach image feedback

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
@@ -45,7 +45,7 @@
         </v-card-text>
         <attach-image-component 
           :maxFileSize="$root.maxFileSize"
-          @attachments-uploaded="canPostActivity = true" />
+          @attachments-uploading="canPostActivity = true" />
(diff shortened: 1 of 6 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
```diff
@@ -43,7 +43,9 @@
             attachment-enabled
             autofocus />
         </v-card-text>
-        <attach-image-component :maxFileSize="$root.maxFileSize" />
+        <attach-image-component 
+          :maxFileSize="$root.maxFileSize"
+          @attachments-uploaded="canPostActivity = true" />
```
boubaker: we may need to disable posting until the images are uploaded or deleted, @SaraBoutej WDYT ?

```suggestion
          @attachments-uploading="canPostActivity = false" />
          @attachments-uploaded="canPostActivity = true" />
```

SaraBoutej: @boubaker yes I changed that, until all file.progress === 100

boubaker (approved):

PR #2540, opened 2023-05-18
feat: Make translation drawer component a common reusable component - EXO-63408 - Meeds-io/MIPs#51

IlhemEssaadi:
Currently the translation drawer is only usable by translation component field of settings page, we need to use it outside. So after this change, we will make it as common reusable component.


webapp/portlet/src/main/webapp/vue-apps/common/initComponents.js
```diff
@@ -54,6 +55,7 @@ const components = {
   'unread-badge': UnreadBadge,
   'alert-notifications': Notifications,
   'ripple-hover-button': RippleHoverButton,
+  'translation-drawer': TranslationDrawer,
```
boubaker, line 58: It's already re-usable, but you just have to add a dependency to the module inside gatein-resources.xml like here:

https://github.com/Meeds-io/social/blob/651de4582bfd9acd6ea4526455b9b9188cf80c03/webapp/portlet/src/main/webapp/WEB-INF/gatein-resources.xml#L1400-L1402

boubaker (changes requested):

webapp/portlet/src/main/webapp/vue-apps/common/initComponents.js
azayati: yes @boubaker I see but in our case we need only to use translation-drawer component from our side not the whole translation-field component, can we make translation-drawer reusable component from commons ?

azayati: ok we will adapt our work by using translation-field, thanks @boubaker

PR #2551, opened 2023-05-23
fix: Fix technical nodes labels in space navigation - EXO-62470 - Meeds-io/MIPs#51

IlhemEssaadi:
Prior to this change, in the edit navigation management popup the space navigation nodes are displayed with their technical labels because they are not well found in the space resource bundle. After this change, we will retrieve them instead from all shared resource bundles.


azayati (approved):
LGTM

PR #2571, opened 2023-05-30
fix: Date spelling of posts should start with a capital letter - EXO-63512 - Meeds-io/meeds#952

akhanfir:
Prior to this change, when set plf language to german/Deutsch, go to space and check the spelling of dates below the posts, The fist letter of the date is in lower case. To fix this problem, when plf language is german/Deutsch add a class text-none to eliminate existing text transformations knowing that this change keeps uppercase applied to the first letter of the text. After this change, the text is well written with capital letters in crowdin yet it is changed by css .

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityHeadTime.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityHeadTime.vue
@@ -25,14 +25,14 @@
             v-if="isActivityEdited"
             :value="activity.updateDate"
             :short="isMobile"
-            :class="truncateText"
+            :class="[truncateText, textNone]"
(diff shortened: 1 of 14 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityHeadTime.vue
```diff
@@ -25,14 +25,14 @@
             v-if="isActivityEdited"
             :value="activity.updateDate"
             :short="isMobile"
-            :class="truncateText"
+            :class="[truncateText, textNone]"
```
ahamdi, line 28: couldn't we just remove the class **text-capitalize-first-letter** and rely on Crowdin texts ?

exo-swf:
Stale PR

PR #2589, opened 2023-06-05
fix: Notficiation administration : disabled notifs should be hidden to users - EXO-64634 - Meeds-io/meeds#951

akhanfir:
Prior to this change, when administrator disable any notification for any option and user displays his notification settings. To fix this problem, show only editable notifications even in the drawer to modify the notification parameters by adding a condition on channel Active when filtering the display options and hidden the group that has none of the notifications has their options activated by adding an init function that will check the existence of the group . in the same option this option will also be disabled even for groups.


webapp/portlet/src/main/webapp/vue-apps/user-setting-notifications/components/UserSettingNotificationPlugin.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div v-if="isEnabledNotifications.length">
```
ahamdi: could you return the length from the computed property as you did in the [UserSettingNotificationGroup.vue](https://github.com/Meeds-io/social/pull/2589/files#diff-d367d001c3f5d3130c3c097edeab782fffb2a8f832840ad33532bc7336dd310d) file

ahamdi (approved):

PR #2605, opened 2023-06-09
fix: Fix renamed space root node label issue - EXO-64817 - Meeds-io/MIPs#51

GouadriaHanen:
Before this fix, when renaming a space, the space user root node label is also renamed with the corresponding space label.
After this change, after renaming a spac, we will modify only the space user root node name and keep its label as is.


azayati (approved):
LGTM

azayati:
@GouadriaHanen add a description of the issue before merging

PR #2628, opened 2023-06-20
fix: Fix top bar navigation behavior - EXO-64657 - Meeds-io/MIPs#51

IlhemEssaadi:
After this change, we will:

- Introduce the possibility to open top bar navigation sub menus by hovering
- Keep the shadow effect on the whole node tree when elapsing it.
- Remove the hand cursor effect from "group" node type


azayati (approved):
LGTM

PR #2635, opened 2023-06-22
fix: Spaces' Members: Root profile displayed twice - EXO-64597 - Meeds-io/meeds#1000 

akhanfir:
Prior to this change, when admin creates space and opens space members app, this space members app displays admin profile twice. To fix this problem, filter before returning the list of space members by their id so as not to duplicate the profile. after this change,Root profile should be displayed only once space members App.


ahamdi (commented):
Do you know why we do have the profile of root returned twice ?

exo-swf:
Stale PR

PR #2657, opened 2023-07-11
fix: Fix space top bar navigation behavior - EXO-65369 - Meeds-io/MIPs#68

IlhemEssaadi:
Prior to this change, when a new application was added to a space top navigation bar then we access to the added application, a new empty tab is opened. This is due to the fact that the URL of the added application is treated as a link. 
After this change, we have implemented the isValidUrl method to verify that the url is a link.


azayati (changes requested):
add a clear description of the fix

PR #2670, opened 2023-07-19
Merge Meeds-io/MIPs#50

### LINKED ISSUE #50: Contributions UX & Navigation
## Rationale

After more than a year of using the contributions center, we need to review the way users access programs, actions and achievements.
Here are some concerns reported by users:
- from the overview page, almost anything is accessible with a drawer and if need to see more, it is possible. Not the case with programs
- programs and related actions have audience. The problem is that when visiting a space, we don't know which actions are proposed to this audience. We need to go see the contributions center to figure it out
- Challenges announcements become more and more noisy for users on the activity stream
- When seeing an announcement activity, users cannot access the action detail without accessing the contributions center
- Apps of the contribution center became more and more detailed, complete and we also have tabs/views for almost each app. Navigation needs to be better provided
- Actions triggered by meeds events are not up to date and needs to be reviewed

## 1. Functional Requirements

1. Actions announcement are comments of a main activity
- When creating an action, option to publish the activity on the related audience's stream (set to true by default)
- Ability to opt it off and then to update the action and choose to publish it
- Each announcement done for achieving the action will generate a comment of this activity
- If the action had'nt been pubished, then activity will be hidden. Activity and related comments will still be accessible from the direct URL
- Activity for announcement will be reviewed to get the program avatar instead of the avatar of the app
- When hovering the action summary, a popover will display to access the summary of it and when clicking to it, the action detail drawer opens without going to the contributions/actions page
- Notifications for action activity and for announcement comment will be added to distinguish activity notifications with those activities type (email, onsite and unread activities notifications). By default, those notifications will be added by default to any existing users

2. Access a program summary drawer from the overview page

3. Actions gadget to improve for overview
- Overview: list 4 items instead of 3
- Overview: merge action and challenge
- List action upon their status: ending soon + available actions
- Propose to have it in the stream page for meeds package (set to false for other)

4. Actions gadget decentralized by space if related to an audience and visible from the stream page as well
- Same gadget UI and UX used in overview page
- Add a gadget in the space when actions are proposed to be done for this audience with a sort as the following:
a. by default from the most rewarding one to the less (4 max) 
b. and if actions needs to be done step by step, propose these one first
c. if action to do once and done as agreed, then list other action
d. No see more as the space can be used as an audience for different programs
- Add this widget in the stream page to list all actions the user can do 

This gadget will be added automatically to any space if a program is related to it, and to the stream page if an action is listed as available, soon available for the user.

5. Contributions center navigation
- From the topbar, three pages are proposed in a submenu of the contributions item: Programs, Actions, Achievements
- Each item is a page
- Review the app from the appcenter to be "Contributions programs" so the user access the programs page
- Review the way programs and actions are listed and can be found: tags by program and actions can help the user find by "category", programs are listed in programs page and actions page alphabetically

6. Actions - Meeds events review
- Some event are not working while proposed: receive like on activity's comment, become space host
- Others are not consistent with what we are looking to trigger: update profile avatar is not ok (only when adding after default one), update profile banner is not ok (only when adding after default one), names for activities actions on own activity stream need to be reviewed
- Others need to be removed: log in, Receive activity (network -> redundant with another one) -> network and own activities triggers will be removed and only one trigger for each will be considered (no matter it is in a space stream or in a network stream). Transition of these changes will be done manually with UI
- one need to be removed from meeds package (to be moved to exo package?): upload document
- Few are missing to engage the user discovering more the platform and sharing more: fill profile informations, update a space as space host, pin an activity, poll creation/participation, bookmark usage, tags use, wallet initialization

### Security 

The widget must display a contextual information and permits to see only permitted challenges to announce. The endpoints has to be secured as well to not retrieve only allowed audience members to access challenges switch associated program target audience.

### Feature Flag

No feature flag will be needed to introduce this feature.

### Upgradability

The widget must be able to be plugged into Space Home page (same as Stream Page) without a migration existing pages. In fact, to do this, we will have to use the already existing dynamic containers to inject the widget.

Besides, the new notifications that will be introduced by this MIP has to be enabled by default to all existing users. Thus a new Upgrade plugin has to be added.

Reuse existing components instead of redefining them to preserve same UX as Oveview Page.

### Access 

The overview widget components has to be added in a reusable module so that we can reuse it. In fact, the reusable component has to be imported using load-group to optimize the number of JS files to load when loading the Space Home page with this widget.

### Services & processing

No change, we will have to reuse the same portlets of Overview page with contextual parameter of current space as audience to retrieve challenges and programs.

boubaker:


component/api/src/main/java/org/exoplatform/social/core/profile/ProfileLifeCycleEvent.java
```diff
@@ -30,7 +32,16 @@ public class ProfileLifeCycleEvent extends LifeCycleEvent<String, Profile> {
 
   public enum Type {ABOUT_ME, AVATAR_UPDATED, BASIC_UPDATED, CONTACT_UPDATED, EXPERIENCE_UPDATED, HEADER_UPDATED, CREATED, BANNER_UPDATED, TECHNICAL_UPDATED}
 
-  private Type type;
+  private Type   type;
+
+  @Getter
+  private String modifierUsername;
```
rdenarie, line 38: Can you explain why we need this new attribute ? I didnt manage to understand the purpose.
As I understand, a profile can be modified by another user. 
So user = the user with the modified profiler
modifierUsername = the user who modify it.

But, what is the functionnal scenario ?

boubaker: Knowing that the events triggering have to consider the case of **asynchronous** listeners execution (like using Event listening using a Redis Service in a Microservices Architecture), the access to `ConversationState` to get the current user isn't recommended. Thus to have the capability of asynchronous listeners execution, the `Payload` has to have the `modifierUsername` attribute.

As first use case of this new attribute, please take a look at https://github.com/Meeds-io/gamification/pull/1155/files#diff-f4e0e8f99a7eba6840085c1765bb5a2af639ed6d40ff58a321560008c578c87e

`public class GamificationProfileListener extends ProfileListenerPlugin`

rdenarie (approved):

PR #2702, opened 2023-08-03
fix: Remove the user  from the old group synchronized profile property groups - EXO-64814 - Meeds-io/meeds#1052

sofyenne:
Prior to this change, after editing a group-synchronized profile property, a new group was created, and the user continued to be displayed among the members of the old group
This change is going to remove the user from the old group synchronized profile property groups.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/listeners/GroupSynchronizationSocialProfileListener.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/listeners/GroupSynchronizationSocialProfileListener.java
@@ -20,6 +20,7 @@
 
 package org.exoplatform.social.core.listeners;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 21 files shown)


component/core/src/main/java/org/exoplatform/social/core/listeners/GroupSynchronizationSocialProfileListener.java
```diff
@@ -126,6 +127,7 @@ private void synchronizeProperty(Map.Entry<String, Object> property, Group profi
       Group newPropertyNameGroup = getOrCreateGroup(propertyName, profileGroup);
       try {
         Group newPropertyValueGroup = getOrCreateGroup(propertyValue, newPropertyNameGroup);
+        removeUserFromExistingPropertyGroup(newPropertyNameGroup, user);
```
ahamdi, line 130: aren't you removing then adding the user to the same group  here ?

ahamdi (changes requested):

component/core/src/main/java/org/exoplatform/social/core/listeners/GroupSynchronizationSocialProfileListener.java
sofyenne: We are removing the user from the existing child group, and then adding them to the newly created child group within the same larger (parent) group!

ahamdi (approved):

PR #2712, opened 2023-08-10
fix: Check user existence before updating his salt - EXO-65151 - Meeds-io/MIPs#69

rdenarie:
Prior to this change, in some particular cases, the picketlink identity is not found (due to old data). This lead to a null pointer exception for theses users. In order to finish the UP correctly, we add a condition to ignore theses users.


rdenarie:
@hakermi This modification was tested on community-dev, in order to fix 4 users which have the problem in tribe data

hakermi (approved):

PR #2714, opened 2023-08-10
feat: Add the ability to send a kudos from the composer - MEED-2071 - meeds-io/meeds#54

MayTekayaa:
This change will add the ability to send a kudos from the composer .

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
@@ -66,8 +66,8 @@
         </v-card-text>
         <v-card-text>
           <extension-registry-components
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
```diff
@@ -64,6 +64,11 @@
             </template>
           </extension-registry-components>
         </v-card-text>
+        <v-card-text>
+          <extension-registry-components
+            name="ActivityComposerKudos"
```
boubaker: Please avoid using `Kudos` term in Social
```suggestion
            name="..."
```

MayTekayaa: done!

boubaker (approved):

PR #2715, opened 2023-08-10
Merge mip 69

### LINKED ISSUE #69: [Technical Improvement] Update Password Encoder Algorithm
## Rationale
Our password encoder algorithm is outdated and not secure (SHA-256)
We need to update to a more secure one

In addition, the usage of RememberMe token is not secure.

## 1. Functional Requirements
No functionnal modification 

### Security 
We want to use the Argon2id algorithm to hash passwords with parameters 
·	int opsLimit = 2; // 2 iterations
·	int memLimit = 32768; // 32 MB
·	int parallelism = 1; // 1 parallel execution
·	int outputLength = 32; // password hash length
The password Salt has to be 128 bits (16 bytes) of length.

For RememberMe Token, we want to generate it with the Argon2id algorithm with same parameters as password

### Configurability 
As Argon2 parameters could not be changed without reseting all password, we not make it configurable

### Upgradability
The migration strategy have 3 options : 
- keep hash with old algo. When user logs in, update the password hash with the new algo
- Invalidate passwords, so thats users have to update password, which will be hashed with new algo. Not applicable in our case as we cannot "invalidate" password
- Update SHA-256 hashed passwords stored in database. Database stores as password hash oldAlgo(password). Upgrade plugin will update theses hashes to have newAlgo(oldAlgo(password)). So that, all password are hashed with the new algo. The login processus must be able to validate password in both options.

We will use the most secure strategy based on data upgrade : the solution 3

### Existing Features
No modification

### Feature Flags
No Feature flag

### Documentation 
Technical documentation to update

## 4. Software Architecture
Store password process : 
image

Store Password Sequence Diagramm
image

Only the red part is to create

Validate Password process
image

Validate Password pseudo code in AttributeManager.validatePassword (afeter data upgrade execution)
```
Input : username, password

If (not user have attribute passwordSalt128 && user have attribute passwordSalt) {
   // user not connected && not migrated by data upgrade
   Check password with old algo (oldAlgo(password) == hashInDatabase ?)
   If validation is ok, call updatePassword to compute and save a hash with the new algo. During passwordUpdate, we need to ensure to remove passwordSalt attribute

} else if(user have attribute passwordSalt128 && user have attribute passwordSalt) (
   // user not connected && migrated by data upgrade
   Hash = Compute the hash of the password by using the old passwordEncoder  DatabaseReadingSaltEncoder. (oldAlgo(password))
   Then check hash with newAlgo (newAlgo(oldAlgo(password)) == hashInDatabase ?)
   If validation is ok, call updatePassword to compute and save a hash with the new algo. During passwordUpdate, we need to ensure to remove passwordSalt attribute

} else {
  //user connected
  use the current way to check password it will get the default configured PasswordEncoder and use it
      return this.getRepository().validateCredential(this.getInvocationContext(), this.createIdentityObject(user), new PasswordCredential(hash, this.getCredentialEncoder(), user.getKey()));
}

```

Remember me generation : 
image

RememberMe Validation
image

Once remember me token are generated with new algo, we need an AuthenticatorPlugin to authenticate the user.

### Access 
   - GUI 
   - API 

### Services & processing
   - Ex: Scheduled Jobs

### Data and persistence
   - DataUpgrade 

```
for all user without attribute “passwordSalt128” :
   Get password hash in database
   Generate a passwordSalt128 and store it as attribute for the user
   Compute newAlgo(hash) and store it as new hash for the user
```
## 5. Annexes
   - Reasoning that explain the technical choices
   - Impacts 
   - References

rdenarie:

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
@@ -93,7 +93,7 @@ public class UserPasswordHashMigration extends UpgradeProductPlugin {
         try {
           String saltString = Hex.encodeHexString(generateRandomSalt());
           User user = picketLinkIDMService.getIdentitySession().getPersistenceManager().findUser(userName);
-          if (picketLinkIDMService.getExtendedAttributeManager().getAttribute(userName, PASSWORD_SALT_USER_ATTRIBUTE) == null) {
+          if (user != null && picketLinkIDMService.getExtendedAttributeManager().getAttribute(userName, PASSWORD_SALT_USER_ATTRIBUTE) == null) {
             picketLinkIDMService.getExtendedAttributeManager().addAttribute(userName, PASSWORD_SALT_USER_ATTRIBUTE, saltString);
             picketLinkIDMService.getExtendedAttributeManager().updatePassword(user, passwordHash);
           }
--- a/component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
@@ -1,0 +1,157 @@
+/*
+ * This file is part of the Meeds project (https://meeds.io/).
+ * Copyright (C) 2023 Meeds Association
+ * contact@meeds.io
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.exoplatform.social.core.upgrade;
+
+import org.apache.commons.codec.binary.Hex;
+import org.exoplatform.commons.persistence.impl.EntityManagerService;
+import org.exoplatform.commons.upgrade.UpgradePluginExecutionContext;
+import org.exoplatform.commons.upgrade.UpgradeProductPlugin;
+import org.exoplatform.container.PortalContainer;
+import org.exoplatform.container.component.RequestLifeCycle;
+import org.exoplatform.container.xml.InitParams;
+import org.exoplatform.services.log.ExoLogger;
+import org.exoplatform.services.log.Log;
+import org.exoplatform.services.organization.idm.PicketLinkIDMService;
+import org.exoplatform.web.security.security.SecureRandomService;
+import org.picketlink.idm.api.User;
+
+import javax.persistence.EntityManager;
+import javax.persistence.Query;
+import java.security.SecureRandom;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class UserPasswordHashMigration extends UpgradeProductPlugin {
+
+  private static final Log           LOG                          = ExoLogger.getExoLogger(UserPasswordHashMigration.class);
+
+  private static final int           CHUNCK                       = 100;
+
+  private final EntityManagerService entityManagerService;
+
+  private final PicketLinkIDMService picketLinkIDMService;
+
+  private final SecureRandomService  secureRandomService;
+
+  private static final String        PASSWORD_SALT_USER_ATTRIBUTE = "passwordSalt128";
+
+  private static final String        DEFAULT_ENCODER              = "org.exoplatform.web.security.hash.Argon2IdPasswordEncoder";
+
+  public UserPasswordHashMigration(EntityManagerService entityManagerService,
+                                   PicketLinkIDMService picketLinkIDMService,
+                                   SecureRandomService secureRandomService,
+                                   InitParams initParams) {
+    super(initParams);
+    this.entityManagerService = entityManagerService;
+    this.picketLinkIDMService = picketLinkIDMService;
+    this.secureRandomService = secureRandomService;
+  }
+
+  @Override
+  public void processUpgrade(String s, String s1) {
+    LOG.info("Start upgrade of users passwords hashing algorithm");
+    long startupTime = System.currentTimeMillis();
+
+    PortalContainer container = PortalContainer.getInstance();
+    AtomicInteger updatedPasswords = new AtomicInteger();
+    EntityManager entityManager = this.entityManagerService.getEntityManager();
+    int totalSize;
+    try {
+      String sqlString = "SELECT jbid_io.NAME, jbid_io_creden.TEXT  FROM"
+          + " (jbid_io_creden INNER JOIN jbid_io ON jbid_io_creden.IDENTITY_OBJECT_ID = jbid_io.ID)"
+          + " INNER JOIN (SELECT jbid_io_attr.IDENTITY_OBJECT_ID FROM jbid_io_attr"
+          + " WHERE jbid_io_attr.IDENTITY_OBJECT_ID NOT IN (SELECT jbid_io_attr.IDENTITY_OBJECT_ID FROM jbid_io_attr WHERE NAME = 'passwordSalt128')"
+          + " GROUP BY jbid_io_attr.IDENTITY_OBJECT_ID) jia ON jbid_io_creden.IDENTITY_OBJECT_ID  = jia.IDENTITY_OBJECT_ID;";
+
+      RequestLifeCycle.begin(container);
+      Query nativeQuery = entityManager.createNativeQuery(sqlString);
+      List<Object[]> result = nativeQuery.getResultList();
+      AtomicLong chunckStartTime = new AtomicLong(System.currentTimeMillis());
+      updatedPasswords.set(0);
+      List<String> userWithErrors = new ArrayList<>();
+      totalSize = result.size();
+      result.forEach(item -> {
+        String userName = (String) item[0];
+        String passwordHash = (String) item[1];
+
+        try {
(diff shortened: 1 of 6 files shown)


component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
```diff
@@ -0,0 +1,157 @@
...
+    try {
+      String sqlString = "SELECT jbid_io.NAME, jbid_io_creden.TEXT  FROM"
+          + " (jbid_io_creden INNER JOIN jbid_io ON jbid_io_creden.IDENTITY_OBJECT_ID = jbid_io.ID)"
+          + " INNER JOIN (SELECT jbid_io_attr.IDENTITY_OBJECT_ID FROM jbid_io_attr"
+          + " WHERE jbid_io_attr.IDENTITY_OBJECT_ID NOT IN (SELECT jbid_io_attr.IDENTITY_OBJECT_ID FROM jbid_io_attr WHERE NAME = 'passwordSalt128')"
+          + " GROUP BY jbid_io_attr.IDENTITY_OBJECT_ID) jia ON jbid_io_creden.IDENTITY_OBJECT_ID  = jia.IDENTITY_OBJECT_ID;";
+
+      RequestLifeCycle.begin(container);
+      Query nativeQuery = entityManager.createNativeQuery(sqlString);
+      List<Object[]> result = nativeQuery.getResultList();
```
boubaker, line 77: Test on result emptiness and `break` the method execution right away when empty.

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
```diff
@@ -0,0 +1,157 @@
...
+      AtomicLong chunckStartTime = new AtomicLong(System.currentTimeMillis());
```
boubaker: Not mandatory to use an `AtomicLong` since it's not in a parallel execution. Please add the loop execution in a dedicated method where you will not need this at all (should be handled inside the method itself)
```suggestion
```

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
```diff
@@ -0,0 +1,157 @@
...
+      updatedPasswords.set(0);
```
boubaker: Add an inner class to aggregate the final report into one single object (you can add usernames as well if needed in displayed report)

```suggestion
  public static class UpgradeReport {

    private long startTime = System.currentTimeMillis();

    @Getter
    private long total;

    @Getter
    private long success;

    @Getter
    private long failure;

   public UpgradeReport(long total) {
     this.total = total;
   }

    public void incrementSuccess() {
      success++;
    }

    public void incrementFailure() {
      failure++;
    }

   public long getDurationInMillis() {
     return System.currentTimeMillis() - startTime;
   }
  }
```

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
```diff
@@ -0,0 +1,157 @@
...
+            LOG.info("{}/{} passwords have been updated ({} ms for the chunck)",
+                     current,
+                     totalSize,
+                     System.currentTimeMillis() - chunckStartTime.get());
+            chunckStartTime.set(System.currentTimeMillis());
+          }
+        } catch (Exception e) {
+          userWithErrors.add(userName);
+          LOG.error("Error while creating attribute salt and updating password hash for user : {}", userName, e);
+        }
```
boubaker: Move this into a dedicated method with `@ExoTransactional` to ensure committing every change apart (`jcabi-plugin` is already in `pom.xml`, so the annotation should work).

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
```diff
@@ -0,0 +1,157 @@
...
+               totalSize - userWithErrors.size(),
+               userWithErrors.size(),
+               (System.currentTimeMillis() - startupTime));
+      if (!userWithErrors.isEmpty()) {
+        LOG.error("Users with not modified password : {}", userWithErrors);
+        throw new IllegalStateException("UserPasswordHashMigration upgrade failed due to previous errors");
+      }
+    } catch (Exception e) {
+      LOG.error("Error while getting old users passwords hash", e);
+      throw new IllegalStateException("UserPasswordHashMigration upgrade failed due to previous errors");
```
boubaker: either log or throw exceptions, never both

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
```diff
@@ -0,0 +1,157 @@
...
+                                 .getName()
+                                 .equals(DEFAULT_ENCODER)
+          && super.shouldProceedToUpgrade(newVersion, previousGroupVersion, previousUpgradePluginExecution);
+    } catch (Exception e) {
+      LOG.error("Error while checking current default credential encoder", e);
+      return false;
+    }
+  }
+
+  private byte[] generateRandomSalt() {
```
boubaker, line 146: Can you please reuse this method from a centralized manner to ensure to be compliant with configured algorithm in PLIDM.

extension/war/src/main/webapp/WEB-INF/conf/social-extension/portal/upgrade-plugins-configuration.xml
```diff
@@ -0,0 +1,60 @@
...
+                <value-param>
+                    <name>plugin.upgrade.execute.once</name>
+                    <description>The plugin must be executed only once</description>
+                    <value>true</value>
+                </value-param>
+                <value-param>
+                    <name>plugin.upgrade.async.execution</name>
+                    <description>The plugin will be executed in an asynchronous mode</description>
+                    <value>true</value>
+                </value-param>
```
boubaker, line 56: It shouldn't be asynchronous IMO to avoid starting the server while not accessible to some users.
```suggestion
```

extension/war/src/main/webapp/WEB-INF/conf/social-extension/portal/upgrade-plugins-configuration.xml
```diff
@@ -0,0 +1,60 @@
...
+                <value-param>
+                    <name>product.group.id</name>
+                    <description>The groupId of the product</description>
+                    <value>org.exoplatform.platform</value>
+                </value-param>
+                <value-param>
+                    <name>plugin.upgrade.target.version</name>
+                    <description>
+                        The plugin target version (will not be executed if previous version is equal or higher than
+                        6.3.1)
```
boubaker:
```suggestion
                        The plugin target version (will not be executed if previous version is higher than 6.5.0)
```

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
rdenarie: The almost next instruction is result.forEach
So if result is empty, we will not enter in the forEach and the UP will successfully finish his execution. 

The break will make the code less readable with not needed lines.

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
rdenarie: Done.

extension/war/src/main/webapp/WEB-INF/conf/social-extension/portal/upgrade-plugins-configuration.xml
rdenarie: We allow a user with a non upgraded hash to log by using oldClassEncoder
In ExtendedAttributeManager.validatePassword, the code check if the hash was updated with the new algo.
If not, the old algo is still used.

On tribe, the up use 30hours to execute, so, it is not possible to make it sync.

extension/war/src/main/webapp/WEB-INF/conf/social-extension/portal/upgrade-plugins-configuration.xml
rdenarie: done

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
rdenarie: Done

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
rdenarie: done

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
rdenarie: done

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
rdenarie: done

extension/war/src/main/webapp/WEB-INF/conf/social-extension/portal/upgrade-plugins-configuration.xml
boubaker: ok, I see, thanks for explanation.

component/core/src/main/java/org/exoplatform/social/core/upgrade/UserPasswordHashMigration.java
boubaker: > The break will make the code less readable with not needed lines.

In contrary IMO, in fact, the code will be clear that without results, the execution is stopped right away.
In addition, if we don't stop, we will have to execute more instructions and create a useless object (like `new UpgradeReport`) and logging an information `End upgrade of users passwords hashing algorithm` when starting a fresh server.

boubaker: I think this is minor anyway, so let's keep it as it is. Thanks for your answer.

boubaker (dismissed):

boubaker (dismissed):

boubaker (dismissed):

boubaker (approved):

PR #2717, opened 2023-08-14
feat: Edit an image - MEED-2043 - Meeds-io/MIPs#53

SaraBoutej:
fix feedback


component/core/src/main/java/org/exoplatform/social/core/attachment/AttachmentServiceImpl.java
```diff
@@ -376,6 +376,11 @@ private void saveAttachment(UploadedAttachmentDetail uploadedAttachmentDetail,
       } else {
         updateAttachment(fileId, objectType, objectId,userIdentityId, altText);
       }
+      broadcastAttachmentChange(ATTACHMENT_CREATED_EVENT,
```
boubaker: Created VS Updated, please trigger both switch case.

SaraBoutej: done

boubaker (approved):

PR #2719, opened 2023-08-15
fix: Issue with cursor icon when hovering over a notification text -EXO-63933 - Meeds-io/meeds#1064.

akhanfir:
Prior to this change, when open the edition drawer of any notification and hover over the text of the one of the 3 options , the cursor icon changes to hand icon but no action can be done by clicking on the text After this change, the cursor should remain with the text icon.


ahamdi (commented):
Clicking on the label does not do anything because the label tag is not attached to the input ID. Could you check this instead of removing the label ?

ahamdi (approved):

PR #2727, opened 2023-08-17
fix: Chat icon should only be displayed for space members - EXO-64806 - Meeds-io/meeds#1069.

akhanfir:
Prior to this change, when create article in spaceX and publish it on target, create program whose audience spaceX, share activity from spaceX to spaceY and create user1 simple user & user2 admin on plf/ spaces manager not members in spaceX open this article from target and hover on spaceX link, chat icon is visible for none members and opens the drawer. To fix this problem, added to the check if this room enebled another check if this user current is a member of this space in the popverChatButton and to ensure that the chat button only appears for members of this space add one another check during their display in the popver if this user is a member in this space. After this change,chat icon is displayed for spaceX members only.

### DIFF
--- a/component/api/pom.xml
+++ b/component/api/pom.xml
@@ -23,7 +23,7 @@
   <parent>
     <artifactId>social-component</artifactId>
     <groupId>org.exoplatform.social</groupId>
-    <version>6.5.x-exo-SNAPSHOT</version>
+    <version>6.5.x-maintenance-SNAPSHOT</version>
   </parent>
   <artifactId>social-component-api</artifactId>
   <name>eXo PLF:: Social API</name>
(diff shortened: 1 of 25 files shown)


webapp/portlet/src/main/webapp/vue-apps/popover/components/SpacePopoverContent.vue
```diff
@@ -55,6 +55,7 @@
         :is-favorite="space.isFavorite"
         :space-id="space.id" />
       <extension-registry-components
+        v-if="space.isMember"
```
ahamdi: This will disable all extension components for non members, I prefer letting it to the logic of the extension component to display or not the extension, WDYT ?

ahamdi (approved):

PR #2729, opened 2023-08-17
feat: Hide global site navigation nodes - EXO-65470 - Meeds-io/MIPs#72

GouadriaHanen:


azayati (changes requested):
In this US we should just hide the global site navigation nodes, the includeGlobal param can be added when adjusting the metasite first level with the FT flag

PR #2747, opened 2023-08-23
feat: Implement left navigation feature - Meeds-io/MIPs#72

### LINKED ISSUE #72: Left navigation
## Rationale

No functional impact for Meeds

This MIP introduces the capabilities to display sites in the left menu, the navigation nodes and allows an administrator to manage the icons associated with each item. Only by configuration (XML)

No big impact on data volume: Only four properties to be added to the PORTAL_SITES table.
Performance when creating or getting navigations and sites should not be affected.

### Security 

All users should access to the meta site and non meta sites with their navigation nodes. 

### Extensibility 

The meta and no meta sites should be created by xml configuration files with the new added properties.

### Upgradability

New properties added to the models should be configured for the existing navigation nodes in the first server start.

### Feature Flags

| Property Name  | Default Value | Target Audience | Functional Behaviour |
| ------------- | ------------- | ------------- | ------------- |
|exo.feature.newLeftNavigationDrawer.enabled  |  false |users   |enable/disable the new left navigation drawer/the old left navigation drawer   |

This feature flag will be removed at the end of the iteration.

### Access

- GUI 
   
We need to update the frontend of HamburgerMenu portlet in order to meet the new look and feel of the left navigation menu.

### Services & processing

SitesRest service will be added with the method:

- **GET: /v1/social/sites (getSites)**: This service will need **expandNavigations** and **displayed** query params in order to retrieve the list of displayed sites with their navigations. 

This service will be moved from layout-management eXo module to social Meeds module.

### Data and persistence
  

- [ ] Four new fields should be added to the PORTAL_SITES table:

(BOOLEAN) DISPLAYED: A boolean which indicates if the site will be displayed in the left navigation of the **meta** site
(INT) DISPLAY_ORDER: An integer which indicates the display order of the site in the left navigation of the **meta** site
(BIGINT) LOGO: A big integer which indicates the file id in the FILES_FILES table corresponding to the site logo
(BIGINT) BANNER_FILE_ID: A big integer which indicates the file id in the FILES_FILES table corresponding to the site banner

- Related JPA Entities (Site and Node) should be updated to add these new attributes in addition to their different layers.
- We should update **NewPortalConfig** to add these new attributes which will be defined into the corresponding object-param of **NewPortalConfigListener** component plugin configuration files.
- We should have **only one meta site**, this will be configured using [default.portal value-param of **NewPortalConfigListener** component plugin configuration files](https://github.com/exoplatform/digital-workplace/blob/develop/digital-workplace-webapps/src/main/webapp/WEB-INF/conf/digital-workplace/portal-configuration.xml#L17).
- We should store the site logo and site banner using file service.

azayati:


component/api/.gitignore
```diff
@@ -0,0 +1 @@
+/bin/
```
boubaker: To delete since we already have a file `.gitignore` in the root folder

component/core/.gitignore
```diff
@@ -0,0 +1 @@
+/bin/
```
boubaker: To delete since we already have a file `.gitignore` in the root folder

component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
```diff
@@ -89,6 +89,10 @@ public class LinkProvider {
 
   public static final String TYPE = "file";
 
+  public static final String     BASE_URL_SITE_REST_API   = "/v1/social/sites";
```
boubaker: Code format
```suggestion
  public static final String BASE_URL_SITE_REST_API   = "/v1/social/sites";
```

component/notification/.gitignore
```diff
@@ -0,0 +1 @@
+/bin/
```
boubaker: To delete since we already have a file `.gitignore` in the root folder

component/service/.gitignore
```diff
@@ -0,0 +1 @@
+/bin/
```
boubaker: To delete since we already have a file `.gitignore` in the root folder

component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
```diff
@@ -145,6 +211,14 @@ public class EntityBuilder {
   public static final CacheControl        NO_CACHE_CC                                = new CacheControl();
 
   private static final JsonEntityProvider JSON_ENTITY_PROVIDER                       = new JsonEntityProvider();
+  
+  public static final String GROUP = "group";
```
boubaker: code format + elements order, public, friendly, protected and then private.

component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
```diff
@@ -1688,4 +1764,222 @@ public static RelationshipManager getRelationshipManager() {
...
+      if (navigation != null) {
+        UserNodeFilterConfig.Builder builder = UserNodeFilterConfig.builder();
+        builder.withReadWriteCheck().withVisibility(convertVisibilities(visibilityNames));
+        if (temporalCheck) {
+          builder.withTemporalCheck();
+        }
+        rootNode = userPortal.getNode(navigation, Scope.ALL, builder.build(), null);
+      }
+    } catch (Exception e) {
+      LOG.error("Error while getting site {} navigations for user {}", site.getName(), currentUser, e);
```
boubaker: Why not throwing the error, knowing that it's blocker to build a consistent information to return to the user

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
```diff
@@ -0,0 +1,290 @@
...
+      siteFilter.setExcludeSpaceSites(excludeSpaceSites);
+      siteFilter.setSortByDisplayOrder(sortByDisplayOrder);
+      if (filterByDisplayed) {
+        siteFilter.setFilterByDisplayed(filterByDisplayed);
+        siteFilter.setDisplayed(displayed);
+      }
+      siteFilter.setLimit(limit);
+      siteFilter.setOffset(offset);
+      List<PortalConfig> sites = layoutService.getSites(siteFilter);
+      return Response.ok(EntityBuilder.buildSiteEntities(sites,
```
boubaker: Shouldn't we use `ETag` to avoid computing the whole response on each page display ? (this way, we send a simple 304 to improve response content download bandwidth + JVM processing)

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
```diff
@@ -0,0 +1,290 @@
...
+                                @QueryParam("isDefault")
+                                boolean isDefault) {
+    try {
+      if (!isDefault && RestUtils.isAnonymous()
+          && !LinkProvider.isSiteBannerTokenValid(token, siteName, LinkProvider.ATTACHMENT_BANNER_TYPE)) {
+        LOG.warn("An anonymous user attempts to access banner of site {} without a valid access token", siteName);
+        return Response.status(Response.Status.FORBIDDEN).build();
+      }
+      InputStream stream = isDefault ? getDefaultBannerInputStream(siteName) : layoutService.getSiteBannerStream(siteName);
+      return Response.ok(stream, "image/png").build();
```
boubaker: Same comment as previous for `ETag` usage.

webapp/portlet/src/main/webapp/images/sites/banner/mycraft.png
boubaker, line 1: This seems not used here, can you delete it please

webapp/portlet/src/main/webapp/images/sites/banner/intranet.png
boubaker, line 1: This seems not used here, can you delete it please

webapp/portlet/src/main/webapp/images/sites/banner/engage.png
boubaker, line 1: This seems not used here, can you delete it please

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
```diff
@@ -0,0 +1,290 @@
...
+public class SiteRest implements ResourceContainer {
+
+  private static final Log    LOG                     = ExoLogger.getLogger(SiteRest.class);
+
+  private LayoutService       layoutService;
+
+  private PortalContainer     portalContainer;
+
+  private static final String SITE_DEFAULT_BANNER_URL = System.getProperty("sites.defaultBannerPath",
+                                                                           "/images/sites/banner/defaultSiteBanner.png");
```
boubaker: Please make this parameterable using regular means (in Service layer using init params with default value in XML instead of hard coding it here)

webapp/portlet/src/main/webapp/vue-apps/common/main.js
```diff
@@ -19,6 +19,7 @@ import * as brandingService from './js/brandingService.js';
 import * as navigationService from '../common/js/NavigationService.js';
 import * as profileSettingsService from '../common/js/ProfileSettingsService.js';
 import * as profileLabelService from '../common/js/ProfileLabelService.js';
+import * as siteService from './js/SiteService.js';
```
boubaker: May be not used a lot as a common service. Can you move it in Hamburger menu please.

component/api/.gitignore
azayati: removed, you are right it was a mistake, thanks

component/core/.gitignore
azayati: removed, you are right it was a mistake, thanks

component/notification/.gitignore
azayati: removed, you are right it was a mistake, thanks

component/service/.gitignore
azayati: removed, you are right it was a mistake, thanks

component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
azayati: fixed

component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
azayati: IMO it is not blocker, because even if when we catch the exception and just log it without throwing, the site Navigations will be null and not returned. In that case, the site will be returned just without the navigations which is not an optional information.

boubaker: Yep, but to diagnose and be aware that the REST API had errors, we will have to ask support or ITOP teams to check on it, while the REST API should be sel explanatory (see REST Best practices) ...

azayati: ok we throw an ObjectNotFoundException in that case to the Rest ?

boubaker: Nope, I think it's an unkown error, thus an HTTP 500 with `e.getMessage()` as entity + log it as WARN and not error since it's not a general Server status Error.

webapp/portlet/src/main/webapp/images/sites/banner/mycraft.png
azayati: removed

webapp/portlet/src/main/webapp/images/sites/banner/intranet.png
azayati: removed

webapp/portlet/src/main/webapp/images/sites/banner/engage.png
azayati: removed

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
azayati: fixed

webapp/portlet/src/main/webapp/vue-apps/common/main.js
azayati: fixed

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
azayati: Here we can't implement the Etag because we can not detect changes on all displayed informations. For example, navigation nodes and translated site label/description are calculated with the EntityBuilder and not with layoutService.getSites, so when we add a new navigation node or modify the translation of site label or site description, these changes can not be detected and then the old values still be displayed from the cache until cleaning the cache browser.

boubaker: The requirement is to have a cached response on such huge and repetitive REST call. We always have technical means to make it work, you can see it on Activity Stream by example.

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
azayati: updated

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
```diff
@@ -0,0 +1,319 @@
...
+                              @Parameter(description = "to exclude sites with empty navigation")
+                              @DefaultValue("false")
+                              @QueryParam("excludeEmptyNavigationSites")
+                              boolean excludeEmptyNavigationSites,
+                              @Parameter(description = "Used to retrieve the site label and description in the requested language")
+                              @QueryParam("lang")
+                              String lang) {
+    try {
+      PortalConfig site = layoutService.getPortalConfig(Long.parseLong(siteId));
+      return Response.ok(EntityBuilder.buildSiteEntity(site,
```
boubaker: Same here for the `304` HTTP response code ?

webapp/portlet/src/main/webapp/WEB-INF/jsp/portlet/bannerLogo.jsp
```diff
@@ -51,7 +56,21 @@
...
     portalPath = portalConfigService.getUserHomePage(request.getRemoteUser());
     if (portalPath == null) {
-      portalPath = defaultHomePath;
+      SiteFilter siteFilter = new SiteFilter();
+      siteFilter.setExcludedSiteName(portalConfigService.getGlobalPortal());
+      siteFilter.setSiteType(SiteType.PORTAL);
+      siteFilter.setSortByDisplayOrder(true);
+      siteFilter.setFilterByDisplayed(true);
+      siteFilter.setDisplayed(true);
+      List<PortalConfig> portalConfigList = portalConfigService.getSites(siteFilter);
```
boubaker: If this is a filter that is used here and in `DefaultRequestHandler`, shouldn't we add a dedicated method for it with the adequate business logic ?

```suggestion
      List<PortalConfig> portalConfigList = portalConfigService.getDisplayedPotalConfigs();
```

webapp/portlet/src/main/webapp/WEB-INF/jsp/portlet/bannerLogo.jsp
```diff
@@ -51,7 +56,21 @@
...
+      if (portalConfigList != null && !portalConfigList.isEmpty()) {
+        String defaultPortal = portalConfigList.get(0).getName();
+        Collection<UserNode> userNodes = portalConfigService.getSiteNavigations(defaultPortal, authenticatedUser, requestContext.getRequest());
+        UserNode userNode = portalConfigService.getFirstAllowedPageNode(userNodes);
+        portalPath = portalConfigService.getDefaultUri(userNode, defaultPortal);
+      } else {
+        portalPath = defaultHomePath;
+      }
```
boubaker: Shouldn't we add this logic to compute the default `portalPath` from Service layer ?

component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
```diff
@@ -588,4 +592,62 @@ public static String getResourceBundleLabel(Locale locale, String label) {
...
+        token = StringUtils.EMPTY;
+      }
+    }
+    boolean isDefault = bannerFileId == 0 ;
+    return new StringBuilder(getBaseURLSiteRest()).append("/")
+                                                  .append(siteName)
+                                                  .append("/")
+                                                  .append(ATTACHMENT_BANNER_TYPE)
+                                                  .append("?r=")
+                                                  .append(token)
```
boubaker: do we need Token generation for Anonymous Access, from Email notifications by example?

```suggestion
```

webapp/portlet/src/main/webapp/WEB-INF/jsp/portlet/bannerLogo.jsp
azayati: fixed

webapp/portlet/src/main/webapp/WEB-INF/jsp/portlet/bannerLogo.jsp
azayati: fixed

component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
azayati: removed

boubaker (approved):

PR #2748, opened 2023-08-23
feat: Implement "Site management feature" - Meeds-io/MIPs#68 

azayati:


webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/administration/AdministrationNavigations.vue
```diff
@@ -127,6 +129,8 @@ export default {
     },
   },
   created() {
+    this.$featureService.isFeatureEnabled('newSiteManagement')
+      .then(enabled => this.newSiteManagementEnabled = enabled);
```
boubaker: Add this feature flag in a gtmpl injected as made here:
https://github.com/Meeds-io/social/blob/develop/webapp/portlet/src/main/webapp/groovy/social/webui/UISocialPortalApplicationHead.gtmpl#L63

```suggestion
```

This will avoid having to a an additional HTTP Request for Meeds while it's not used in fact.
Besides, you can even extend Vue components defined in default packages using extensionRegistry to redefine one or multiple components when an addon is installed, see it here:
https://github.com/Meeds-io/social/blob/develop/webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/main.js#L23-L27

BTW, `newSiteManagementEnabled` will not remain `new` all time, so please use more appropriate name.

boubaker (approved):

PR #2772, opened 2023-08-30
 feat: Prevent the default event when clicking to delete an attached image - MEED-2036 - Meeds-io/MIPs#53

SaraBoutej:
Prior to this change when clicking to delete an attached image, the click event is propagated. This change will avoid to propagate this event.


boubaker:
please review the title and description to describe better what this change do

boubaker (approved):

boubaker (approved):

PR #2776, opened 2023-08-30
fix: [People search] no results displays if field has special char - EXO-65931 - Meeds-io/meeds#1083

akhanfir:
Prior to this change, when create user then Connect with different user and in people directory use search filter with a term, no profile displayed, special chars in profile aren't allowing to display the good users.To fix this problem, ignored in the query of the elastic search the special char.After this change, users which has the search filter values used are displayed.


ahamdi:
@akhanfir Could you resolve conflicts ?

rdenarie (changes requested):
can you try to create unit tests to cover added lines ?

exo-swf:

ahamdi:
@akhanfir could you resolve the conflicts ?

PR #2777, opened 2023-08-30
Merge Meeds-io/MIPs#63

### LINKED ISSUE #63: Hub Access
## Rationale

It is possible to configure a Meeds server as _Opened_ so anyone can sign up. Alternatively, it is also possible to set a Meeds server as _Restricted_ so only known users can sign in. Besides, it's also possible to configure how self registered users will be considered (members or external).

Currently, these settings can only be set by updating platform properties on the server host. As a hub admin, you can't switch these settings via UI. While at meeds, we aim to give capabilities for any Hub Host to manage and administrate their hub on their own as a self service software.

## 1. Functional Requirements

From the administration menu, General/Access settings, 
admin display access settings to precise how users will be able to login into the platform
1. First option:  
* Platform is set as _Open_ by default: anyone can self-register and anyone is considered a full "member" (/platform/user) even when invited from a space.
* -> Ability for the admin to authorize external user when opened. That means that self-registrated users and ones invited by space host will be considered external
* Email signup/signin options is proposed by default and cannot be disabled by UI. Indeed, the authentication topic will be elaborated in another MIP, including metamask, oAuth 

image

2. Second option: 
* Platform is set as _Restricted_  by default: only pre-registered user can access the platform and any user invited from a space is considered external (/platform/external)
* Only email sign in is possible (no sign up)

image

In addition to this, admin will find an easy access to create users from that page so the path to create users is easy and quickly understandable.

And finally, to ease the onboarding of any user accessing the platform the first time, the admin would get an option to set spaces as default for platform/users or platform/externals

> Precision: 
- when switching from a type to another, no automatic update of users. This only has an impact on login access to hub. 
- external access permissions from spaces management is removed as it is here provided in this administration page
- Same for default spaces: no upgrade will be done to previous registered users. This only aims new users
- opened will be the default choice for meeds package. If needed, this choice can be set differently with a property

### Extensibility 

No extensible area nor service is needed for now.

### Configurability 

| Parameter | Default value | Description |
| ------ | ------ | ------ |
| `meeds.settings.access.type.default` | `open` | Possible values : `open` or `restricted`. This will allow to choose the default setting of package. For Meed the access type will be put by default to `open`  | 
| `meeds.settings.access.externalUsers` | `false` | Possible values : `false` or `true`. This will allow to choose the default setting of package. For Meed the registered users will be put by default to `internal`  | 

### Upgradability

No upgrade is needed.

### Existing Features

The configuration, enablement of `Oauth` provider for login shouldn't be impacted by this change. It should remain exactly with the same property settings to be enabled.

### Feature Flags

No Feature Flag is needed.

### Access 

The same Portlet `General Settings` will be used an refactored to allow the display of three Tabs.
A REST endpoint for Hub Access configuration have to be made in order to allow read & write the Hub access settings. This REST endpoint has to be accessible for administrators only (`/platform/administrators`)

### Data and persistence

Since the settings are stored globally for the whole platform. One single entity instance will be needed, thus no need to add a dedicated Storage & DAO layers. The `SettingService` should be used instead to store the Hub Access settings.

boubaker:


component/oauth-auth/src/main/java/io/meeds/oauth/service/impl/OAuthRegistrationServiceImpl.java
```diff
@@ -92,7 +98,8 @@ public OAuthRegistrationServiceImpl(PortalContainer container,
 
   @Override
   public boolean isRegistrationOnFly(OAuthProviderType<? extends AccessTokenContext> oauthProviderType) {
-    return registerOnFly.contains(oauthProviderType.getKey());
```
rdenarie: In customer contexts, we need registration on the fly even if registrationType is restricted : 

Functionnal need : 
exo is configured with openid or oauth, but not connected to the AD/ldap. Users are not known in the platform until they log the first time.
In this case, the registration type is restricted AND registerOnTheFly contains OPENID.

With this modification, the user cannot log in and create an account, but is it a valid behaviour

rdenarie (changes requested):

component/oauth-auth/src/main/java/io/meeds/oauth/service/impl/OAuthRegistrationServiceImpl.java
boubaker: ok, this was discussed in a Task as functional behavior need. @srenault-meeds , we may missed a special use case, where it's restricted but Open on some Registration providers. WDYT ?

srenault-meeds: Hi, can we discuss it please @rdenarie? 
Indeed, I need to understand the case to make sure we adapt or not.

srenault-meeds: After discussion with both parties, here is what we will do:
- enable registeronthefly option even when the platform is restricted
- no ability to change this by UI, only by configuration
- so signup by email won't be possible but with oAuth provider enabled and configured with proper clientsecret
Thanks for your understanding

boubaker: ok done here 119557e9ab18e7af7e0fad2054b2c44d554c065e
Thanks for both.

rdenarie (approved):

PR #2787, opened 2023-09-01
Fix : Update Profile when synchronized profile property setting from AD - EXO-65023 - Meeds-io/meeds#1084

sofyenne:
Prior to this change, when we searched for synchronized people by their phone numbers, no results were displayed. This issue was due to the multi-valued synchronized property, which was saved as a map of string key-value pairs like "phones.home."

This change is going to prevent the profile property from being saved using the old user profile format and will instead be saved in the format of parent and child using a map of string and object key-value pairs.

### DIFF
--- a/component/api/pom.xml
+++ b/component/api/pom.xml
@@ -23,7 +23,7 @@
   <parent>
     <artifactId>social-component</artifactId>
     <groupId>org.exoplatform.social</groupId>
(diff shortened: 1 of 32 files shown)


ahamdi (dismissed):

exo-swf:

ahamdi (approved):

PR #2804, opened 2023-09-07
 feat: Adjust the style of the activity stream title when the activity composer is enabled for users - MEED-2402 - Meeds-io/meeds#54

MayTekayaa:
This change will adjust the style of the activity stream title when the activity composer is enabled for users.


boubaker:
Need to make Automatic Tests ok first please.

PR #2820, opened 2023-09-10
feat: Move Notification Rest & UI from Commons - MEED-2445 - Meeds-io/MIPs#79

### LINKED ISSUE #79: Notification Administration - Upgrade UI
## Rationale

Notifications Administration UI is not consistent with other platform UI.
Indeed, the following UI is not accurate with what is expected to see:

image

In addition to the fact that UX is not the best here:
* Sender informations are "hidden" as it is at the bottom of the app
* To disable / enable a notification channel, it is needed to do it notification per notification

## 1. Functional Requirements

The UI will be updated to make sure it is consistent with other UI in the platform.

View of future notifications admin page:
image

With ability for admin to enable/disable any notification by channel type.  When disabled, users won't be able to enable it.
With ability for admin to set the notification sender settings with a proper drawer:
* to set the sender 
* to edit the sender mail without changing the domain name

### Upgradability

The old portlet `NotificationsAdminJuzuPortlet` has to be replaced in social by a portlet with name `NotificationAdministration`.

The new portlet has to be made using Vue & REST endpoints.

### Migration strategy

The definition of UI & REST endpoint has to be moved to social. Knowing that the old portlet will be replaced, the `importMode` of administrators portal has to be of type `merge` or `overwrite`.

boubaker:
This change will move Notification Administration UI from Commons into Social, same as the REST API endpoint. The new portlet is defined using Vue instead of Juzu.


rdenarie (commented):
on mips-ft instance, on notification administration, plugin names are not translated when I'm in french :
image

As I saw you create french labels files, I think there is a problem with that.

In addition, in the top of the new portlet, there is also i18n pb : 
image

boubaker:
> on mips-ft instance, on notification administration, plugin names are not translated when I'm in french :

It's new labels in fact and not translated yet, see it here
https://github.com/Meeds-io/social/pull/2820/files#diff-bc05a177b12dcaa482ca2afc7871e95161a84b1685af5d8d175beb262e124afc

rdenarie (approved):

PR #2828, opened 2023-09-12
feat: Review the activity stream filter option - MEED-2422 - Meeds-io/MIPs#54 

MayTekayaa:
This change will fix the filter icon color.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityStreamToolbar.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityStreamToolbar.vue
@@ -151,7 +151,7 @@ export default {
       this.$userService.getUser(eXo.env.portal.userName)
         .then(user => this.user = user);
     }
-    document.addEventListener('activity-stream-type-filter-changed', event => {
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityStreamToolbar.vue
```diff
@@ -139,13 +141,19 @@ export default {
...
+    filterIconColor() {
+      return this.streamFilter !== 'all_stream' && 'primary';
     }
   },
   created() {
     if (!this.user) {
       this.$userService.getUser(eXo.env.portal.userName)
         .then(user => this.user = user);
     }
+    document.addEventListener('activity-stream-type-filter-changed', event => {
```
boubaker: Can you use the existing event `activity-stream-type-filter-applied` ?

```suggestion
    document.addEventListener('activity-stream-type-filter-applied', event => {
```

MayTekayaa: done!

boubaker (approved):

PR #2885, opened 2023-09-18
fix: Notification read display - EXO-66167 - Meeds-io/meeds#1103

akhanfir:
Prior to this change, when open notif drawer, click on see all button, click on any notif unread and check notif in drawer or on full page, notif colored in blue as if it was unread. To fix this problem, added a script to trigger the rest of the change to read after clicking on this notif. After this change, this notif is colored white and marked as read.


ahamdi (commented):
Could you fix the code conflicts ?

PR #2945, opened 2023-09-26
fix: Review the display of the composer drawer when space audience chosen - MEED-2569 - Meeds-io/MIP#54

MayTekayaa:
This change will review the display of the composer drawer when space audience chosen.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
@@ -29,47 +29,86 @@
     <template #content>
       <v-card flat>
         <div v-if="audienceTypesDisplay" class="mt-1 px-4 pt-4">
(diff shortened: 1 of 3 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
```diff
@@ -29,47 +29,86 @@
...
-              close
-              @click:close="removeAudience">
-              <v-avatar left>
-                <v-img :src="audience.profile.avatarUrl" role="presentation" />
-              </v-avatar>
-              <span class="text-truncate">
-                {{ audience.profile.fullName }}
-              </span>
-            </v-chip>
+          <div v-if="(postInYourSpacesChoice && !audience) || postInYourNetwork">
```
boubaker: May be add this in a dedicated computed param, WDYT ?

MayTekayaa: yess, done!

boubaker (approved):

PR #2955, opened 2023-09-28
Merge Review composer UX - Meeds-io/MIPs#54

### LINKED ISSUE #54: Composer UX additions
## Rationale

The composer provides different options currently:
1. write a message
2. create a poll

It misses a way to send kudos as it is also a message. For clarity sake and for unified UX, it seems to be interesting to get this access from the composer as well.

In addition to this, when it is possible to write a message from the general stream, end-users cannot choose their audience. The UX is not clear to end-users and thus, this option is disabled.
While it is requested by end-users to easily write a message and choose an audience.

## 1. Functional Requirements

1. From the general stream, propose to write a message 
- with a call to action clearly proposed to the user

- with a way to choose an audience among his network or one of spaces he is member of - or redactor if needed (only one space, and if current user is member of it)

> this option of choosing "your network" audience will be provided by default. 
Information: A property will be added to remove this so only spaces can be chosen as audience

image

> This whole option of posting from general stream needs to be enabled/disabled by config

- with option left to filter so it is known by users they can send a kudos from there:
image

When sending a kudos from there, an audience can be chosen as well as the kudos receiver 

image

5. From a space, propose same UX but filtered to the actual audience of course with CTA to send kudos and attach poll:
 
- create a poll for a community
- send kudos in the current community

image

### Extensibility 

The list of **Activity Composer Quick Actions** must be extensible the same way as done for **Activity Composer Drawer Actions** with a different __extensionName__.

### Existing Features

**Activity Composer Drawer Actions** must remain compatible with its extensibility API using extensionRegistry.

### Feature Flags

| Property Name  | Default Value | Target Audience | Functional Behaviour |
| ------------- | ------------- | ------------- | ------------- |
| exo.feature.PostToNetwork.enabled | true (in Meeds Package only) | All | Allows to a user to post on his network or not. Default set to true to allow users to post in their personal stream. |

### Access 

- The main change is about Frontend components update, thus no change will be needed on REST endpoints
- A new component has to be defined for **Activity Composer Quick Actions** where we will be able to define by extensionRegistry the actions that will be listed there

### Services & processing

No notable backend

MayTekayaa:
This change will merge the review composer UX MIP.


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityStreamFilterDrawer.vue
```diff
@@ -0,0 +1,128 @@
+<!--
+  This file is part of the Meeds project (https://meeds.io/).
+  Copyright (C) 2022 Meeds Association
```
rdenarie:
```suggestion
  Copyright (C) 2023 Meeds Association
```

MayTekayaa: done!

webapp/portlet/src/main/webapp/groovy/social/webui/UISocialPortalApplicationHead.gtmpl
```diff
@@ -65,4 +65,5 @@
 
      eXo.env.portal.SpaceHomeLayoutResetEnabled = <%=featureService.isFeatureActiveForUser("SpaceHomeLayoutReset", userName)%>;
      eXo.env.portal.PublisherRolePromotionFeatureEnabled = <%=featureService.isFeatureActiveForUser("publisherRolePromotion", userName)%>;
+     eXo.env.portal.postToNetworkEnabled = <%=PropertyManager.getProperty("exo.feature.PostToNetwork.enabled")%>;
```
rdenarie: @boubaker should we use featureService instead of propertyManager ?
```suggestion
     eXo.env.portal.postToNetworkEnabled =  <%=featureService.isFeatureActiveForUser("PostToNetwork", userName)%>;
```

rdenarie (changes requested):

webapp/portlet/src/main/webapp/groovy/social/webui/UISocialPortalApplicationHead.gtmpl
boubaker: Yes, good idea, if we need to enable the feature for some users only.

rdenarie (dismissed):

rdenarie (approved):

PR #2982, opened 2023-10-04
fix: Fix the space stream filter - MEED-2622 - Meeds-io/meeds#1148

MayTekayaa:
This change will fix the space stream filter.


webapp/portlet/src/main/webapp/vue-apps/activity-stream/js/ActivityUtils.js
```diff
@@ -0,0 +1,8 @@
+export function getStreamFilter() {
+  console.warn('test', eXo.env.portal.spaceId || 'all');
```
boubaker:
```suggestion
```

MayTekayaa: done!

boubaker (approved):

PR #2986, opened 2023-10-04
feat: Redirect to the first allowed page or link sub root node when clicking on a group non meta site root node - EXO-66391 - Meeds-io/MIPs#72

GouadriaHanen:


azayati (approved):
LGTM

PR #3060, opened 2023-10-17
fix: Fix the users mention in CKeditor - MEED-2666 - Meeds-io/meeds#1160

MayTekayaa:
This change will fix the users mention in CKeditor when switching audience from composer drawer.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
@@ -294,7 +294,6 @@ export default {
       }
     },
     spaceURL() {
-      this.ckEditorInstance.refreshSuggester();
       this.resetRichEditorData();
     }
   },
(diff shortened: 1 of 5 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
```diff
@@ -444,6 +458,90 @@ export default {
...
+                          <del>${profile.fullname}${profile.isExternal === 'true' ? ` (${  this.$t('UsersManagement.type.external')  })` : ''}</del>
+                          <a href="#" class="remove">
+                            <i class="uiIconClose uiIconLightGray"></i>
+                          </a>
+                        </span>
+                      </span>
+                    `));
+                    this.containInvalidUsers = true;
+                  }
+                });
```
boubaker: It would be better to set a new value of `containInvalidUsers` when switching audiences, else, the warning can be displayed whereas it could display a wrong information to end user.

```suggestion
                  }
                });
                this.containInvalidUsers = userProfiles.find(p => !p.isMember);
```

MayTekayaa: I have set a false value for the variable when switching audience

boubaker: ok, there are two `if` in fact, we will have to set a value when switch the audience to "my network" by example.

boubaker: BTW, it doesn't contain all time invalid users since we can switch to a space where the mentioned users are member as well, thus the message shouldn't be displayed.

MayTekayaa: done!

exo-swf:

boubaker (approved):

PR #3063, opened 2023-10-17
feat: Delete TimeZone UI from User Settings - MEED-2691 - Meeds-io/meeds#993

boubaker:
Prior to this change, the TimeZone User Settings UI was proposed to end user while it's not used anymore. This change will delete this UI from Settings page. At the same time, this change will preload the REST API call for User Notification Settings to improve Page performances loading.


SaraBoutej (dismissed):

webapp/portlet/src/main/webapp/WEB-INF/jsp/portlet/userSettingNotifications.jsp
```diff
@@ -0,0 +1,21 @@
+<%@page import="org.exoplatform.social.core.space.SpaceUtils"%>
```
SaraBoutej, line 21: missing file header

boubaker: done

SaraBoutej (approved):

webapp/portlet/src/main/webapp/WEB-INF/jsp/portlet/userSettingNotifications.jsp
boubaker: As explained in one to one, it's not needed in general to have the license header in JSP and HTML files to avoid increasing the HTML DOM response size as you can see it here :
Capture d’écran du 2023-10-17 13-14-46

But in JSP context, we can use JSP based comments or Java comments which will not be interpreted as a part of DOM response, that's why I added it here. Besides, this requirement is excluded even in sonar checks to not have such HTML comments when developers attempts to resolve the Sonar checks.

PR #3072, opened 2023-10-19
fix: Space Activity Stream Load more issue - EXO-66818 - Meeds-io/meeds#1199

sofyenne:
Before this change, when attempting to load more than 50 space-posted activities, an  HeadersTooLargeException exception would be raised. This issue was caused by the addPreloadActivityIds method, which added the entire list of Activity IDs to the response headers when the preload limit value exceeded the length of the activity IDs list. This change resolves this issue by recalculating the limit and offset in the specific case where the limit is greater than the length of the activity IDs list

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
@@ -1075,8 +1075,11 @@ public class ActivityRestResourcesV1 implements ResourceContainer {
 
   private void addPreloadActivityIds(List<String> activityIds, int preloadLimit, String expand, ResponseBuilder responseBuilder) {
     int offset = preloadLimit > MAX_TO_PRELOAD ? preloadLimit - MAX_TO_PRELOAD : 0;
-    List<String> preloadActivityIds =
-                                    activityIds.size() >= preloadLimit ? activityIds.subList(offset, preloadLimit) :  activityIds.subList(offset > MAX_TO_PRELOAD ? offset : 0, activityIds.size());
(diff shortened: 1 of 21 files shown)


component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
```diff
@@ -1076,7 +1076,7 @@ private List<DataEntity> convertToEntities(List<ExoSocialActivity> activities,
   private void addPreloadActivityIds(List<String> activityIds, int preloadLimit, String expand, ResponseBuilder responseBuilder) {
     int offset = preloadLimit > MAX_TO_PRELOAD ? preloadLimit - MAX_TO_PRELOAD : 0;
     List<String> preloadActivityIds =
-                                    activityIds.size() >= preloadLimit ? activityIds.subList(offset, preloadLimit) : activityIds;
+                                    activityIds.size() >= preloadLimit ? activityIds.subList(offset, preloadLimit) :  activityIds.subList(offset > MAX_TO_PRELOAD ? offset : 0, activityIds.size());
```
ahamdi: recalculate the preload limit the salme as the offset in a dedicated statement to make the code more readable

ahamdi (changes requested):
Could you add a Unit test ?

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java
sofyenne: done !

exo-swf:

ahamdi (approved):

PR #3087, opened 2023-10-20
feat: Desktop Notification - MEED-1651 - Meeds-io/meeds#617

sergeByishimo:
Prior to this change you have to periodically visit the web page to check if there is a new notification or mention.This change notifies you right away after receiving a Web Notification by using Desktop Notifications when a new Web Notification is received.

Hi @boubaker  this is partially done. I only did notifications of the `social` module successfully, I need your review about the code structure and approach so that I can continue with the rest of the modules. 

As you can see I used a switch case on each pluginId, on the rest of the modules I would do the same by adding more switch cases in the file `TopBarNotification.vue`. I want to know beforehand if the approach is alright with you or if there is a better way to do it.

Thanks 🙏🏼


boubaker:
Not necessary anymore since we have PWA based Push Notifications

PR #3097, opened 2023-10-25
fix: add missing id attributes to the Share activity drawer - EXO-67173

ahamdi:
The fix will add missing id attributes to the interactive elements of the share activity drawer


azayati (dismissed):
LGTM

azayati (approved):
LGTM

PR #3102, opened 2023-10-26
feat: Implement Public Site Applications - MEED-2505 - Meeds-io/MIPs#84

### LINKED ISSUE #84: Public site management
## Rationale

Once connected to the WoM, Hubs can be listed in the [hub explorer](https://meeds.io/hubs)
To avoid having something with no contents and only user engagement data, we aim to showcase organization's purpose.
That way, visitors and future contributors or leaders can understand how the hub manager is engaging its contributors.

### Top User Stories

**Public site Management**

As a platform administrator, I can manage the public site from the main settings:
* Make it visible to anyone
* Copy address
* Access the edition of the site

image

**Public site layout**
The layout provided by this public site is dedicated to this site. 
That means the site is considered as "standalone" and provides its own layout:
- topbar
- portlets

**Public site Access**

As platform admin or platform publisher,
- even if not made visible to anyone, I can access the public site to edit it (from the option or from the direct link (/portal/public)
- I can edit portlets proposed in this public site

As regular user, 
- if not made visible to anyone, then when trying to access the public site, I am sent to home or login page if not logged in yet
- if made visible to anyone, then I only see portlets that have been filled by the administrator
- if logged in, the topbar reminds me it and I can access the platform from there
- if not logged in, the topbar suggest me to login

When logged in
image

When not logged in
image

**Public Site Edition**

As platform administrator or platform publisher, I can edit the content proposed in the layout:
* image portlet (MIP #98)
* link portlet (MIP #92)
* single note view portlet (MIP #93)

image

And other content are not editable from there:
* topbar with branding logo and name + logout/login button to access the platform
* if secondary pages are identified for the public site, then display in the topbar the menu portlet
* To be done later: List of programs and related actions/leaderboard that are "public"

### Impacts
Permissions to access the public site
Topbar - public case

#### Gamification
NA

#### Notifications
NA

#### Analytics
Users coming from public site are identified as well as users who sign-up from public site.

#### Unified Search
NA

### Security 
The public site has to be accessible to `*:/platform/administrators` and `*:/platform/web-contributors` until it's made accessible to `Everyone` when an administrator switches the site as publically accessible.

### Configurability 

The Public site pages has to hide the display of shared layout components. The list of topbar applications has to be defined in portal layout.
This site has to be defined and configured to not be re-imported on server restart.

### Feature Flags

No feature flag is needed.

### Security

The Site `public` of type `Portal` has to define `access-permissions` attribute to `*:/platform/administrators` and `*:/platform/web-contributors`. The `edit-permission` attribute allows only one value, thus it will be made possible to only an administrator (member of  `*:/platform/administrators`) to switch the site as publically accessible.

#### Applications

Two new portlets has to be added:
* Login button / Access button: displayed in topbar
* Public status switcher application: displayed in page body for `*:/platform/administrators` only when the site isn't publically accessible yet.

##### Site definition configuration

The XML definition of the `Public` portal has to be added in `Meeds` module to allow defining the whole page without dynamic containers.
Knowing that :
* the extension `meeds.war` holds the definition of `meeds` site, which is specific to the Meeds packaging (can be excluded by Meeds DAO members to define their own site `Meta`)
* the `public` site is a core feature of the Meeds product and its administration is part of `Main Settings` administration UI

We will have to:
* Define the `public` site in a common `Web Archive Extension` that is decoupled from `meeds.war`
Thus, a new `Web Archive Extension` named `sites.war` will be added in order to centralize the `Meeds core sites` into it including `global` site and `public` site.

##### Layout configuration

The public page has to be configured to hide the shared layout. Thus, until the MIP #85 is ready, we will use the following attribute in page definition :
```xml
<hide-shared-layout>true</hide-shared-layout>
```
The page, applications and containers has to use `access-permission` with value `Everyone` so that the visibility switch operation will take effect on `Portal` access permission only.

#### REST API

To switch the site permissions, anew Method will be added in `org.exoplatform.social.rest.impl.site.SiteRest` with `PATCH` method that will allow to alter the site `access-permissions` value.

boubaker:
This change will implement necessary public site applications such as Login Button and an Alert display to make the site publically accessible or not. In addition this change will update Main Settings UI to allow manage public site from it.


component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
```diff
@@ -250,6 +257,12 @@ public Response getSiteById(@Context
                               String lang) {
     try {
       PortalConfig site = layoutService.getPortalConfig(Long.parseLong(siteId));
+      if (site == null) {
+        return Response.status(Status.NOT_FOUND).build();
+      }
+      if (!userACL.hasAccessPermission(site)) {
+        return Response.status(Status.UNAUTHORIZED).build();
+      }
```
rdenarie, line 265: This code will allow anonymous user to find informations about sites : 
By making the search with different siteId, he will be able to find ones corresponding to a real site : 
When search return 404 => it is a not existing id
When search return 401 => it is an existing id.

2 options here : 
- Check the hasAccessPermission before the site == null : it will reduce the possibility but to remove completly the problem
- Return not authorized in both case : it is the solution I prefer

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
```diff
@@ -279,6 +292,48 @@ public Response getSiteById(@Context
...
+                                 @Parameter(description = "Site attribute value", required = true)
+                                 @FormParam("value")
+                                 String value) {
+    try {
+      PortalConfig site = layoutService.getPortalConfig(Long.parseLong(siteId));
+      if (site == null) {
+        return Response.status(Status.NOT_FOUND).build();
+      }
+      if (!userACL.hasEditPermission(site)) {
+        return Response.status(Status.UNAUTHORIZED).build();
```
rdenarie, line 322: Same remark

rdenarie (changes requested):

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
boubaker: `This code will allow anonymous user to find informations about sites`

Nope, this endpoint is protected by `@RolesAllowed("users")`, so not visible publically. Besides, the existence of `siteId` parameter isn't a sensitive information that we will need to protect.

In fact, having a 404 or 401 on URL `/v1/social/sites/1` will not give a sensitive information to the end user. The fact to return the `Not found` and `Not authorized` code is better to give more details about error and make the API more self explanatory.

The fact to return the same response code when resource not found and resource not authorized is effectively relevant when the identifier of the resource itself is security concern and shouldn't be known if it exists or not, like using site name, user names, user emails .... but here for an auto generated and incremented id, no need to avoid the difference in favor of the REST API lisibility.

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
boubaker: Same comment as https://github.com/Meeds-io/social/pull/3102/files#r1375954876

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
rdenarie: Ok. I miss the @Allowed

rdenarie (approved):

component/service/src/main/java/org/exoplatform/social/rest/impl/site/SiteRest.java
boubaker: ok, what about the answer about the other concern: the auto incremented id information isn't relevant to protect ? WDYT ?

rdenarie: No, it is ok

boubaker: > No, it is ok

Cool, thank. I asked because the `@RolesAllowed` will be removed in Meeds-io/MIPs#100, so just to be sure that it will not be a concern for you when it will be available.

PR #3104, opened 2023-10-27
Feat: Add new property to the generic widget component- MEED-2803 - Meeds-io/MIPs#81

SaraBoutej:
This PR adds a new property **minHeight** to the generic widget component.


webapp/portlet/src/main/webapp/vue-apps/common/components/Widget.vue
```diff
@@ -18,6 +18,7 @@
     :loading="loading"
     :class="extraClass"
     :height="height"
+    :min-height="minHeight"
     :min-width="minWidth"
     class="d-flex flex-column white card-border-radius overflow-hidden"
```
boubaker, line 23: shouldn't we add here class `fill-height` or `full-height` to let the container determine the minimum height ?
BTW, the application can also determine the minimum height of the content using a `<v-card min-height` so I think no need for this change in the Widget reusable component.

SaraBoutej: ok so no need for this PR!

PR #3106, opened 2023-10-30
fix: Clean the profile property group synchronized name of special characters - EXO-66936 - Meeds-io/meeds#1228

sofyenne:
Prior to this change, after synchronizing a profile property value containing a special character, the group name caused an error when we attempted to retrieve a process. 
This change will clean the group name.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/listeners/GroupSynchronizationSocialProfileListener.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/listeners/GroupSynchronizationSocialProfileListener.java
@@ -150,14 +150,16 @@ public class GroupSynchronizationSocialProfileListener extends ProfileListenerPl
   }
 
   private Group getOrCreateGroup(String groupName, Group parentGroup) throws Exception {
+    String groupLabel = groupName;
+    groupName = Utils.cleanString(groupName);
     Group group = getGroup(buildGroupId(parentGroup, groupName));
     if (group != null) {
       return group;
     }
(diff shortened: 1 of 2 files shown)


rdenarie (changes requested):
Let say that the function getOrCreateGroup was firtsly called with groupName = "test'property"
the created group is /profile/testproperty
Then when getOrCreateGroup is called a second time, does this line 
` Group group = getGroup(buildGroupId(parentGroup, groupName));`
will find the group ? 
In buildGroupId, the groupName is lowered, but not cleaned. So it will return null as group /profile/test'property not exists, so it will try to recreate a group.


I think that, if you add this line 
`identityManager.updateProfile(profile, true);`

at the end of your unit tests (to simulate a second save), it will throw an error.

You should fix this point

sofyenne:
> Let say that the function getOrCreateGroup was firtsly called with groupName = "test'property" the created group is /profile/testproperty Then when getOrCreateGroup is called a second time, does this line ` Group group = getGroup(buildGroupId(parentGroup, groupName));` will find the group ? In buildGroupId, the groupName is lowered, but not cleaned. So it will return null as group /profile/test'property not exists, so it will try to recreate a group.
> 
> I think that, if you add this line `identityManager.updateProfile(profile, true);`
> 
> at the end of your unit tests (to simulate a second save), it will throw an error.
> 
> You should fix this point

fix updated by cleaning the group name before calling the getGroup method to attempt to retrieve the group by the correct group name if it exists , unit test also updated!

rdenarie (approved):

PR #3182, opened 2023-11-14
fix: Correctly display the space after space selection in composer - EXO-67423 - meeds-io/meeds#1273

rdenarie:
After previous fix, the publish button is always disabled when publishing in a space with PostToNetwork = false This is because audience is always empty.
This fix do not reset audience when the user is in a space context


azayati (approved):
LGTM

PR #3238, opened 2023-11-27
fix: fix Left navigation level 2 and vertical menu portlet site icon and label alignement - EXO-67683 - Meeds-io/meeds#1330

GouadriaHanen:
before this fix, the site icon and label were not aligned
after this fix, the margins and paddings were decreased and the items were aligned:

- Icon is centered and aligned vertically by including it in "v-list-item-icon" so all icons are aligned the same way
- Each label is aligned to the icons with 12px of padding


srenault-meeds:
Hello,
Can you please detail what this will fix? Horizontal only or also Vertical alignement.
Indeed, we have added an issue about the vertical alignement but we can't see any issue about what you are reporting in your description.

https://github.com/Meeds-io/meeds/issues/1330

cc @azayati

GouadriaHanen:
> Hello, Can you please detail what this will fix? Horizontal only or also Vertical alignement. Indeed, we have added an issue about the vertical alignement but we can't see any issue about what you are reporting in your description.
> 
> [Meeds-io/meeds#1330](https://github.com/Meeds-io/meeds/issues/1330)
> 
> cc @azayati

description updated! both mentioned expected behaviors were fixed 

```
1. Each icon is centered and aligned vertically so all icons are aligned the same way. Might be interesting to provide a "box" where icons can be added there so we make sure it is still well aligned no matter which icon is used
2. Each label is aligned to the left vertically with a minimum of 12px padding between the icon and the label
```

GouadriaHanen:
already fixed
https://github.com/Meeds-io/social/pull/3241/files

PR #3249, opened 2023-11-29
feat: Dispatch dismissed event in common snackbar - EXO-67599 - Meeds-io/meeds#1367

hakermi:
Current common snackbar doesn't satisfy some functional needs for specific applications. The non existing of a dispatched event on alert dismiss prevents triggering actions on some app side on the dismiss state. 
This PR makes sure to dispatch a dismissed event on auto dismiss or a dismiss by close button of the alert.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/common/components/Notifications.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/common/components/Notifications.vue
@@ -28,7 +28,8 @@
     class="z-index-snackbar"
     color="transparent ma-0"
     elevation="0"
-    app>
+    app
+    @input="dispatchDismissed">
     <confeti-animation
       v-if="confeti"
       class="overflow-hidden" />
@@ -92,7 +93,7 @@
       <template v-if="!isMobile" #close="{toggle}">
         <v-btn
           icon
-          @click="toggle">
+          @click="dispatchDismissed(!toggle)">
           <v-icon size="16" class="icon-default-color">fa-times</v-icon>
         </v-btn>
       </template>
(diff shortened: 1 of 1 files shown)


rdenarie (commented):
It is not a fix, but a feature, can you update your commit message

hakermi:
> It is not a fix, but a feature, can you update your commit message

Done

webapp/portlet/src/main/webapp/vue-apps/common/components/Notifications.vue
```diff
@@ -223,6 +224,11 @@ export default {
     });
   },
   methods: {
+    dispatchDismissed(opened) {
+      if (!opened) {
+        document.dispatchEvent(new CustomEvent('alert-message-dismissed'));
```
boubaker, line 229: I propose to add a watch on `snackbar` when it's turned off, in order to centralize the event triggering.

hakermi: already tested that and it didn't work correctly because of delay issues in dispatched event from watch. the issue appears when there are several alerts when the second alert forces the close of the first one.
It caused an issue in application side

hakermi: This may explains more :)
Result when dispatch event in watch sanckbar (the issue is that the warning alert didn't even appear) :
watch_snackbar
Result of current work
usemethod

boubaker: ok, seems not clear why it doesn't work, but it didn't as you demonstrate with images. Not blocker for me, even if I would like to analyse more why it didn't work using the watch while it should. Anyway, I resolve the conversation.

rdenarie (approved):

PR #3257, opened 2023-12-01
feat: Refresh Space Cache even when user is already member of space - MEED-2602 - Meeds-io/meeds#1142

boubaker:
Prior to this change, when the user is added to a space via `OrganizationService.MembershipHandler`, the caches aren't updated as when doing it via `SpaceService.addMember`. This change ensures to clar the cache even after adding the user via `OrganizationService.MembershipHandler.linkMembership`


AzmiTouil (approved):

srenault-meeds (commented):
Test

srenault-meeds:
test

srenault-meeds (dismissed):
test1

PR #3258, opened 2023-12-01
fix: Change the banner of Platform settings site - EXO-67724 - Meeds-io/meeds#1354

IlhemEssaadi:
After this change, we change the default platform settings banner by the new one.


azayati (approved):
LGTM

PR #3268, opened 2023-12-01
fix: Fix single attributes when it is considered as parent  - EXO-67825 - Meeds-io/meeds#1383

GouadriaHanen:
before this change, while retrieving profile attributes are acting like a single String value, after considering it as a parent when retrieving children a cast exception is thrown due to the non-match of types (String and ArrayList)
After this change, only multivalued attributes are suggested as a parent, and for all the default attributes their multivalued values is the default value (could not be updated)


ahamdi (commented):
IMO those attributes should never be used as parent attributes, could you check this from functional POV
besides, couldn't we do a check on the type of the value before casting ?

ahamdi (commented):
Could you add a Unit test for the new added changes ?
Besides, what happens if there is already a similar change in a server ? what is the expected behaviour ?

ahamdi (approved):

PR #3269, opened 2023-12-04
feat: Dispatch dismissed event in common snackbar - EXO-67599 - Meeds-io/meeds#1367 (#3249)

Jihed525:
Current common snackbar doesn't satisfy some functional needs for specific applications. The non existing of a dispatched event on alert dismiss prevents triggering actions on some app side on the dismiss state.
This PR makes sure to dispatch a dismissed event on auto dismiss or a dismiss by close button of the alert.

(cherry picked from commit 6bfe6abce5f5533e6286b0c7449f7aceb1c87509)

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/common/components/Notifications.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/common/components/Notifications.vue
@@ -28,7 +28,8 @@
     class="z-index-snackbar"
     color="transparent ma-0"
     elevation="0"
-    app>
+    app
+    @input="dispatchDismissed">
     <confeti-animation
       v-if="confeti"
       class="overflow-hidden" />
@@ -92,7 +93,7 @@
       <template v-if="!isMobile" #close="{toggle}">
         <v-btn
           icon
-          @click="toggle">
+          @click="dispatchDismissed(!toggle)">
           <v-icon size="16" class="icon-default-color">fa-times</v-icon>
         </v-btn>
       </template>
@@ -223,6 +224,11 @@ export default {
     });
   },
   methods: {
+    dispatchDismissed(opened) {
+      if (!opened) {
+        document.dispatchEvent(new CustomEvent('alert-message-dismissed'));
+      }
+    },
     openAlert(params) {
       this.reset();
       this.closeAlert();


webapp/portlet/src/main/webapp/vue-apps/common/components/Notifications.vue
```diff
@@ -92,7 +93,7 @@
       <template v-if="!isMobile" #close="{toggle}">
         <v-btn
           icon
-          @click="toggle">
+          @click="dispatchDismissed(!toggle)">
```
boubaker, line 96: The `toggle` is a function that must be called to toggle alert closing as well.

boubaker: Never-mind, will fix it when needed.

ahamdi: @boubaker Could you five me more details, I will check it with the developers and fix it.

ahamdi (approved):

webapp/portlet/src/main/webapp/vue-apps/common/components/Notifications.vue
boubaker: > @boubaker Could you five me more details, I will check it with the developers and fix it.

Yes, if you want, but it's for better usage of Vue & Vuetify builtin functions.
As said, the `toggle` is a function that will be used to update the `v-model` of `v-alert` (as you can see [it here](https://v2.vuetifyjs.com/en/api/v-alert/#slots)). In order to drive the v-alert hide, we can rely on `@input` event of `v-alert` or add a dedicated `v-model` for the `v-alert` and thus a `watch`ed `data` variable.

This is just to ensure that the events are triggered when clicking on button or elsewhere, even when the alert is closed due to timeout. So this is just an enhancement and not blocker, that's why it was marked as resolved.

But, even with the current change, knowing that `toggle` variable is never `falsy`, `dispatchDismissed(!toggle)` is exactly the same as `dispatchDismissed(false)`.

To not have to retest other changes, knowing that all of this are enhancements, no need IMO to apply further change. We will change it `when needed` as said :+1:

PR #3275, opened 2023-12-05
fix: update upgrade plugin to support Posgresql DB - EXO-67619 - Meeds-io/meeds#1373

ahamdi:
the upgrade plugin **SpaceNavigationIconUpgradePlugin** fails to execute, the sql query run inside it fails with Postgresql DBs. The fix updates that SQL query to run correctly on Mysql, Postgresql and HSQL databases.


azayati (approved):
LGTM

PR #3280, opened 2023-12-05
fix: Fix topbar menu accessibility - EXO-62481

ahamdi:
before this change, an accessibility issue for the role and aria-label of the topbar items are missing or don't match to their role

after this change, the right roles and aria-label have been added


azayati (approved):
LGTM

PR #3317, opened 2023-12-13
fix: Notif snackbar not closed - EXO-68048 - Meeds-io/meeds#1425

akhanfir:
Before this change, when snackbar diplayed then clicking on X, doesn't close the alert snackbar. To resolve this problem, we change the click event from close to event toggle declared in closing and we launch the alert-message-dismissed event only when the snackbar diplayed and the handle alert is clicked. After this change, the snackbar works correctly.


rdenarie (commented):
Please update the commit message with an english message.
Thanks

ahamdi (approved):

PR #3318, opened 2023-12-13
fix: fix displaying activity by id in stream - EXO-68049 - Meeds-io/meeds#1460

GouadriaHanen:
Before this change, for non meta sites when displaying an activity with its id, the activity stream [checks](https://github.com/Meeds-io/social/blob/develop/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/ActivityStream.vue#L71) first of all if the current URL contains the activity base link which takes the meta site as the portal name instead of the current portal which leads to non coherent activity URL. 
After this change, the activity base URL should take the current site name instead of the meta site name.


azayati (approved):
LGTM

azayati:
@GouadriaHanen can you create the meeds issue ?

azayati (changes requested):
@GouadriaHanen can you create the meeds issue ?

PR #3320, opened 2023-12-14
fix: Notification snack-bar not closed - EXO-68048 - Meeds-io/meeds#1425 (#3317)

Jihed525:
Before this change, when snack-bar alert is diplayed & clicking on X, it doesn't close the alert snack-bar.
To resolve this problem, closing click event was replaced by toggle event on closing & launch event alert-message-dismissed only when snack-bar alert is displayed & handle alert is clicked.
Thus, snack-bar alert closing works as expected.
(cherry picked from commit https://github.com/Meeds-io/social/commit/a664de20e3106284dcdba2d0d4e7f9e8a8b2a72c)


sofyenne (approved):

sofyenne (commented):
there is a build failure!

PR #3327, opened 2023-12-15
Add the space administration portlet to the ManageSpaceGRP load group - EXO-68206 -  Meeds-io/meeds#1467

IlhemEssaadi:
Prior to this change, the site navigation icon is not displayed on the space administration page due to the hide of sharedLayout for the aggregated site containing the site navigation portlet.This commit creates a load-group for site navigation, which will then be loaded with the Spaces Administration Portlet.


rdenarie (commented):
I'm not sure this is correct to add the js of a portlet in a load group
Can you ping me to discuss about this point

cc @ahamdi

rdenarie (approved):

PR #3328, opened 2023-12-18
fix: Handle 'Everyone' permission when computing site permissions - EXO-68132 - Meeds-io/meeds#1474

sofyenne:
Prior to this change, after setting the site access permission to 'Everyone' and saving, the displayed access permission on the manage site permissions drawer was incorrect. This was due to mishandling of the 'Everyone' permission when computing site permissions. This change addresses the issue by appropriately handling the 'Everyone' permission case during the computation of site permissions.


azayati (approved):
LGTM

PR #3363, opened 2023-12-27
fix: Problem when option post to network is set to false - EXO-67423 - Meeds-io/meeds#1273

GouadriaHanen:
Before this fix, when the feature PostToNetwork is disabled, the suggester does not correctly disappear to display the chosen space, as when the feature is activated. This is due to the audience, which is not correctly set to oneOfYourSpaces when the PostToNetwork is disabled. To fix it, we set the value of audienceChoice in the function of the feature PostToNetwork status.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
@@ -438,7 +438,7 @@ export default {
       }
     },
     resetAudienceChoice() {
-      this.audienceChoice =  eXo.env.portal.postToNetworkEnabled && 'yourNetwork' ||  'oneOfYourSpaces';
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
```diff
@@ -438,7 +438,7 @@ export default {
       }
     },
     resetAudienceChoice() {
-      this.audienceChoice = 'yourNetwork';
+      this.audienceChoice =  eXo.env.portal.postToNetworkEnabled && 'yourNetwork' ||  'oneOfYourSpaces';
```
azayati:
```suggestion
      this.audienceChoice = eXo.env.portal.postToNetworkEnabled && 'yourNetwork' || 'oneOfYourSpaces';
```

azayati (changes requested):

azayati (approved):
LGTM

PR #3369, opened 2023-12-29
fix: Suggestor isn't displaying users - EXO-68574 - Meeds-io/meeds#1503

Jihed525:
Before this change, when create user1 & user2, add user & user2 in spaceX, on spaceX, user1 Sends kudos and on suggestor types user2 name, suggestor isn't displaying any results. To resolve this problem, added the default profile.fullName value to the itemText props. After this change, suggestor should display results related to typed string.


azayati (approved):
LGTM

PR #3388, opened 2024-01-09
feat: Implement a quick search drawer based on user profile information - EXO-68495 - Meeds-io/MIPs#110

hakermi:
Implement a quick search result drawer based on user profile information


ahamdi (approved):
LGTM

PR #3401, opened 2024-01-15
fix: fix User and space avatar are not accessiblility - EXO-68785_68787

GouadriaHanen:
this change fix user and space avatar accessibility problem


mkrout (dismissed):
LGTM

rdenarie (changes requested):
I think the proposed pr is KO.
For space and user afvatar, we want to exclude it from assistances tools.
So we need 2 things :
Remove the presentation role which is not suitable for an image
Set the 'alt' attribute to "". This a valid value, which inform assistance tools to not read this icon. So we do not want aria-label, we only want an empty alt attribute.
As I can see, the second point was not done in the PR

rdenarie (approved):

PR #3402, opened 2024-01-15
feat: UI improvement of the people list in mobile - EXO-68642 - Meeds-io/MIPs#110

hakermi:
UI improvement of the people list in mobile


ahamdi (approved):
LGTM

PR #3407, opened 2024-01-16
fix: Fix Saving Branding form the second time - MEED-3151 - Meeds-io/meeds#1522 (#3389)

rdenarie:
Prior to this change, when saving the branding form the first time, it's not reset including Uploaded Files Ids which makes the second save operation fails. This change will fix the propagation of saving information to the parent in order to make the re-initialize the form.


azayati (approved):
LGTM

PR #3408, opened 2024-01-16
feat: Add quick filter toolbar in quick search drawer -  EXO-68495 - Meeds-io/MIPs#110

hakermi:
Add quick filter toolbar in quick search drawer and fix some ui issues


ahamdi (approved):
LGTM

PR #3417, opened 2024-01-19
feat: Fix default avatar in mention user popover- MEED-3111 - Meeds-io/MIPs#83

SaraBoutej:
This change allows to add a new property (enabled) to the identity-popover values to display the new default user avatar in the user popover when hovering on a mention user if the user is enabled.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
@@ -200,7 +200,7 @@ public class LinkProvider {
                .append(identity.getProfile().getProperty(Profile.EXTERNAL) != null && StringUtils.equals("true", String.valueOf(identity.getProfile().getProperty(Profile.EXTERNAL))))
                .append("',")
                .append("enabled: '")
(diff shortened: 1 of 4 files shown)


component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
```diff
@@ -199,6 +199,9 @@ public static String getProfileLink(String username, String portalOwner) {
                .append("external: '")
                .append(identity.getProfile().getProperty(Profile.EXTERNAL) != null && StringUtils.equals("true", String.valueOf(identity.getProfile().getProperty(Profile.EXTERNAL))))
                .append("',")
+               .append("enabled: '")
+               .append(identity.isEnable() && !identity.isDeleted() && StringUtils.equals("true", String.valueOf(identity.isEnable())))
```
boubaker: why parsing boolean to String again ?

```suggestion
               .append(identity.isEnable() && !identity.isDeleted())
```

SaraBoutej: removed!

boubaker (approved):

PR #3419, opened 2024-01-22
fix: For space avatar, aria-label on link does not contains the text - EXO-69059 - Meeds-io/meeds#1564

rdenarie:
Before this fix, the space avatar link has an incorrect aria-label. This label must contains the text of the link, and as it display 'Open the {prettyName}' instead of 'Open the {displayName}', this is shown as error in lighthouse This commit correctly use displayName instead of prettyname

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/common/components/SpaceAvatar.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/common/components/SpaceAvatar.vue
@@ -35,7 +35,7 @@
       v-on="on"
       :id="id"
       :href="url"
-      :arial-label="$t('space.avatar.href.title',{0:prettyName})"
+      :aria-label="$t('space.avatar.href.title',{0:displayName})"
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/common/components/SpaceAvatar.vue
```diff
@@ -181,6 +181,7 @@ export default {
   },
   computed: {
     displayName() {
+      console.log('compute displayName', this.space);
```
azayati: console log needed ?

azayati (changes requested):

webapp/portlet/src/main/webapp/vue-apps/common/components/SpaceAvatar.vue
rdenarie: nope, updated

azayati (approved):
LGTM

PR #3420, opened 2024-01-22
Meed 3111 merge

SaraBoutej:


component/core/src/main/java/org/exoplatform/social/core/image/ImageUtils.java
```diff
@@ -149,6 +154,56 @@ public static AvatarAttachment createResizedAvatarAttachment(InputStream imageSt
...
+    try {
+      ImageIO.write(image, "png", outputStream);
+      newAvatarAttachment = new AvatarAttachment(null,
+                                                 "DEFAULT_AVATAR",
+                                                 "image/png",
+                                                 new ByteArrayInputStream(outputStream.toByteArray()),
+                                                 System.currentTimeMillis());
+      return newAvatarAttachment;
+    } catch (IOException e) {
+      LOG.error("Fail to create file avatar : " + e);
```
boubaker: Should well log the exception not concatenating the exception object.

```suggestion
      LOG.warn("Fail to create default avatar for identity {}. Use default static avatar instead.", identityId, e);
```

SaraBoutej: done

boubaker (approved):

PR #3421, opened 2024-01-22
feat: Adjust the behavior when scrolling down the left menu -MEED-3169 - meeds-io/MIP#117 

MayTekayaa:
This change will adjust the behavior when scrolling down the left menu.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/HamburgerMenuNavigationFirstLevel.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/HamburgerMenuNavigationFirstLevel.vue
@@ -43,14 +43,14 @@
         :sticky-allowed="stickyAllowed"
         class="flex-grow-0 flex-shrink-0"
         @input="$emit('stickyPreference', $event)" />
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/HamburgerMenuNavigationFirstLevel.vue
```diff
@@ -138,5 +141,14 @@ export default {
       return this.stickyPreference && this.stickyAllowed;
     },
   },
+  mounted() {
+    window.addEventListener('scroll', this.onScroll);
```
boubaker:
```suggestion
    window.addEventListener('scroll', this.onScroll, false);
```

boubaker:
Sorry, I can't understand why we should use the JS here of something that we should resolve using CSS (more performant and more compatible)

MayTekayaa:
> Sorry, I can't understand why we should use the JS here of something that we should resolve using CSS (more performant and more compatible)

Done!

boubaker (approved):

PR #3432, opened 2024-01-24
feat: fix wallet identity fileItem error - MEED-3267 -Meeds-io/MIPs#83

SaraBoutej:
Prior to this change, on prod and pre-prod environments  we have a NPE when getting the identity avatar in wallet and reward configurations, This change allows to fix the NPE exception by testing if the object is null before the edit.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
@@ -482,7 +482,9 @@ public class RDBMSIdentityStorageImpl implements IdentityStorage {
       profile.setId(String.valueOf(entity.getId()));
       if (entity.getAvatarFileId() != null && entity.getAvatarFileId() > 0) {
         FileItem fileItem = fileService.getFile(entity.getAvatarFileId());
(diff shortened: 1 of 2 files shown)


component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
```diff
@@ -482,7 +482,9 @@ public Profile loadProfile(Profile profile) throws IdentityStorageException {
       profile.setId(String.valueOf(entity.getId()));
       if (entity.getAvatarFileId() != null && entity.getAvatarFileId() > 0) {
```
boubaker, line 483: Just a small question, may be even a unit test please:
`entity.getAvatarFileId() > 0` we shouldn't have an avatar generated for identities other than of type space or user. Is it the case for wallet ? If yes, could you check and fix the why + add a unit test that, by example for GroupIdentityProvider, no avatar file is generated/attached.

SaraBoutej: > we shouldn't have an avatar generated for identities other than of type space or user. Is it the case for wallet ?

Yes for identities other than USER or space we don't have generated default avatar, 





> Is it the case for wallet ?

yes, for wallet admin identity we haven't default avatar too

boubaker (approved):

PR #3437, opened 2024-01-25
Merge: Review the settings and logout options of the left menu - Meeds-io/MIPs#117

### LINKED ISSUE #117: Bottom Bar in Left menu
## Rationale

Now that the administration site is by default something we can access from the topbar, it would be interesting to review the access to the settings and to the logout.

Indeed, it is not a major access as any other pages (spaces, my craft, contribute, etc.) and it would be more interesting to list these options at the bottom of the left menu.
That way, only pages the user access the most are listed in first place.

### Top User Stories

Given I access the left menu
Then at the bottom, I can see three informations:
1. Meeds (or anything else depending the package*)
2. User Settings icon to access the user settings
3. Logout icon to log out from the platform

Given I have a small device OR/AND I have lots of spaces listed
Then the bottom of the menu is sticked so I can see these options 
AND I can scroll down when needed

Given I click to "Meeds"
Then I am sent to a link that we can configure. 
For meeds for example, it can be the product updates blog

image

Given I want to check the version
Then I can do it using rest services as currently

_***Label and link need to be extensible so we can update it depending on the package used**_

#### Gamification
NA

#### Notifications
NA

#### Analytics
Access to meeds version link

### Extensibility

The Package Link displayed in bottom left of the menu should be extensible so that it can be set by package.

### Existing Features

> [!NOTE]
> No change on current behavior of features, this is just a UI change

### Feature Flags

> [!NOTE]
> No feature flag is needed

### Access

Two properties should be available in frontend to be able to display information:
- `eXo.env.portal.productName` ([already exists](https://github.com/Meeds-io/meeds/blob/develop/services/plf-community-edition-service/src/main/resources/conf/platform.properties#L18))
- `eXo.env.portal.productLink` to add the same way as `eXo.env.portal.productName`, packaged in product artifacts which defines the product characteristics.

Both variables should be used to display the Product Label.

MayTekayaa:
This change will review the settings and the logout options of the left menu.

( Resolves Meeds-io/MIPs#117 )

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/user/UserHamburgerNavigation.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/user/UserHamburgerNavigation.vue
@@ -44,6 +44,7 @@
           <template #activator="{ on, attrs }">
             <v-btn
               :href="settingsUrl"
+              :aria-label="$t('menu.settings')"
               height="50px"
               min-width="40px"
               v-bind="attrs"
@@ -64,6 +65,7 @@
           <template #activator="{ on, attrs }">
             <v-btn
               :href="logoutUrl"
+              :aria-label="$t('menu.logout')"
               height="50px"
               min-width="40px"
               v-bind="attrs"
--- a/webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/user/UserHamburgerNavigation.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/user/UserHamburgerNavigation.vue
@@ -28,7 +28,7 @@
           <a
             :href="productLink"
             target="_blank"
-            class="body-2 text-sub-title my-auto ms-3"
+            class="body-2 text-sub-title my-auto"
             v-bind="attrs"
             v-on="on">
             {{ productName }}
@@ -48,7 +48,8 @@
               min-width="40px"
               class="px-0"
               v-bind="attrs"
-              v-on="on">
+              v-on="on"
+              icon>
               <v-icon size="22">
                 fa-sliders-h
               </v-icon>
@@ -59,7 +60,7 @@
           </span>
         </v-tooltip>
       </div>
-      <div class="logoutLinks">
+      <div class="logoutLinks me-n3">
         <v-tooltip top>
           <template #activator="{ on, attrs }">
             <v-btn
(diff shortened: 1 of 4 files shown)


webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/user/UserHamburgerNavigation.vue
```diff
@@ -19,31 +19,76 @@
...
+        <span>
+          {{ $t('menu.productName.seeProduct') }}
+        </span>
+      </v-tooltip>
+      <v-spacer />
+      <div class="settingsTitle">
+        <v-tooltip top>
+          <template #activator="{ on, attrs }">
+            <v-btn
+              :href="settingsUrl"
```
rdenarie, line 46: An aria-label attribute is missing to make the link in the button accessible.
Without that, there is an a11y error shown by lighthouse : 
image

webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/user/UserHamburgerNavigation.vue
```diff
@@ -19,31 +19,76 @@
...
+          <span>
+            {{ $t('menu.settings') }}
+          </span>
+        </v-tooltip>
+      </div>
+      <div class="logoutLinks me-n3">
+        <v-tooltip top>
+          <template #activator="{ on, attrs }">
+            <v-btn
+              :href="logoutUrl"
```
rdenarie, line 67: An aria-label attribute is missing to make the link in the button accessible.
Without that, there is an a11y error shown by lighthouse : 
image

rdenarie (changes requested):

webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/user/UserHamburgerNavigation.vue
MayTekayaa: done!

webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/user/UserHamburgerNavigation.vue
MayTekayaa: done!

rdenarie (approved):

PR #3438, opened 2024-01-25
feat: Non editable field ui/ux - EXO-68605 - Meeds-io/MIPs#111

hakermi:
Non editable field ui/ux: Display non editabel fields as disabled


ahamdi (approved):
LGTM

PR #3447, opened 2024-01-26
fix: Make users and space popovers controllable from the keyboard - EXO-68900

GouadriaHanen:
Before this change, users and spaces popovers were not displayed when navigating by the keyboard (TAB action) and did not close when tapping ESC.
after this change, the space and user popovers are displayed with the TAB when the TAB action is on the popover element and  it closes when the ESC key is taped

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/popover/components/PopoverMenu.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/popover/components/PopoverMenu.vue
@@ -23,6 +23,7 @@
     v-if="menu"
     v-model="menu"
     rounded="rounded"
+    role="tooltip"
     :close-on-content-click="false"
     :position-x="offsetX"
     :position-y="offsetY"
@@ -119,6 +120,12 @@ export default {
     document.addEventListener('drawerClosed', () => this.closePopover(true));
     document.addEventListener('modalOpened', () => this.closePopover(true));
     document.addEventListener('modalClosed', () => this.closePopover(true));
+    document.addEventListener('popover-identity-hide', () => this.closePopover(true));
+    window.addEventListener('keydown', (event) => {
+      if (event.key === 'Escape') {
+        this.closePopover(true);
+      }
+    });
     this.$root.$on('popover-hovered', this.setPopoverHovered);
     this.$root.$on('popover-not-hovered', this.setPopoverNotHovered);
   },
--- a/webapp/portlet/src/main/webapp/vue-apps/popover/main.js
+++ b/webapp/portlet/src/main/webapp/vue-apps/popover/main.js
@@ -25,10 +25,8 @@ Vue.directive('identity-popover', (el, binding) => {
   el.addEventListener('mouseover', () => {
     showPopover(el, identity, isUser);
   });
-  el.addEventListener('keyup', (event) => {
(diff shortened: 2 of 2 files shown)


rdenarie:
Can you add more details on the commit message ?

You should answer two questions : 

- why is this change needed ?
- how does this change address the issue.

Your current commit message only give information about what is the result. This is already explained in the commit title

webapp/portlet/src/main/webapp/vue-apps/popover/main.js
```diff
@@ -23,19 +23,31 @@ Vue.directive('identity-popover', (el, binding) => {
...
-        element: el,
-      }, identity || {})
-    }));
+    showPopover(el, identity, isUser);
+  });
+  el.addEventListener('keyup', (event) => {
+    if (event.key === 'Tab') {
+      showPopover(el, identity, isUser);
+    }
+  });
```
rdenarie: As I understand, when I type "Tab", all popovers will be displayed in same time, rigth ? 
It is not the expected.

We want to make popover part of the tabulation navigation.

rdenarie (changes requested):

webapp/portlet/src/main/webapp/vue-apps/popover/main.js
GouadriaHanen: the event listener is set on the element not on all the documents so it appears when the TAB on the element 
and closes when leaving the element with "focusout" event

rdenarie: Can we simply use events onfocus and onblur ?

I think theses standard event are preferable to a hardcoded key press

GouadriaHanen: done

webapp/portlet/src/main/webapp/vue-apps/popover/components/PopoverMenu.vue
```diff
@@ -119,6 +120,12 @@ export default {
     document.addEventListener('drawerClosed', () => this.closePopover(true));
     document.addEventListener('modalOpened', () => this.closePopover(true));
     document.addEventListener('modalClosed', () => this.closePopover(true));
+    document.addEventListener('popover-identity-hide', () => this.closePopover(true));
```
rdenarie, line 123: What is this event for ?

rdenarie (changes requested):

webapp/portlet/src/main/webapp/vue-apps/popover/components/PopoverMenu.vue
GouadriaHanen: for the focusout event we need to hide the popover

GouadriaHanen: https://github.com/Meeds-io/social/pull/3447/files#diff-49e191f2a3cc7bb6bf8168652aa3d5cc741da346cf329456db20bb9e8bc9aa5bR32

rdenarie: ok

rdenarie (approved):

PR #3458, opened 2024-01-31
fix: Add aria-label for settings and logout buttons - MEED-3286 - Meeds-io/MIPs#117 

MayTekayaa:
This change will add aria-label for settings and logout buttons.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/user/UserHamburgerNavigation.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/hamburger-menu/components/user/UserHamburgerNavigation.vue
@@ -44,6 +44,7 @@
           <template #activator="{ on, attrs }">
             <v-btn
               :href="settingsUrl"
+              :aria-label="$t('menu.settings')"
               height="50px"
               min-width="40px"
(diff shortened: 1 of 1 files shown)


extension/war/src/main/resources/locale/portal/HamburgerMenu_en.properties
```diff
@@ -22,4 +22,6 @@ menu.spaces.exploreSpaces=Explore Spaces
 menu.spaces.noSpacesFound=No spaces found
 menu.productName.seeProduct=See product updates
 menu.settings=Access User Settings
-menu.logout=Log out
\ No newline at end of file
+menu.logout=Log out
+menu.settings.label=Settings
+menu.logout.label=Logout
```
rdenarie: Why adding a new label as menu.logout is existing and the same one ?

MayTekayaa: Just to keep the old one for logout tooltip

MayTekayaa: Used the same labels for settings and logout

boubaker (approved):

PR #3479, opened 2024-02-05
feat : Initial Letters Avatars - Merge meeds-io/MIPs#83

### LINKED ISSUE #83: Initial Letters Avatars
## Rationale

Default avatars used for space and user are outdated.
It proposes something like a default icon which is not accurate with what other platform can propose to users.

### Top User Stories

* **User - Avatar proposed by default**

When logging in the first time, my avatar is set as followed:
a. round avatar with no border, and background color randomly chosen from a color palet of 15 colors
b. initials uppercase (white) and 2 letters max

When updating my name AND the default avatar is still on, then the avatar is updated accordingly

*Example:*
image

* **Space - Avatar proposed by default**

When creating a space, the avatar is set as followed:
a. square avatar with no border, and background color randomly chosen from the same color palet of 15 colors
b. initials uppercase (white) and 2 letters max chosen among 2 first words of the space name

When updating the name of the space, then the avatar is updated accordingly

*Example:*
image

* **Update of avatars elsewhere**
Of course, each time an update is done, it is done anywhere the avatar are displayed (notifications, search card, mention, popover, etc.)

No special change for the stored information. The new avatar should be stored the same way as when a user uploads his own avatar.

### Configurability 

The branding colors list has to be provided and modifiable using properties.

### Upgradability

This shouldn't be applied to old empty avatars, thus no upgrade plugin is necessary.

### Feature Flags

No Feature Flag is needed.

### Access 
   - No change on GUI nor on REST endpoint

### Services & processing

In order to make the default avatar generally applied including email notifications, the default avatar should be an image rather than a Vue component which displays initials when the avatar is a default one. To be able to generate images, we will use regular Java **AWT package** to generate the avatar.
The avatar generation has to happen in storage layer to make sure that retrieving the default data is **centralized** and **concurrently** secured. Thus, the method `RDBMSIdentityStorageImpl#getAvatarFile` has to be implemented in `CachedIdentityStorage#getAvatarFile` using a `FutureCache` which returns a `FileItem`. The default cache size should be small (500) with a small live time (300 seconds).

The business logic of `RDBMSIdentityStorageImpl#getAvatarFile` has to be implemented this way:

AvatarGenerationFlowChart

The default avatar generation process **shouldn't trigger** the listener of avatars update since it's a placeholder file generation and no action was made by users to attach avatars. Consequently, the gamification and analytics listeners shouldn't be triggered.

When a user or a space is renamed, if the identity has a default avatar, then a new default avatar has to be generated. In order to be able to know whether the attached avatar to a **space** or a **user** is a default generated one, a special `FileItem#FileInfo#name` has to be used with name `DEFAULT_AVATAR`. This information has to be added in Profile DTO `private boolean defaultAvatar`. The avatar update process must not trigger any listener as for default avatar creation since it's a placeholder, thus this update has to operate in Storage layer by simply deleting the associated `FileItem` and let the above described generation flow happens when retrieving the Space/User identity profile.

**hint**: In order to make the renaming detection easy to do, the `SpaceIdentityProvider#populateProfile` has to change in order to map `Profile.FULL_NAME` with `Space#displayName` as made in `OrganizationIdentityProvider#populateProfile`, so that we can compare the `DisplayName` and detect if we need to update the generated avatar or not (we may need to update `RDBMSSpaceStorageImpl#renameSpace` to add this mapping as well).

In term of UI, when the user provides a new name, it should refresh to display the new avatar without requesting to refresh the page.

SaraBoutej:


boubaker (approved):
[Validated by eXo](https://github.com/Meeds-io/MIPs/issues/83#issuecomment-1938374172)

PR #3480, opened 2024-02-06
fix: Make GroupSynchronizationSocialProfileListener Asynchronous - EXO-69333

mkrout:
Prior to this change, the group synchronization listener was executed in a synchronous way, so it may block the user profile update when we need to synchronize many profile fields since adding and removing users to groups may take a long time.
This commit will change the way the listener is executed so it will be in a separate thread.
The commit also avoids saving the profile for each property, so the listener will be executed once for all properties updated in the rest call


component/api/src/main/java/org/exoplatform/social/common/lifecycle/LifeCycleCompletionService.java
```diff
@@ -75,17 +77,11 @@ public LifeCycleCompletionService(InitParams params) {
...
-      this.executor = Executors.newFixedThreadPool(this.configThreadNumber);
-    }
-    else {
-      this.executor = new DirectExecutor();
-    }
+    this.synchronousExecutor = new DirectExecutor();
 
-    //
-    this.ecs = new ExecutorCompletionService(executor);
+    this.ecs = new ExecutorCompletionService(asynchronousExecutor);
```
rdenarie: Im not sure to understand here. As tasks are added with ecs.submit, I m afraid taht all task will be added as async.
But we want only listener with asynchronous annotation. Other must still be with DirectExectutor, no ?

component/core/src/main/java/org/exoplatform/social/core/listeners/GroupSynchronizationSocialProfileListener.java
```diff
@@ -102,7 +104,7 @@ private void profileUpdated(ProfileLifeCycleEvent profileLifeCycleEvent) {
     }
   }
 
-  private void synchronizePropertyGroups(Profile profile, User user) {
+  private static synchronized void synchronizePropertyGroups(Profile profile, User user) {
```
rdenarie: Why put all in static ? 
Im not sure it is the correct way to do

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -1121,8 +1118,8 @@ public Response updateUserProfileAttributes(@Context
         LOG.error("Error updating user {} attributes", currentUser, e);
         return Response.serverError().build();
       }
-
     }
+    identityManager.updateProfile(profile, getCurrentUser(), true);
```
rdenarie, line 1122: What if nothing change ? we save the profil event if there is no modification ?

rdenarie (changes requested):

component/core/src/main/java/org/exoplatform/social/core/listeners/GroupSynchronizationSocialProfileListener.java
mkrout: if we put just synchronized, it will be  applied only in the thread instance, for multi threading we need to make it static so we synchronize the execution of the calss instance.

component/api/src/main/java/org/exoplatform/social/common/lifecycle/LifeCycleCompletionService.java
mkrout: addTask is used by the AbstractLifeCycle only when the listener has the asynchronous annotation to create a new thread and dispatch the event. In the case of an asynch listener, AbstractLifeCycle will dispatch the event without creating a new thread, so it will be executed in the same thread.

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
mkrout: I juste put the updateProfile out of the loop, initially the rest service don't check if the property is changed or not, this check is done in the app level (client side). I'm not sure if this should be handled in this PR.

component/api/src/main/java/org/exoplatform/social/common/lifecycle/LifeCycleCompletionService.java
rdenarie: okWhat if lifecycleCompletionService is configured as async (async=true) ? 
Should we use direct executor ? 
Should we use asyncexectutor only with async=true AND Asynchronous annotation ?

component/core/src/main/java/org/exoplatform/social/core/listeners/GroupSynchronizationSocialProfileListener.java
rdenarie: We use synchronized nowhere in a listener, nor static in fact. 

I think we should manage the synchronized in the service level, but not in the listener. 
As we have only one instance of the service, using synchronized should be ok.

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
rdenarie: ok

rdenarie (changes requested):

component/api/src/main/java/org/exoplatform/social/common/lifecycle/LifeCycleCompletionService.java
mkrout: > what if lifecycleCompletionService is configured as async (async=true) ? 

if async is true all listeners will be asynchronous

> Should we use direct executor ? Should we use asyncexectutor only with async=true AND Asynchronous annotation ?

fyi I changed the code to only use Executors.newFixedThreadPool(), we don't need the directExecutor anymore.
so in case the listener is async, a new thread will be created; otherwise, the code will be executed in the same thread.

rdenarie: Let me understand :

Can you complete theses cases : 
- async = true : 
   - listener with @Asynchronous : same thread or new thread ?
   - listener without @Asynchronous : same thread or new thread ?
- async = false : 
   - listener with @Asynchronous : same thread or new thread ?
   - listener without @Asynchronous : same thread or new thread ?

mkrout: > Let me understand :
> 
> Can you complete theses cases :
> 
> * async = true :
>   
>   * listener with @asynchronous : same thread or new thread ?

New Thread
>   * listener without @asynchronous : same thread or new thread ?

New Thread
> * async = false :
>   
>   * listener with @asynchronous : same thread or new thread ?

New Thread
>   * listener without @asynchronous : same thread or new thread ?

Same Thread

rdenarie: ok for this part

component/core/src/main/java/org/exoplatform/social/core/listeners/GroupSynchronizationSocialProfileListener.java
mkrout: As discussed, putting the updateProfile out of the loop avoids the problem of concurrent changes in the organization service, so there is no need to synchronize the groups operation

rdenarie (approved):

PR #3501, opened 2024-02-12
fix: upgrade and reindex user profile ES index - EXO-64127 - Meeds-io/meeds#1631

ahamdi:
This fix will upgrade and reindex profile ES index using the mechanism created in ElasticIndexingProcessor


mkrout (approved):
LGTM

PR #3507, opened 2024-02-14
fix: fix add links issue with overlay effect - EXO-69307 (#3499)

GouadriaHanen:
Before this change, When the drawer parent was inside a container with a position: sticky, it was always displayed under the overlay and we could not interact with it. after this change, The fix changes the CSS property position when the drawer is opened and restores it to the initial value once closed.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/links/components/settings/LinkFormDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/links/components/settings/LinkFormDrawer.vue
@@ -204,7 +204,8 @@ export default {
           }
         }, 200);
       }
-    },
+      this.$el.closest('#stickyBlockDesktop').style.position = 'static';
+    }
   },
   created() {
     this.$root.$on('links-form-drawer', this.open);
--- a/webapp/portlet/src/main/webapp/vue-apps/links/components/settings/LinkSettingsDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/links/components/settings/LinkSettingsDrawer.vue
@@ -387,6 +387,13 @@ export default {
         this.stepper = 1;
       }
     },
+    drawer(){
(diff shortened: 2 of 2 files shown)


Jihed525 (approved):

mkrout (approved):

webapp/portlet/src/main/webapp/vue-apps/links/components/settings/LinkFormDrawer.vue
```diff
@@ -204,7 +204,8 @@ export default {
           }
         }, 200);
       }
-    },
+      this.$el.closest('#stickyBlockDesktop').style.position = 'static';
```
boubaker, line 207: @GouadriaHanen @rdenarie 

Should we add all links portlet inside containers having `stickyBlockDesktop` all time ? What if we add Links portlet outside `stickyBlockDesktop` container ? This container id exists in Activity Stream only if I remember well.

To fix the display of drawer when sticky, please use this workaround instead, which mounts the drawer outside the sticky block:

https://github.com/Meeds-io/gamification/blob/2.5.0/portlets/src/main/webapp/vue-app/usersLeaderboardCommon/components/ProgramSelectorDrawer.vue#L99-L101

```
  mounted() {
    document.querySelector('#vuetify-apps').appendChild(this.$el);
  },
```

rdenarie: @mkrout can you check the @boubaker 's proposition ? should we adapt the fix ?

Thanks

mkrout: I think yes, the Boubaker workaround would fix the problem no metter if we have a stikyBlock or not, @GouadriaHanen ccan you please create a new PR please?

PR #3548, opened 2024-02-22
fix : Adjust the space's navigation stream icon after the migration context - EXO-69389 - Meeds-io/meeds#1677

sofyenne:
Prior to this change, after updating the space navigation upgrade plugin type to resolve the PostgreSQL issue, the stream navigation icon was not updated following the migration context. This issue arose due to the query which did not retrieve the stream navigation correctly for updating it. This change will update the query to set the stream navigation icon and update the plugin configuration to be re-executed, resolving this issue

### DIFF
--- a/component/core/src/main/java/io/meeds/social/core/upgrade/SpaceNavigationIconUpgradePlugin.java
+++ b/component/core/src/main/java/io/meeds/social/core/upgrade/SpaceNavigationIconUpgradePlugin.java
@@ -41,7 +41,7 @@ public class SpaceNavigationIconUpgradePlugin extends UpgradeProductPlugin {
                                                            UPDATE PORTAL_NAVIGATION_NODES
                                                            SET ICON =
                                                              CASE
-                                                               WHEN (SELECT pn.NAME FROM (SELECT * FROM PORTAL_NAVIGATION_NODES) pn WHERE pn.NODE_ID = PARENT_ID) = 'default' THEN TRIM('fas fa-stream')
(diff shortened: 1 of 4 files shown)


component/core/src/main/java/io/meeds/social/core/upgrade/SpaceNavigationIconUpgradePlugin.java
```diff
@@ -77,14 +80,23 @@ public boolean shouldProceedToUpgrade(String newVersion, String previousVersion)
...
   }
 
+  @Override
+  public boolean shouldProceedToUpgrade(String newVersion,
+                                        String previousGroupVersion,
+                                        UpgradePluginExecutionContext previousUpgradePluginExecution) {
+
+    int executionCount = previousUpgradePluginExecution == null ? 0 : previousUpgradePluginExecution.getExecutionCount();
+    return !isExecuteOnlyOnce() || executionCount == 0;
+  }
```
azayati: I think that this is useless since the UP is renamed, it will be executed for the first time

azayati (changes requested):

component/core/src/main/java/io/meeds/social/core/upgrade/SpaceNavigationIconUpgradePlugin.java
sofyenne: done !

mkrout (dismissed):

azayati (approved):
LGTM

PR #3595, opened 2024-03-07
fix: Adjust the space's navigation stream icon after the migration context - EXO-69389 - Meeds-io/meeds#1677

sofyenne:
Prior to this change, after updating the space navigation upgrade plugin type to resolve the PostgreSQL issue, the stream navigation icon was not updated following the migration context. This issue arose due to the query which did not retrieve the stream navigation correctly for updating it. This change will update the query to set the stream navigation icon and update the plugin configuration to be re-executed, resolving this issue


azayati (approved):
LGTM

PR #3607, opened 2024-03-11
fix: Disallow attachment of non-image types in task description - EXO-70180 - Meeds-io/meeds#1767

sofyenne:
Prior to this change , attaching a non-image type to a task description by the task description rich editor image attachment button was possible , this change is going to prevent the non-image type attachment from being attached to a task description .


mkrout (approved):
LGTM

PR #3612, opened 2024-03-13
fix: fix UI glitch in activity reactions when mobile - MEED-3908 - Meeds-io/meeds#1747

SaraBoutej:


webapp/portlet/src/main/webapp/vue-apps/activity-reactions/components/ActivityReactions.vue
```diff
@@ -101,8 +101,8 @@ export default {
     displayAnimation() {
       return this.showAvatarAnimation;
     },
-    large() {
-      return this.$vuetify.breakpoint.lg;
+    isDesktop() {
+      return this.$vuetify.breakpoint.width >= this.$vuetify.breakpoint.mobileBreakpoint;
```
boubaker: The work made in Meeds-io/MIPs#113 has made changes to make the breakpoint detection smother. Can you use `!this.$vuetify.mobile`. If it doesn't work, the there is a bug in provided breakpoints of Vuetify.

SaraBoutej: Ok, I'll try with !this.$vuetify.mobile
but I think there is a bug in vuetify breakpoints because I tried with lgAndUp and doesn't' work

SaraBoutej: it's okey for !this.$vuetify.breakpoints.mobile

SaraBoutej: test

boubaker (approved):

PR #3614, opened 2024-03-13
fix: Disallow attachment of non-image types in task description - EXO-70180 - Meeds-io/meeds#1767

Jihed525:
Prior to this change , attaching a non-image type to a task description by the task description rich editor image attachment button was possible , this change is going to prevent the non-image type attachment from being attached to a task description .

(cherry picked from commit d6662686910fa5a206e4be3219d6c42fa8fde2e9)

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/attach-image/components/form/ImageInputMultiUpload.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/attach-image/components/form/ImageInputMultiUpload.vue
@@ -77,6 +77,9 @@ export default {
         return;
       } else {
         filesArray.forEach((file) => {
+          if (!file?.type?.includes('image/')) {
+            return;
+          }
           if (file.size > this.maxFileSize) {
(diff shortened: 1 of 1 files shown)


GouadriaHanen (approved):

mkrout (approved):

webapp/portlet/src/main/webapp/vue-apps/attach-image/components/form/ImageInputMultiUpload.vue
```diff
@@ -77,6 +77,9 @@ export default {
         return;
       } else {
         filesArray.forEach((file) => {
+          if (!file?.type?.includes('image/')) {
```
boubaker, line 80: to use the `?.` and to ensure that it's efficient, you should use it on all elements

```suggestion
          if (!file?.type?.includes?.('image/')) {
```

Jihed525: I will update it @boubaker

PR #3633, opened 2024-03-21
fix: Link Node Type : redirection target ko on left navigation menu drawer space details - EXO-70359 - Meeds-io/meeds#1790

GouadriaHanen:
Before this change, in the left navigation drawer space node details, when a node is of link type and have a new page as target, on click blank page is opened in the same tab.
This change applies the target option to the space nodes drawer since it was not treated


azayati (changes requested):
please explain more the problem/solution for the commit message, and modify the title (left space navigation menu and not space drawer)

azayati (approved):
LGTM

PR #3643, opened 2024-03-27
feat: Implement getMetadataItemsByMetadataNameAndTypeAndObjectAndSpaceIds service - EXO-70405

azayati:


component/api/src/main/java/org/exoplatform/social/metadata/MetadataService.java
```diff
@@ -391,13 +390,12 @@ default List<MetadataItem> getMetadataItemsByMetadataNameAndTypeAndSpaceIds(Stri
...
    * @return {@link List} of linked {@link MetadataItem}
    */
-  default List<MetadataItem> getMetadataItemsByMetadataNameAndTypeAndObjectAndSpaceIdsOrByCreatorId(String metadataName,
-                                                                                                    String metadataTypeName,
-                                                                                                    String objectType,
-                                                                                                    List<Long> spaceIds,
-                                                                                                    Long creatorId,
-                                                                                                    long offset,
-                                                                                                    long limit) {
+  default List<MetadataItem> getMetadataItemsByMetadataNameAndTypeAndObjectAndSpaceIds(String metadataName,
```
boubaker, line 393: Just a proposition, we may need to define an object (`Filter`) where you can add a `set` of attributes to be use to filter metadata instead of adding a list of methods to define the possible combinaison of attributes that we can filter.

azayati: @boubaker I have removed the creatorId, no need to that, I just added the method in service layer to retrieve metadata items corresponding to a list of spaces, the called storage layer method is already existing

sofyenne (approved):

PR #3664, opened 2024-04-03
fix : extension point in space top bar popover does not display extensions the second time - EXO-70575

### LINKED ISSUE #1839: In space top bar popover, a specific extension is not correctly loaded
**Steps to reproduce**
You need to have webconferencing, so the problem is visible on eXo Platform version

1. Go to a space home page
2. hover the space name in the top bar - the visio button is correctly displayed
3. Move out to close the popup
4. Hover the space name a second time 

**Current behaviour**

The visio button is not present the second time

rdenarie:
Before this fix, when hovering the space name in top bar, the last extension point 'space-title-action-components' does not load correctly On first hover, it is displayed, on second it is not.

This commit force the reload of the extension to make is displayed each times

Resolves meeds-io/meeds#1839


azayati (approved):
LGTM

PR #3665, opened 2024-04-03
fix : extension point in space top bar popover does not display extensions the second time - EXO-70575

rdenarie:
Before this fix, when hovering the space name in top bar, the last extension point 'space-title-action-components' does not load correctly On first hover, it is displayed, on second it is not.

This commit force the reload of the extension to make is displayed each times

Resolves meeds-io/meeds#1839


azayati (approved):
LGTM

PR #3669, opened 2024-04-04
feat: Add chart organizational chart  CTA button option in user popover - EXO-69210 - Meeds-io/MIPs#112

hakermi:
Add organizational chart CTA button option in user popover


ahamdi (approved):
LGTM
We will rework it later

PR #3690, opened 2024-04-17
fix: In topbar menu the link Node Type redirection ko if '/#/' is included - EXO-70678 - Meeds-io/meeds#1815 (#3687)

mkrout:
Before this change, When the href attribute contains a "#" followed by a fragment identifier, the browser treats it as an anchor link and tries to navigate to that fragment within the current page, instead of navigating to a different URL
After this change, the redirection will be done by a OnClick event, without removing the href attribute.


azayati (approved):
LGTM

PR #3703, opened 2024-04-19
fix: ui/ux and accessibility issues in user card - EXO-71144 -Meeds-io/MIPs#127

hakermi:
ui/ux and accessibility issues in user card :
- fix user role icons titles on hover
- fix accessibility labels in user card action buttons
- some code cleanup


webapp/portlet/src/main/webapp/vue-apps/people-list/components/usercard/PeopleUserCompactCard.vue
```diff
@@ -197,9 +197,6 @@ export default {
     }
   },
   computed: {
-    filteredUserNavigationExtensions() {
```
ahamdi: This computed property should not be removed here , no?

hakermi: yes it' should n't

ahamdi (approved):

PR #3736, opened 2024-04-29
feat: parse correctly the value of the manager from Active Directory when it is single-value - EXO-69802 - Meeds-io/MIPs#112

ahamdi:
the property manager is multi-value when users are stored in DB or in LDAP. But it is single value for Active directory.
This fix makes sure to parse it correctly to get the value and convert it to the selected user.


hakermi (approved):

azayati (dismissed):
need rebase IMO

hakermi (approved):

PR #3738, opened 2024-04-30
fix: Update people card profile action extension point to be able to load a vue component as extension - EXO-71271 - Meeds-io/MIPs#127

rdenarie:
Before this fix, the extension point in user card for profile action, is only able to display and icon, with a onclick action. In some cases, we need to load a complete vue component. This fix update the extension point to able to load such component


rdenarie:
@ahamdi as this commit is part of MIP#127, dont forget to taking it when backporting in develop branch.
Thanks

ahamdi (approved):

PR #3754, opened 2024-05-02
feat: Allow metadata filter to combine properties with or condition - EXO-70705 - Meeds-io/MIPs##130

hakermi:
Allow metadata filter to combine properties with or condition


azayati (approved):
LGTM

PR #3760, opened 2024-05-03
feat: fix see more display when blockquote - MEED-5901 - Meeds-io/MIPs#124

SaraBoutej:
Prior to this change, there is a UI glitch in the 'see more' label within the activity content when an HTML blockquote is displayed.
This change fix this UI issue.


boubaker:
Will be treated apart in a dedicated Task, thus you can close this PR for now.

PR #3766, opened 2024-05-06
fix: do not allow a multivalued property to be a parent property too - EXO-69915 - Meeds-io/meeds#112

ahamdi:
Multi-value properties should not be parent properties too.
The fix will not add a multivalue property to the list of parent when adding or editing a profile property.


hakermi:
rebase needed

PR #3771, opened 2024-05-07
feat: Allow metadata filter to fetch depends on multiple properties within same item -EXO-70705 - Meeds-io/MIPs#130

hakermi:
Allow metadata filter to fetch depends on multiple properties within same item


azayati (approved):
LGTM

PR #3775, opened 2024-05-08
fix: Fix include properties in the metadata filter query when spaceIds list is empty - EXO-70706 - Meeds-io/MIPs#130

hakermi:
fix include properties in the metadata filter query when spaceIds list is empty


azayati (approved):
LGTM

PR #3778, opened 2024-05-09
feat: Allow metadata filter to fetch depends on multiple properties within same item - EXO-70705 - Meeds-io/MIPs#130

hakermi:
Allow metadata filter to fetch depends on multiple properties within same item
Allow metadata filter to combine properties with or condition
fix include properties in the metadata filter query when spaceIds list
is empty


azayati (approved):
LGTM

azayati (approved):

PR #3783, opened 2024-05-10
fix: user card settings are not retrieved for simple users - Meeds-io/meeds#127 - EXO-71337

ahamdi:
Before this fix, we used the Setting Rest service to load the user card properties. This didn't work for simple users as this Rest service does not return settings when the Context is Global.
The fix creates a new Rest endpoint to return the card settings from User Rest service

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
@@ -32,6 +32,10 @@ import javax.ws.rs.core.Response.ResponseBuilder;
 import javax.ws.rs.core.Response.Status;
 
 import org.apache.commons.lang3.StringUtils;
+import org.exoplatform.commons.api.settings.SettingService;
+import org.exoplatform.commons.api.settings.SettingValue;
+import org.exoplatform.commons.api.settings.data.Scope;
+import org.exoplatform.settings.rest.SettingEntity;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.picocontainer.Startable;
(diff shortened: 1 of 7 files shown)


webapp/portlet/src/main/webapp/vue-apps/common/js/UserService.js
```diff
@@ -435,3 +435,16 @@ export function multiSelectAction(action, selectedUsers) {
...
   });
 }
+
+export function getUserCardSettings() {
+  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/v1/social/users/userCardSettings`, {
+    method: 'GET',
+    credentials: 'include',
+  }).then((resp) => {
+    if (!resp || !resp.ok) {
+      throw new Error('Error when updating users');
```
hakermi: Error while getting user card settings

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRestResourcesV1.java
```diff
@@ -1525,6 +1535,42 @@ public Response importUsers(@Context HttpServletRequest request,
     return errorResponse == null ? Response.noContent().build() : errorResponse;
   }
 
+  @GET
+  @Path("userCardSettings")
+  @Produces(MediaType.APPLICATION_JSON)
+  @RolesAllowed("users")
+  @Operation(
+          summary = "Gets the filed settings of a user card",
+          description = "Gets the filed settings of a user card",
```
hakermi: field

Christyempire: Okay so this code are not really clear to understand 

Please specify what we are to do here

Christyempire:
These codes are excellently written 
Kudos 👏 💐

hakermi (approved):

Christyempire:
How can user card be retrieve please 🙏

Christyempire:
Can card be retrieved

PR #3784, opened 2024-05-10
fix: enable the 3 dots menu for the current ser - Meeds-io/meeds#127 - EXO-71312

ahamdi:
Before this fix, the 3-dots menu was not displayed for the current user. 
The fix enables the 3-dots menu and the user can open his Org chart page from his user card.


hakermi (approved):

Christyempire:
Yeah I think the 3 dot will be helpful to the community builders

PR #3785, opened 2024-05-10
fix: allow to define email as a field in the user card - Meeds-io/meeds#127 - EXO-71336

ahamdi:
Allow the email field to be listed with proposed fields for the user card.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/ProfileSettings.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/ProfileSettings.vue
@@ -52,7 +52,7 @@
     <user-card-settings-drawer
       ref="userCardSettings"
       :user="user"
-      :settings="filteredSettings"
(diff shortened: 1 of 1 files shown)


hakermi:
By doing this the admin will see weird settings listed. Those are saved in gatein profile specialy in the tribe.
I'm telling this to be aware of.

hakermi (approved):

Christyempire:
Emails are the user cards of course it will be allowed

ahamdi:
> By doing this the admin will see weird settings listed. Those are saved in gatein profile specialy in the tribe. I'm telling this to be aware of.

@hakermi Thanks for the review. Well, all fields (espaecially technical ones) will be created as hidden, and administrators will have to make them visible if needed. Thus, it is the role of administrators to make just needed fields appear in the list.
Unhiddenable properties is the list of properties that are displayed by default, thus we should remove them. Otherwise, all other non parent/multivalued and visible properties will eb available to select for the three fields of the user card.

PR #3786, opened 2024-05-10
fix: disable invisible fields from Edit Profile Contact - Meeds-io/meeds#110 - EXO-70160

ahamdi:
If a field is hidden by an administrator, the user should not be able to re-hide it , and the Hide button should be disabled.


Christyempire:
Please can someone give directions on what you do here

hakermi (approved):

Christyempire:
Agreed 👍

Christyempire:
If the field is invisible how can we access our profile 🤷‍♂️

srenault-meeds:
> If the field is invisible how can we access our profile 🤷‍♂️

Please use the [forum](https://github.com/orgs/Meeds-io/discussions) for such questions and avoid adding off-topic comments to work we are doing. Thanks

PR #3796, opened 2024-05-15
fix: hide org chart pages when manager field became inactive - Meeds-io/MIPs#112 - EXO-71333

ahamdi:
When the manager field is inactive, we need to hide the MyTeam and Organizational chart pages.
This fix introduces a listener for changes on Profile property settings to hide/show those pages when the manager field is updated. It uses pages permission to perform this action.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/listeners/ManagerPropertySettingUpdatedListener.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/listeners/ManagerPropertySettingUpdatedListener.java
@@ -1,0 +1,57 @@
+package org.exoplatform.social.core.listeners;
+
+import org.exoplatform.commons.utils.CommonsUtils;
+import org.exoplatform.container.PortalContainer;
+import org.exoplatform.container.component.RequestLifeCycle;
+import org.exoplatform.portal.config.UserACL;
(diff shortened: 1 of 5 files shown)


component/core/src/main/java/org/exoplatform/social/core/listeners/ManagerPropertySettingUpdatedListener.java
```diff
@@ -0,0 +1,58 @@
...
+  public ManagerPropertySettingUpdatedListener(PageStorage pageStorage, UserACL userACL) {
+    this.pageStorage = pageStorage;
+    this.userACL = userACL;
+  }
+
+  @Override
+  public void onEvent(Event<ProfilePropertyService, ProfilePropertySetting> event) throws Exception {
+    ProfilePropertySetting propertySetting = event.getData();
+    if ("manager".equalsIgnoreCase(event.getData().getPropertyName())) {
+      RequestLifeCycle.begin(PortalContainer.getInstance());
```
hakermi: IMO we can use @ContainerTransactional in class level instead of requestlifecyclebegin and end()
wdyt ?

and we need to add the header

hakermi (commented):
LGTM

hakermi (approved):

PR #3799, opened 2024-05-16
fix: Update linkedin oauth to use openid endpoints - EXO-71792

rdenarie:
Before this fix, the linkedin oauth was not working because it use oauth url to get user info, and linkedin deactivate it in profit of openid url (/v2/userinfo) The commit update url to get user info, to make linkedin oauth implementation working

Resolves meeds-io/meeds#1996


mkrout (approved):

Treasureth:
Great job here

PR #3809, opened 2024-05-20
feat: External link must be opened in a new tab - MEED-6879 - Meeds-io/MIPs#124

SaraBoutej:
Before this change when we copy/paste an external link in the rich editor no target added to to link, so the link opened in the same tab.
This PR allows to add target _blank to external links to open in a new tab

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/common/components/RichEditor.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/common/components/RichEditor.vue
@@ -592,14 +592,19 @@ export default {
         tempdiv.innerHTML = content;
         const links = tempdiv.getElementsByTagName('a');
         for (const link of links) {
-          if (!link.href.includes(eXo.env.portal.context) && !link.hasAttribute('target')) {
+          if (link.href.indexOf(window.location.origin) === -1) {
             link.setAttribute('target', '_blank');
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/common/components/RichEditor.vue
```diff
@@ -587,7 +587,19 @@ export default {
         content = content.replace(/<oembed>(.*)<\/oembed>/g, '');
         content = content.replace(/<oembed>(.*)<\/oembed>/g, `<oembed>${oembedUrl}</oembed>`);
       }
+      if (content.includes('<a') && content.includes('</a>')) {
+        const tempdiv = document.createElement('div');
+        tempdiv.innerHTML = content;
+        const links = tempdiv.getElementsByTagName('a');
+        for (const link of links) {
+          if (!link.href.includes(eXo.env.portal.context) && !link.hasAttribute('target')) {
```
boubaker: You may need to use the same attributes and `origin` comparaison as https://github.com/Meeds-io/social/blob/develop/webapp/portlet/src/main/webapp/WEB-INF/gatein-resources.xml#L2439-L2445

SaraBoutej: done

boubaker (approved):

PR #3823, opened 2024-05-23
feat: Fix Purify Lib regression - MEED-6928 - Meeds-io/MIPs#124

SaraBoutej:
Prior to this change, there was a console JS error related to the DOM Purify library. 
This code fixes that error.

### DIFF
--- a/webapp/portlet/src/main/webapp/js/ExtendedDomPurify.js
+++ b/webapp/portlet/src/main/webapp/js/ExtendedDomPurify.js
@@ -53,7 +53,7 @@
         const nodeText = node.textContent;
         const nodeLink = node.getAttribute('href');
         // add noopener attribute to external links to eliminate vulnerabilities
-        if (nodeLink.indexOf(window.location.origin) === -1) {
+        if (nodeLink && nodeLink.indexOf(window.location.origin) === -1) {
           node.setAttribute('target', '_blank');
           node.setAttribute('rel', 'nofollow noopener noreferrer');
         }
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/js/ExtendedDomPurify.js
```diff
@@ -53,7 +53,7 @@
         const nodeText = node.textContent;
         const nodeLink = node.getAttribute('href');
         // add noopener attribute to external links to eliminate vulnerabilities
-        if (nodeLink.indexOf(window.location.origin) === -1) {
+        if (nodeLink && nodeLink.indexOf(window.location.origin) === -1) {
           node.setAttribute('target', '_blank');
           node.setAttribute('rel', 'nofollow noopener noreferrer');
         }
```
boubaker: Please move ellipsis processing only when link has a `href`

```suggestion
        if (nodeLink) {
          // Open external links in a new Browser Tab
          if (nodeLink.indexOf(window.location.origin) === -1) {
            node.setAttribute('target', '_blank');
            node.setAttribute('rel', 'nofollow noopener noreferrer');        
          }
          // add text ellipsis when link length is up to 75 characters
          if (nodeText && nodeText.length > 75) {
            node.setAttribute('title', nodeLink);
            node.setAttribute('Aria-label', nodeLink);
            node.textContent = node.textContent.substring(0,75)+'...';
          }
        }
```

SaraBoutej: done

boubaker (approved):

PR #3831, opened 2024-05-27
feat : Add 'updatedDate' filed to the metadataItem entity and enable sorting by this field - EXO-71874 - Meeds-io/MIPs#130

sofyenne:


component/core/src/main/java/org/exoplatform/social/core/jpa/storage/entity/MetadataItemEntity.java
```diff
@@ -296,6 +299,10 @@ public void setCreatedDate(Date createdDate) {
     this.createdDate = createdDate;
   }
 
+  public Date getUpdatedDate() { return updatedDate; }
+
+  public void setUpdatedDate(Date updatedDate) { this.updatedDate = updatedDate; }
```
azayati:
```suggestion
  public Date getUpdatedDate() { 
    return updatedDate; 
  }

  public void setUpdatedDate(Date updatedDate) { 
    this.updatedDate = updatedDate; 
  }
```

Format code

azayati (changes requested):

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/entity/MetadataItemEntity.java
sofyenne: done!

azayati (approved):
LGTM

PR #3844, opened 2024-05-30
fix: check for nullity before getting the value of Card settings - Meeds-io/meeds#2070 - EXO-72158

ahamdi:
Before this fix, when loading the user card settings, a NPE is thrown if those settings are not yet saved by administrators.
The fix checks for the nullity of those settings and returns the default values in that case.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/drawers/UserCardSettingsDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/drawers/UserCardSettingsDrawer.vue
@@ -64,7 +64,7 @@
             outlined
             @blur="$refs.firstField.blur();" />
         </label>
-        <label for="firstField">
+        <label for="secondField">
           {{ $t('profileSettings.userCard.settings.secondField.label') }}
           <v-select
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/drawers/UserCardSettingsDrawer.vue
```diff
@@ -64,7 +64,7 @@
             outlined
             @blur="$refs.firstField.blur();" />
         </label>
-        <label for="firstField">
+        <label for="secondField">
```
rdenarie, line 67: what if firstField or secondField or thirdField is not existing on front side ?

hakermi (approved):

webapp/portlet/src/main/webapp/vue-apps/profile-settings/components/drawers/UserCardSettingsDrawer.vue
ahamdi: @rdenarie those are the ids of the three selects. they should be available in the front . I may not understand your question

PR #3846, opened 2024-05-31
feat : Add 'updatedDate' filed to the metadataItem entity and enable sorting by this field - EXO-71874 - Meeds-io/MIPs#130

sofyenne:


azayati (approved):
LGTM

PR #3848, opened 2024-05-31
fix: Notification are updated even if nothing modified - EXO-72161

rdenarie:
Before this fix, when updating a notification, we do not check if something change. This can lead to a notification refresh on front side (time updated, notification go in the top of the list), even if nothing change into. This commit add a check and persist the change only if it is needed

Resolves meeds-io/meeds#2075


boubaker (commented):
Sorry, this doesn't happen in meeds, do you know why it happens on eXo only ?

rdenarie:
It is reproduced on version [plf-community-tomcat-standalone-7.0.x-meed-20240531.103535-204.zip](https://repository.exoplatform.org/content/groups/public/io/meeds/distribution/plf-community-tomcat-standalone/7.0.x-meed-SNAPSHOT/plf-community-tomcat-standalone-7.0.x-meed-20240531.103535-204.zip)

boubaker (approved):
Sorry, I thought it was about the notification date update that is perceptible on Tribe (not even eXo ACC acceptances). Thanks for the fix and explanation.

PR #3851, opened 2024-06-03
fix: Handle NullPointerException for the metadataItem updatedDate field - EXO-72221 - Meeds-io/MIPs#130

sofyenne:


azayati (approved):
LGTM

PR #3863, opened 2024-06-05
fix: display dynamically complementary filters based on drawer width - Meeds-io/meeds#2102 - EXO-72210

ahamdi:
before this fix, suggested filters were displayed on 2 lines when they contain a long text and does not fit all the space when the drawer is expanded.
The fix will calculate the available space and fit the number of displayed filters to the whole space when the drawer is collapsed or expanded.


hakermi (commented):
There are some console.logs to be removed

ahamdi:
> There are some console.logs to be removed

Done

hakermi (approved):

PR #3868, opened 2024-06-06
fix: Editing an activity with files removes the mention - MEED-6898 - Meeds-io/meeds#2029

SaraBoutej:
Before this change, editing an activity attachment without changing the activity text impacted user mentions.
This fix ensures that when the editor is ready, the suggester's HTML is automatically converted to text.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/common/components/RichEditor.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/common/components/RichEditor.vue
@@ -211,7 +211,6 @@ export default {
   watch: {
     inputVal() {
       this.updateInput(this.inputVal);
-
       if (this.supportsOembed) {
         this.setOembedParams({
           default_title: this.getContentToSave(this.inputVal),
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/common/components/RichEditor.vue
```diff
@@ -409,6 +408,7 @@ export default {
               });
 
             self.setEditorReady();
+            self.inputVal = textValue && self.getContentToSave(textValue);
```
boubaker, line 411: Please change `getContentToEdit` instead

boubaker: Please change `getContentToEdit` instead
```suggestion
```

SaraBoutej: fixed

boubaker:
As agreed will have to fix `getContentToEdit` instead

boubaker (approved):

PR #3914, opened 2024-06-19
fix: hide org chat button for the card of external users - Meeds-io/meeds#2209 - EXO-72074

ahamdi:
When checking external users in the space members page, or hovering over their names in activities, those users should not have the Organizational chart button.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/organizational-chart/extensions/extensions.js
+++ b/webapp/portlet/src/main/webapp/vue-apps/organizational-chart/extensions/extensions.js
@@ -5,7 +5,7 @@ export function registerExtension(title) {
     icon: 'fas fa-sitemap',
     class: 'fas fa-sitemap',
     order: 10,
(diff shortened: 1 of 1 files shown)


webapp/portlet/src/main/webapp/vue-apps/organizational-chart/extensions/extensions.js
```diff
@@ -5,7 +5,7 @@ export function registerExtension(title) {
     icon: 'fas fa-sitemap',
     class: 'fas fa-sitemap',
     order: 10,
-    enabled: (user) => eXo.env.portal.isExternal === 'false' && user.enabled,
+    enabled: (user) => eXo.env.portal.isExternal === 'false' && user.enabled && (!user.external || user.external === 'false'),
```
hakermi:
```suggestion
    enabled: (user) => eXo.env.portal.isExternal === 'false' && user.enabled && user?.external !==  'true',
```

hakermi (approved):

PR #3924, opened 2024-06-24
fix: [ExoDrawer component] When using attribute attached on a second level drawer, the drawer is not attached to the application

### LINKED ISSUE #2226: [ExoDrawer component] When using attribute attached on a second level drawer, the drawer is not attached to the application
When using vue component ExoDrawer as a second level drawer, with attached paramater, the drawer is not attached to the application.

rdenarie:
Before this fix, when the ExoDrawerComponent is used with attached parameter, with a second level drawer, the drawer is not attached to the application. This is due to a test which verify if there is more that one opened drawer. This test comes from olf part of the code and is no more necessary.

Resolved meeds-io/meeds#2226

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/common/components/ExoDrawer.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/common/components/ExoDrawer.vue
@@ -261,7 +261,7 @@ export default {
   },
   methods: {
     open() {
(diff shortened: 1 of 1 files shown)


rdenarie:
This PR is related to the fix in Time Tracking application

boubaker (approved):

PR #3927, opened 2024-06-25
correction : modification de l'UX/UI de la carte d'espace - EXO-72642- Meeds-io/meeds#2212.

akhanfir:
Before this change, when creating a spaces and go to spaces interface and check its card, members count is displayed and there is a margin between the bottom of the card and join/leave button. After this change, remove the info related to the space members number and set the join/leave button at the bottom of the card.


mkrout (approved):

ahamdi:
@akhanfir @mkrout commit message should be in English :smile:

PR #3972, opened 2024-08-12
Right to left display of ko in admin interface and some labels are in English instead of Arabic - EXO-73560 - Meeds-io/meeds#2308.

akhanfir:
Before this change, when you set the language to Arabic as a plf admin and access the admin site, the platform settings menu is positioned on the left and is not fully displayed due to an overlap with the main settings block. After this change, the right to left display in the platform settings menu without overlap is displayed correctly.

### DIFF
--- a/webapp/portlet/src/main/webapp/vue-apps/vertical-menu/components/VerticalMenuApp.vue
+++ b/webapp/portlet/src/main/webapp/vue-apps/vertical-menu/components/VerticalMenuApp.vue
@@ -16,7 +16,7 @@
 -->
 <template>
   <v-app>
-    <vertical-menu-content v-if="fullDisplayAllowed" extra-class="VerticalMenu" />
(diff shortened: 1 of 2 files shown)


webapp/portlet/src/main/webapp/vue-apps/vertical-menu/components/VerticalMenuApp.vue
```diff
@@ -29,6 +29,9 @@ export default {
     fullDisplayAllowed() {
       return this.$vuetify.breakpoint.width >= this.$vuetify.breakpoint.thresholds.lg;
     },
+    isLanguageArabic() {
```
mkrout: IMO, it will better to use eXo.env.portal.orientation to check the text orientation, then testing the laguage

mkrout (approved):

PR #4067, opened 2024-09-26
feat: Upgrade Spring and replace Jcabi Plugin by AspectJ Maven Plugin - Meeds-io/meeds#2417

boubaker:
Prior to this change, the jcabi maven plugin was used to weave classes for annotation processing, especially  and  annotations. This plugin isn't maintained and should be replace by the default AspectJ libraries for Annotation Processing. In addition, the Jcabi plugin makes the build longer and takes a lot of time in addition to heigh I/O operations on disk. This change applies the required configuration (In addition to https://github.com/Meeds-io/maven-parent-pom/commit/bea09a8fcbc02b87ee8748cd9bbf9a524e4d74cd) in order to replace usage of Jcabi plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -206,7 +206,6 @@
   </dependencyManagement>
 
   <build>
-    <finalName>${project.artifactId}</finalName>
     <pluginManagement>
       <plugins>
         <plugin>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -235,6 +234,41 @@
 
   <!-- This profile is used to allow github action to build branches. The github action is used for sonar analysis -->
   <profiles>
+    <profile>
```
rdenarie, line 237: why we need to define the profile here as it is already in maven-parent-pom ?

boubaker: It's to declare the `AspectJ custom Annotation processing` defined in module `portal-component-common` to be used when weaving/processing the class file, here is the main declaration needed for this profile ( **aspectLibraries** not a `<dependency>` ):

```xml
              <aspectLibraries>
                <aspectLibrary>
                  <groupId>io.meeds.portal</groupId>
                  <artifactId>portal.component.common</artifactId>
                </aspectLibrary>
              </aspectLibraries>

```

rdenarie (approved):

PR #4119, opened 2024-10-16
fix: Link settings are not correctly read from server and display an old config - EXO-71985

rdenarie:
Before this fix, the links settings are server with Cache-Control = must-revalidate, and maxAge = 1year must-revalidate means : if the asset is expired in cache, revalidate it against the server. Else, serve the cache content. Firefox correctly implements this behaviour as Chrome revalidate assets for each request. This behaviour means that if an admin change the configuration of the setting, a user on FF could see an old version of the configuration, until his cache is refreshed or asset expired.

The commit change the cache control to no-cache, which means : revalidate the asset for each request. As the LinkSettings endpoint correctly implements Etag feature, if the settings did not change, the service return correctly a 304 and the browser serve the cached asset. If the configuration change, the service serves the new version of the configuration.

This value is different from no-store which means no cache at all.

Resolves meeds-io/meeds#71985

### DIFF
--- a/component/service/src/main/java/io/meeds/social/link/rest/LinkRest.java
+++ b/component/service/src/main/java/io/meeds/social/link/rest/LinkRest.java
@@ -73,8 +73,7 @@ public class LinkRest implements ResourceContainer {
   private static final Log          LOG                    = ExoLogger.getLogger(LinkRest.class);
 
   static {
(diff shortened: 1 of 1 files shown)


rdenarie:
@boubaker It could interest you

boubaker:
> @boubaker It could interest you

cool, well done :+1:

mkrout (approved):

PR #4129, opened 2024-10-22
feat: Display the activity views count - EXO-74854 - Meeds-io/MIPs#161

sofyenne:


webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/content/ActivityLink.vue
```diff
@@ -122,6 +122,19 @@
           @click="displayFullContent">
           <span class="pl-6">{{ $t('UIActivity.label.seeMore') }}</span>
         </v-btn>
+        <div
+          v-if="activityViews"
+          :title="activityViewsTooltip"
```
azayati, line 127: it should be v-tooltip instead IMO as it is done in news detail

azayati (changes requested):

webapp/portlet/src/main/webapp/vue-apps/activity-stream/components/activity/content/ActivityLink.vue
sofyenne: To follow the guidelines, we use the same navigator tooltip as the activity link title and summary, as confirmed by the PO!

azayati (approved):
LGTM

azayati (changes requested):
The label displayed by the tooltip should be updated same as content detail, we need two labels "view" and "views"

sofyenne:
> The label displayed by the tooltip should be updated same as content detail, we need two labels "view" and "views"

done!

azayati (approved):
LGTM

PR #4160, opened 2024-11-05
feat: Hide activity reaction based on activity extension option - EXO-72747 - Meeds-io/MIPs#161

hakermi:
Hide activity reaction based on activity extension option


azayati (approved):
LGTM

PR #4278, opened 2024-12-18
feat:  Update site navigation tree component to display trash administration node based on trash navigation feature flag - EXO-76028

sofyenne:
This change will display the trash navigation note based on the trash navigation feature flag property. It will be removed when the feature flag is removed!

### DIFF
--- a/webapp/src/main/webapp/vue-apps/site-details/components/SiteNavigationTree.vue
+++ b/webapp/src/main/webapp/vue-apps/site-details/components/SiteNavigationTree.vue
@@ -119,6 +119,15 @@ export default {
           if (n.children?.length
               || n.pageLink
               || n.pageKey) {
+            if (!eXo.env.portal.documentsTrashNavigationEnabled) {
+              const documentsSettingsNode = n.children.find(e => e.name === 'documents-settings');
+
+              if (documentsSettingsNode) {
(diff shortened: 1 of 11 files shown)


webapp/src/main/webapp/vue-apps/site-details/components/SiteNavigationTree.vue
```diff
@@ -119,6 +119,15 @@ export default {
           if (n.children?.length
               || n.pageLink
               || n.pageKey) {
+            if (!eXo.env.portal.documentsTrashNavigationEnabled) {
```
azayati, line 122: this should not be done in social since it is specific to eXo

azayati (changes requested):

webapp/src/main/webapp/vue-apps/site-details/components/SiteNavigationTree.vue
sofyenne: This is a temporary change , it will be removed by removing the feature flag , **see PR description!**
else do you have any other idea on how to hide a navigation by FF that created by another addon , knowing that the navigation are created by xml config!

rdenarie: agree with @azayati

PR #4325, opened 2025-01-07
feat: Add translate option for property setting  dropdown list option values - MEED-7999 - Meeds-io/MIPs#171

hakermi:
Add translate option for profile property dropdown list option values

### DIFF
--- a/component/core/src/test/java/org/exoplatform/social/core/plugin/ProfilePropertySettingOptionTranslationTest.java
+++ b/component/core/src/test/java/org/exoplatform/social/core/plugin/ProfilePropertySettingOptionTranslationTest.java
@@ -1,0 +1,61 @@
+/*
+ * This file is part of the Meeds project (https://meeds.io/).
+ *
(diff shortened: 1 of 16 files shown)


component/core/src/test/java/org/exoplatform/social/core/plugin/ProfilePropertySettingOptionTranslationTest.java
```diff
@@ -0,0 +1,41 @@
+package org.exoplatform.social.core.plugin;
```
rdenarie, line 21: header is missing

rdenarie (changes requested):

component/core/src/test/java/org/exoplatform/social/core/plugin/ProfilePropertySettingOptionTranslationTest.java
hakermi: added

rdenarie (approved):

PR #4348, opened 2025-01-16
feat: Reindex targeted identities when used dropdown option value translation updated - EXO-8007 - Meeds-io/MIPs#171

hakermi:
Reindex targeted identities when used option value translation updated

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/listeners/ProfilePropertySettingOptionTranslationListener.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/listeners/ProfilePropertySettingOptionTranslationListener.java
@@ -1,0 +1,79 @@
+/*
+ * This file is part of the Meeds project (https://meeds.io/).
+ *
+ * Copyright (C) 2025 Meeds Association contact@meeds.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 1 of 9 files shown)


component/core/src/main/java/org/exoplatform/social/core/listeners/ProfilePropertySettingOptionTranslationListener.java
```diff
@@ -0,0 +1,79 @@
...
+import org.exoplatform.social.core.profileproperty.ProfilePropertyService;
+import org.exoplatform.social.core.profileproperty.model.ProfilePropertyOption;
+import org.exoplatform.social.core.profileproperty.model.ProfilePropertySetting;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+public class ProfilePropertySettingOptionTranslationListener extends Listener<ProfilePropertyOption, Map<Locale, String>> {
```
rdenarie, line 43: Should this listener be @Asynchronous ?

rdenarie (changes requested):

component/core/src/main/java/org/exoplatform/social/core/listeners/ProfilePropertySettingOptionTranslationListener.java
hakermi: yes, done

rdenarie (approved):

PR #4376, opened 2025-01-27
fix : Fix the avatar display when using arabic language - MEED-8132 - Meeds-io/meeds#2496

MayTekayaa:
This change will fix the avatar display when using arabic language.


component/core/src/main/java/org/exoplatform/social/core/jpa/storage/RDBMSIdentityStorageImpl.java
```diff
@@ -1266,7 +1267,11 @@ private String getNameAbbreviation(Identity identity) {
         }
       }
     }
-    String result = name.replaceAll("\\B.|\\P{L}", "").toUpperCase();
+    String result = Arrays.stream(name.split(" "))
+                          .filter(word -> word != null && word.length() > 0)
```
boubaker:
```suggestion
                          .filter(StringUtils::isNotBlank)
```

boubaker (approved):
ok for me, approved. But just a small code optimization suggestion

PR #4395, opened 2025-02-03
feat: Implement simple storage attachments upload - MEED-8222,MEED-8238,MEED-8224,MEED-8223 - Meeds-io/MIPs#176

hakermi:
Implement simple storage attachments upload


component/core/src/main/java/org/exoplatform/social/core/attachment/AttachmentServiceImpl.java
```diff
@@ -109,6 +109,10 @@ public ObjectAttachmentOperationReport saveAttachments(FileAttachmentResourceLis
                                                          Identity userAclIdentity) throws ObjectNotFoundException,
                                                                                    IllegalAccessException {
     checkEditPermissions(attachmentList, userAclIdentity);
+    if (!this.attachmentPlugins.get(attachmentList.getObjectType()).canUpdateAttachmentList()) {
```
rdenarie, line 112: if this.attachmentPlugins.get(attachmentList.getObjectType() == null => you will hace a NPE.

Possible to extract this if in a separated method ?

rdenarie (changes requested):

component/core/src/main/java/org/exoplatform/social/core/attachment/AttachmentServiceImpl.java
hakermi: since it passed the  checkEditPermissions then it will never be null and since its not yet a common function i didn't extract it.

rdenarie (approved):

rdenarie (approved):

PR #4461, opened 2025-02-14
Merge mip 178

AzmiTouil:


webapp/src/main/webapp/vue-apps/common/main.js
```diff
@@ -22,6 +22,7 @@ import * as profileLabelService from '../common/js/ProfileLabelService.js';
 import * as siteService from './js/SiteService.js';
 import * as navigationUtils from './js/NavigationUtils.js';
 import * as spaceTemplateService from './js/SpaceTemplateService.js';
+import * as registrationService from '../platform-access/js/RegistrationService';
```
boubaker: May be it would be better to import it only when needed, or to move into `common/js` folder, WDYT ?

```suggestion
```

AzmiTouil: Yes, I think it's better to move into common!

boubaker (approved):

PR #4489, opened 2025-02-27
fix: Trigger a listener event after space templates import - EXO-66372 - Meeds-io/meeds#66372

mkrout:
Prior to this fix, default spaces cannot be created on first server start since space templates are not yet initialized. 
This commit adds a listener event when the default space templates are created so other services can run the spaces creations

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/SpaceStorage.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/SpaceStorage.java
@@ -795,7 +795,7 @@ public class SpaceStorage {
               String name = Utils.cleanString(displayName);
               int index = 0;
               while (spaceDAO.getSpaceByPrettyName(name) != null) {
-                // Use sapme algorithm to compte new pretty name as for
+                // Use the same algorithm to compute new pretty name as for
                 // creation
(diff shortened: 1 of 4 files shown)


azayati (approved):

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/SpaceStorage.java
```diff
@@ -788,15 +788,22 @@ private Instant computeCreatedDate(Token token) {
...
-    }
-    return prettyName;
+  private String buildPrettyName(String... names) {
+    return Arrays.stream(names)
+            .filter(StringUtils::isNotBlank)
+            .map(displayName -> {
+              String name = Utils.cleanString(displayName);
+              int index = 0;
+              while (spaceDAO.getSpaceByPrettyName(name) != null) {
+                // Use sapme algorithm to compte new pretty name as for
```
azayati:
```suggestion
                // Use the same algorithm to compute new pretty name as for
```

azayati (dismissed):

azayati (changes requested):
review

Jihed525 (dismissed):

azayati (approved):

PR #4566, opened 2025-03-19
fix: who's online portlet display js erreur when loaded by external user - MEED-8644

rdenarie:
Before this fix, when we upgrade from 6.5 to 7.0, the whosonline portlet generates and error for externals only because Vue.prototype.$updateApplicationVisibility is not loaded. On a new create space, this component is loaded when portlet sidebar is mounted, but on a space coming from 6.5, it still use the old layout, and the vue object is not loaded. This commit ensure that the page is loaded before calling updateApplicationVisibility, so that the Vue object exists.

Resolves meeds-io/meeds#3140

### DIFF
--- a/webapp/src/main/webapp/WEB-INF/jsp/portlet/whoIsOnline.jsp
+++ b/webapp/src/main/webapp/WEB-INF/jsp/portlet/whoIsOnline.jsp
@@ -59,7 +59,11 @@
 <% } else { %>
   <div id="OnlinePortlet">
     <script type="text/javascript">
-      require(['SHARED/vue'], () => Vue.prototype.$updateApplicationVisibility(false, document.querySelector('#OnlinePortlet')));
+      document.onreadystatechange = () => {
+        if (document.readyState === 'complete') {
+          require(['SHARED/vue'], () => Vue.prototype.$updateApplicationVisibility(false, document.querySelector('#OnlinePortlet')));
+        }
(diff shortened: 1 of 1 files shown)


Jihed525 (approved):

webapp/src/main/webapp/WEB-INF/jsp/portlet/whoIsOnline.jsp
```diff
@@ -59,7 +59,11 @@
 <% } else { %>
   <div id="OnlinePortlet">
     <script type="text/javascript">
-      require(['SHARED/vue'], () => Vue.prototype.$updateApplicationVisibility(false, document.querySelector('#OnlinePortlet')));
+      document.onreadystatechange = () => {
+        if (document.readyState === 'complete') {
+          require(['SHARED/vue'], () => Vue.prototype.$updateApplicationVisibility(false, document.querySelector('#OnlinePortlet')));
```
boubaker, line 64: you can change the required module "SHARED/commonVueComponents" to ensure the init is made, rather than waiting on document init. In fact, in layout editor, the portlet won't show because the document has become ready already. Using the right dependant module will fix the behavior.

rdenarie: ok, I will check this modification

PR #4588, opened 2025-03-25
feat: Unauthorize importing portlets' instances in page templates app -  MEED-8604 - Meeds-io/MIPs#180

AzmiTouil:
This PR will avoid importing portlets' instances in page templates app

### DIFF
--- a/component/core/src/main/java/io/meeds/social/databind/service/DatabindServiceImpl.java
+++ b/component/core/src/main/java/io/meeds/social/databind/service/DatabindServiceImpl.java
@@ -28,11 +28,14 @@ import org.exoplatform.upload.UploadService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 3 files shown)


component/core/src/main/java/io/meeds/social/databind/service/DatabindServiceImpl.java
```diff
@@ -28,11 +28,14 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Service;
+import org.springframework.web.server.NotAcceptableStatusException;
```
boubaker: Should be replaced by

```suggestion
import java.lang.IllegalStateException;
```

In fact, `NotAcceptableStatusException` is an exception to use in Presentation Layer as REST response, see its javadoc:

image

AzmiTouil: Yes, thanks!

boubaker (approved):

PR #4647, opened 2025-04-14
fix: Fix posting message when the post button is clicked many times - MEED-8762 - Meeds-io/meeds#3038

MayTekayaa:
This change will fix posting message when the post button is clicked many times.

Resolves Meeds-io/meeds#3038

### DIFF
--- a/webapp/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
+++ b/webapp/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
@@ -378,13 +378,13 @@ export default {
           activityType = 'LINK_ACTIVITY';
         }
         if (this.activityType && this.activityType.length !== 0) {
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityComposerDrawer.vue
```diff
@@ -345,6 +345,10 @@ export default {
       });
     },
     postMessage() {
+      if (this.loading) {
+        return;
+      }
+      this.loading = true;
```
boubaker, line 351: There is some cases where `loading` will remain true after the method finishes its execution (See `if` condition in line 380)

MayTekayaa: yes true, thanks
Fixed now!

boubaker (approved):

PR #4651, opened 2025-04-15
fix: Fix the bad mention user display - MEED-8760 - Meeds-io/meeds#3120

MayTekayaa:
This change will fix the bad mention user display.

Resolves Meeds-io/meeds#3120

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
@@ -201,7 +201,7 @@ public class LinkProvider {
                .append(identity.getProfile().getAvatarUrl())
                .append("',")
                .append("position: '")
-               .append(identity.getProfile().getPosition() == null ? "" : identity.getProfile().getPosition().replace("'", "\\\\'"))
+               .append(identity.getProfile().getPosition() == null ? "" : identity.getProfile().getPosition().replace("'", "\\\\'").replace("\"", "\\\\\""))
                .append("',")
                .append("external: '")
                .append(identity.getProfile().getProperty(Profile.EXTERNAL) != null && StringUtils.equals("true", String.valueOf(identity.getProfile().getProperty(Profile.EXTERNAL))))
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/js/ExtendedDomPurify.js
```diff
@@ -22,7 +22,7 @@
   ExtendedDomPurify.prototype.purify = function(content) {
     content = content.replace(/<div> <\/div>/g, '<div><br><\/div>');
     content = content.trim().replace(/>[ \n]+</g, '> <').replace(/  /g, '&nbsp;&nbsp;');
-    const pureHtml = DOMPurify.sanitize(Autolinker.link(content, {
+    let outputHtml = /<a\s/i.test(content) ? content : Autolinker.link(content, {
```
boubaker: It would be better to not cancel using Autolink. In fact, the fix should go in `org.exoplatform.social.core.service.LinkProvider.getProfileLink(String, String)` 

IMO, we need to add a fix in order to allow escaping the double quote (as made in simple quote) inside the method
image

component/core/src/main/java/org/exoplatform/social/core/service/LinkProvider.java
```diff
@@ -201,7 +201,7 @@ public static String getProfileLink(String username, String portalOwner) {
                .append(identity.getProfile().getAvatarUrl())
                .append("',")
                .append("position: '")
-               .append(identity.getProfile().getPosition() == null ? "" : identity.getProfile().getPosition().replace("'", "\\\\'"))
+               .append(identity.getProfile().getPosition() == null ? "" : identity.getProfile().getPosition().replace("'", "\\\\'").replace("\"", "\\\\\""))
```
boubaker, line 204: Can you please do the same for the `FullName`

MayTekayaa: done!

boubaker (approved):

PR #4661, opened 2025-04-17
fix: Fix the bad mention user display - MEED-8760 - Meeds-io/meeds#3120

MayTekayaa:
This change will hide attributes temporarily when using autolinker to avoid breaking the html.


boubaker (commented):
> This change will hide attributes temporarily when using autolinker to avoid breaking the html.

This will cancel the main purpose of using `DOMPurify` (which will sanitize the attributes)

MayTekayaa:
> > This change will hide attributes temporarily when using autolinker to avoid breaking the html.
> 
> This will cancel the main purpose of using `DOMPurify` (which will sanitize the attributes)

You are right, i have change it in order to hide attribute only when using autolinker and then restore it when sanitize html element.

boubaker (approved):

PR #4739, opened 2025-05-08
feat: Allow hide drawer footer divider - MEED-8964 - Meeds-io/MIPs#163

hakermi:
Allow hide drawer footer divider


ahamdi:
What is the case that needs to display the divider while hiding the footer ? could you give more context ?

hakermi:
> What is the case that needs to display the divider while hiding the footer ? could you give more context ?

https://xd.adobe.com/view/4b28bebc-4ba3-4235-a553-130595a47c00-1e7b/

ahamdi (approved):

PR #4782, opened 2025-05-16
fix: Modify the color of summary/description and the number of views of activity in the stream - MEED-9099 - Meeds-io/meeds#3311

akhanfir:
Before this change, when Create an article/notes and publish it then view it by some users, The Color of the summary/description and the number of views in the stream is #707070. To fix this problem, change the class text-light-color if it is displayed and add a class text-color. After this change, the color of summary/description and the number of views is changed from #707070 to #20282C.
Resolves https://github.com/Meeds-io/meeds/issues/3311


azayati (changes requested):
This change will impact any activity link not only news/notes ?

azayati (approved):

PR #4799, opened 2025-05-21
fix: Fix Site Pages Retrieval in Sidebar Navigation Admin UI - MEED-9105 - Meeds-io/meeds#3316

azayati:
Prior to this change, when opening the Site Navigation Config Drawer the second time, the 'initialized' flag is set as 'false' which doesn't permit to refresh the list of pages. This change will fix the bahavior by setting 'initialized' flag as 'false' all time in a finally block at the end of Drawer opening operation.


boubaker:
@azayati can you please Squash and Merge by UI in order to preserve the original author in the commit

rdenarie (approved):

PR #4813, opened 2025-05-22
feat: Hide the Topbar loading after the activity search module initialization - MEED-9139 - Meeds-io/MIPs#193

sofyenne:


webapp/src/main/webapp/vue-apps/search-activity/main.js
```diff
@@ -13,8 +13,9 @@ if (extensionRegistry) {
 export function formatSearchResult(result) {
   return result;
 }
-Vue.prototype.$utils.includeExtensions('ActivityStreamExtension');
+Vue.prototype.$utils.includeExtensions('ActivityStreamExtension').then(()=> {
+  // hide the TopBar loading started by the activity stream (dependency) initialization
+  document.dispatchEvent(new CustomEvent('hideTopBarLoading'));
+});
```
boubaker, line 19: You can move this

https://github.com/Meeds-io/social/blob/develop/webapp/src/main/webapp/vue-apps/activity-stream/main.js#L25

inside init method instead

sofyenne: done! thanks

PR #4894, opened 2025-06-09
feat: Implement a custom common emojiPicker - MEED-9135 - Meeds-io/MPIs#164 

hakermi:
Implement a custom common emojiPicker to be used a cross the platform

### DIFF
--- a/webapp/src/main/webapp/vue-apps/emoji-picker/main.js
+++ b/webapp/src/main/webapp/vue-apps/emoji-picker/main.js
@@ -1,0 +1,43 @@
+/*
+ This file is part of the Meeds project (https://meeds.io/).
+
+ Copyright (C) 2025 Meeds Association contact@meeds.io
+
+ This program is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 3 of the License, or (at your option) any later version.
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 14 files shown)


ahamdi (approved):

webapp/src/main/webapp/vue-apps/emoji-picker/main.js
```diff
@@ -0,0 +1,43 @@
...
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public License
+ along with this program; if not, write to the Free Software Foundation,
+ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import './initComponents.js';
+import {emojiBank} from './js/emojiBank.js';
```
boubaker, line 20: Can you please change this file into a JSON file which you will fetch with an efficient cache time (ttl = 365days)
In fact, the file is very big (750 KB) and shouldn't be parsed as a JS file in browser. In addition, I think that it would be better to not block the display of Rich Editor until this file is downloaded due to the fact that the Emojis data is mixed with real JS files holding Business Behavior.

hakermi: Yes, its already planned to change that to a json file, it will come with other improvement to the component.

boubaker: Thanks :+1:

PR #4923, opened 2025-06-13
fix : Add validators when using Social Rest Endpoint to create users - MEED-9303

rdenarie:
Before this fix, when calling Social endpoint to create users, there was any validators like for Portal Endpoint This commit add the validators on username, password, email, firstname and lastname

Resolves meeds-io/meeds#3417

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRest.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRest.java
@@ -487,6 +487,7 @@ public class UserRest implements ResourceContainer, Startable {
       @ApiResponse (responseCode = "200", description = "Request fulfilled"),
       @ApiResponse (responseCode = "400", description = "Invalid query input") })
   public Response addUser(@Context UriInfo uriInfo,
+                          @Context HttpServletRequest request,
                           @Parameter(description = "Asking for a full representation of a specific subresource if any") @QueryParam("expand") String expand,
                           @RequestBody(description = "User object to be created, ex:<br />" +
                               "{<br />\"username\": \"john\"," +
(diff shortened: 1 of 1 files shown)


azayati (approved):

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRest.java
```diff
@@ -513,13 +514,40 @@ public Response addUser(@Context UriInfo uriInfo,
       throw new WebApplicationException(Response.Status.FORBIDDEN);
     }
 
+    Locale locale = request == null ? Locale.ENGLISH : request.getLocale();
```
boubaker, line 517: This method is duplicated with the one defined in Portal which already makes the correct validation:

`org.exoplatform.portal.rest.UserRestResourcesV1.createUser(HttpServletRequest, UserRestEntity)`

It may be better to delete this endpoint and move the Portal endpoint to social since it's the REST API that we use in Users Management UI, WDYT ?

rdenarie: In fact, there is 2 things : 
- Cleanup code : can only occur in a minor version : 7.1. This is to let time to customers using the existing API to see it as deprecated
- Fix the existing code : for version 7.0.x, there is no validation at all when this endpoint is called, so we need to fix it. 

This task aim to realize the second point in a maintenance task, for all customers using 7.0.x version. 

The first point about clean up should be done in an improvment task, and only on a minor version.

So I have no objection about movide the portal endpoint into social, but not in the scope of this task

boubaker: ok I understand about splitting concerns, thus I 'll keep this in [Technical Backlog](https://github.com/Meeds-io/meeds/issues/3425) for now.

PR #4946, opened 2025-06-17
Merge mip 187

### LINKED ISSUE #187: Space creation instanciation
## Rationale

Meeds Association aims to let any visitor create their own community in just a few clicks.
The current process—filling out a form to apply for a Hub—creates friction and doesn't align with our mission of offering a simple, fast onboarding experience for communities.

This MIP introduces a portlet for admins to simplify and streamline the community creation experience.

#### Instanciation of the portlet

As an admin, I can add in any page a new portlet called `Space Creation`
Once added, I can apply any layout options provided by the framework (background, alignement, height, text, border, radius, background, margins).

Once added, I can configure the instance at my convenience:
- Update the button name (including translations, emojis)
   - Default value: Create
- Choose a template for space creation
   - Any template (default value)
   - Only few templates (multiple templates can be selected)

Image

This instanciation is contextual to this page. Indeed, adding the same portlet elsewhere does not mean we use the previous customized configuration.

#### Portlet Use

Once the portlet is added to a page, any user can see and use it.
Indeed, no matter you are logged in, you can start creating a space using this CTA.

When clicking to it, then the usual space creation journey is provided.
As for example, if multiple space templates are selected, then these are listed so the user can choose which one to use to start.
And if only one space template is chosen, then the user can start right away the space creation.

Once the space template is selected, I can start creating my space following the suggested steps.
When using desktop, a space card is provided at the left of the creation form so I can see my space being created.

Image

Once I confirm the space creation, then a check must be done to verify if I am already logged in or not:
- If yes, then the space is created as expected
- If not, then I am redirected to the login page AND the space creation is suspended until I am logged in.

Once I am logged in (including after signing up for a new account):
- the space is created as expected
- I must be redirected to the space previously created

That means we store the informations previously added for creating a space until the user confirm its profile.
If no log in or registration in the following time, then the space creation is cancelled.

#### Gamification
NA

#### Notifications
NA

#### Analytics
NA

#### Unified Search
NA

- The Space Creation Form Data shouldn't remain after the space is created or aborted
- Knowing that this CTA will be displayed in public pages, but rarely clicked, the Space Creation portlet shouldn't affect global performance page rendering by loading all JS files on display

### Security 

The Space Creation shouldn't be allowed only to users who are allowed to do so, using Spaces Administration parameters (Space Creation permission). In fact, we should secure the flow willing to help unregistered users to create spaces once registered.

### Extensibility 

The existing `SHARED/spaceForm` JS Module has to be extended in order to allow supporting features of the current MIP.

### Configurability 

N/A

### Upgradability

N/A

### Existing Features

No change on existing features

### Feature Flags

N/A

N/A

### Security

In order to secure the space creation form when accessed by a non registered user, then the space creation flow has to be secured as follows:
- Give access to  Space Templates (the list of available space templates/space categories should be done using JSP and not using REST endpoint due to Security limitations when accessing Space templates/categories through REST endpoint)
- The Space Form will not send its content into a REST, but a handler (similar to `ExternalRegisterHandler`) which will create a Token using `CookieTokenService.createToken`: the handler will be available under URL `/portal/cs`
- Once the random Token is created, place it as a cookie
- Redirect the user from the handler to `/portal/login?initialURI=%2Fportal%2Fcs`
- Once the user authenticated, it will be automatically redirected to the handler thanks to `initialURI` param
- The Handler reads the cookie previously saved and retrieves the corresponding cached Space Form Data (a cache with a TTL of some minutes)
- The handler, after finding the space data to create, clears the corresponding cache data
- The handler creates the corresponding Space (if an error happens, open the space form again from `Spaces` UI). In fact, the user may not having access to the selected category or Space Template
- Once the space created, redirect the user to `/portal/s/{spaceId}`

### Access 
   - The same UI of `SHARED/spaceForm` module has to be used and adapted to this new usage (See as example, how the sidebar calls Space Creation Form using this JS Module call)
   - No modification is expected in REST endpoints

### Services & processing
A new Portlet has to be created with its specific Java Class in order to handle Portlet Preferences saving and manage the display of Space Templates/Categories for end user, knowing that an anonymous user will not have access to both entities through REST endpoints. Thus, the `View` method of the portlet has to manage retrieving and providing the selected Space Template(s)/categorie(s) to an anonymous user when clicking on "Create" CTA button.
This portlet has to save the portlet preferences (Space Template(s)/Categorie(s) **identifier(s)**) and other preferences.

### Data and persistence

N/A

MayTekayaa:
This MIP add a portlet for admins to simplify the community creation experience and to let any visitor create their own community.

Resolves Meeds-io/MIPs#187


component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRest.java
```diff
@@ -385,6 +386,41 @@ public Response createSpace(
...
+    String tokenId = spaceService.prepareSpaceInstance(new SpaceCreationInstance(model.getId(),
+                                                                                 model.getDisplayName(),
+                                                                                 model.getDescription(),
+                                                                                 model.getTemplateId(),
+                                                                                 model.getVisibility(),
+                                                                                 model.getSubscription(),
+                                                                                 model.getBannerId(),
+                                                                                 model.getAvatarId(),
+                                                                                 model.getPrettyName()));
+    return Response.ok(tokenId).cookie(new NewCookie("spaceCreationCookie",
```
rdenarie, line 414: It seems that the cookies is created without the option HttpOnly. 
I think we should create it with httpOnly option = true to prevent usage of cookie on client side (https://contentsquare.com/fr-fr/blog/httponly/)

boubaker: Thanks for reporting, this is an enhancement in `Meeds WS` module in fact which should be improved. This generic concern has to be considered in a Maintenance Task apart since the JAX-RS JSR 311 does't define it.

Will be considered in the issue https://github.com/Meeds-io/meeds/issues/3441

rdenarie: Why to not simply add a boolen for httpOnly at the end of the list ? 
Currently the contructor used is 
NewCookie(String name, String value, String path, String domain, String comment, int maxAge, boolean secure)
but we could use 
NewCookie(String name, String value, String path, String domain, String comment, int maxAge, boolean secure, boolean httpOnly)

I'm not sure to understand the relation with the Meeds WS module ?

boubaker: `NewCookie` is a class provided by JSR 311 API:
`javax.ws.rs.core.NewCookie`

rdenarie: ok, so the constructor I mentionned does not exists.

Ok for me

boubaker: Do you mean we need to patch the JAR `jsr311-api.jar` or upgrade its version ?

boubaker: > ok, so the constructor I mentionned does not exists.

yes, indeed

boubaker (approved):

PR #4953, opened 2025-06-18
feat: content link not well displayed in activity search result - MEED-9305 - Meeds-io/MIPs193

sofyenne:


component/api/src/main/java/io/meeds/social/html/model/HtmlTransformerContext.java
```diff
@@ -38,6 +38,8 @@ public class HtmlTransformerContext {
 
   private Locale   locale;
 
+  private boolean  isTextOnly;
```
boubaker, line 41: `
  private boolean  textOnly;
`

component/core/src/main/java/io/meeds/social/cms/plugin/ContentLinkHtmlTransformerPlugin.java
```diff
@@ -171,6 +180,30 @@ private String replaceDataObjectTag(String html) {
...
   }
 
+  private String replaceContentLinkText(String html, HtmlTransformerContext context) {
+    List<String> extensionObjectTypes = contentLinkPluginService.getExtensions()
+                                                                .stream()
+                                                                .map(ContentLinkExtension::getObjectType)
+                                                                .toList();
+
+    for (String type : extensionObjectTypes) {
+      String regex = "/" + Pattern.quote(type) + ":([^/\\s]+)";
```
boubaker, line 190: Build the **compiled Pattern** once in `init()` method and then put it inside a `Map` in order to reuse it each time (`Pattern.compile` is a slow operation)

PR #5006, opened 2025-07-04
Merge Mip 203

### LINKED ISSUE #203: Login Portlets & Pages
## Rationale

Login related pages are not customizable yet and multiple times, it happens that CSS is applied to personnalize labels, to add contents above the login form. 
Indeed, as for example, when the login page suggest an additional login method in addition to the meeds login form (OIDC, SAMLv2, etc.), then the labels are not understandable enough for users.
Hence, they sometimes try to enter their credentials in the Meeds Login Form instead of clicking to the SSO authentication CTA.

In order to avoid adding CSS to these pages and to help end-users accessing the platform the easiest as possible, we thus imagine this MIP to enable admins customize any related login pages:
- login page
- registration page
- forgot password page

Please find below the details.

#### Login portlets

Different portlets need to be created, based on what it is currently provided:
- Login form

Image

- Registration form

Image

- Forgot password form

Image

In addition, the user journey when multiple authentication methods are provided will be reviewed:
- as far as there is one authentication method is provided in addition to the meeds login form
- then the login form will not be provided
- a CTA will be suggested to finally access the login form
This must improve the experience when multiple authentication are listed.

Image

Each of these portlets are listed as available in instances in a new category called "Login". Only platform admins are able to use those ones.

#### Portlets customization

Admins will only have customization options for login portlet in order to review the text, the options to set the authentication options.
Other portlet will only be a portlet so it is editable using layout options.

External providers will be provided depending on the configuration.

Image

#### Other portlets

To enable admins to reuse items listed defaultly in login pages, other portlets are needed:
- Platform Name:
    - with default style of title
    - if needed, the style can be changed using layout options
- Platform Logo:
    - different than topbar logo as we only need the platform logo and not the name
    - if needed, the style can be changed using layout options (margin, radius, background, etc.)
- Sidebar login:
   - App displayed at the left of the desktop view of login page
   - Settings for this app will be provided in a drawer thanks to a settings access for admin when hovering it

#### Pages management

In order to let admins edit the layout of each page, those ones will be listed in the global site with the option to edit it if needed.

Each page is defaultly organized the same in a grid view
- Page design specifications:
   - full window
   - no margins
- Design to be adapted after some tests using portlets

#### Review of login page customization page

Since login page can be customized using layout management and global pages, then the login page customization page is suggesting the edition of each page.

Image

#### Gamification
NA

#### Notifications
NA

#### Analytics
NA

#### Unified Search
NA

### Security 

Login, Register and Forgot password portlet will be accessible by anonymous users.

### Existing Features

Currently, the url /portal/login is managed by LoginHandler, which allow to display the login form, and to capture the authentication flow. 

We will add a new handler : PageAccessHandler. This handler will allow to access to a page of global site directly, if permissions allows.

With this handler, when accessing to url /portal/login, will display page login in portal global, if permissions are suitable.
This handler can be used for all pages of global portal, for all levels in navigation.

This login handler will be kept. If the request is not a POST containing parameters needed for the authentication, it do nothing, and let the other handler process the request.
Else, the LoginHandler still process the authentication, we will not change the login handler which will continue to perform the authencation, or delegate it to SSO configurations.

The pageAccessHandler will be used for other page : Register and onboarding

### Feature Flags

No feature flags. Portlets will be created in step 1, without adding it to pages, then we conserve current authentication process. 
When portlets are ready, we will process to step 2 : creating the new login page.

### Security
The new pages created in global portal must be accessible to all users.

### Access 
   - GUI 
   - API 

### Services & processing
   - Ex: Scheduled Jobs

### Data and persistence

New portlets will be created and will handle settings in Portlets preferences

## 5. Annexes
   - Reasoning that explain the technical choices
   - Impacts 
   - References

rdenarie:
This commit add the portlet Platform Name
This portlet display the branding name as a title


component/api/src/main/java/io/meeds/social/translation/service/TranslationService.java
```diff
@@ -32,6 +32,43 @@ public interface TranslationService {
...
+   *                                 with a corresponding label for each saved
+   *                                 {@link Locale}
+   * @throws IllegalAccessException  When user doesn't have access permission to
+   *                                   object
+   * @throws ObjectNotFoundException When object wasn't found
+   */
+  Map<String, TranslationField> getAllTranslationFields(String objectType,
+                                                 String objectId,
+                                                 String username) throws IllegalAccessException, ObjectNotFoundException;
+
```
boubaker, line 55: Code formatting

component/core/src/main/java/org/exoplatform/social/core/attachment/AttachmentServiceImpl.java
```diff
@@ -360,6 +360,39 @@ public void moveAttachments(String sourceObjectType,
...
+  }
+
+
+
+  private void moveAttachments(String sourceObjectType,
+                              String sourceObjectId,
+                              String destinationObjectType,
+                              String destinationObjectId,
+                              String destinationParentObjectId,
+                              long userIdentityId, boolean deleteAfterMove) {
```
boubaker:
```suggestion
                              long userIdentityId,
                              boolean deleteAfterMove) {
```

component/core/src/test/java/io/meeds/social/translation/service/TranslationServiceTest.java
```diff
@@ -39,7 +39,7 @@ public class TranslationServiceTest extends AbstractCoreTest { // NOSONAR
 
   private String                 objectType  = OBJECT_TYPE;
 
-  private long                   objectId    = 322l;
+  private String                   objectId    = "322";
```
boubaker:
```suggestion
  private String                 objectId    = "322";
```

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
```diff
@@ -0,0 +1,109 @@
...
+import org.exoplatform.social.attachment.AttachmentService;
+import org.exoplatform.social.rest.api.RestUtils;
+
+import java.security.SecureRandom;
+import java.util.Map;
+
+public class CmsPortletWithMetadata extends CMSPortlet {
+
+  public static final String OBJECT_TYPE = "cmsPortlet";
+  public static final String TRANSLATION_IDENTIFIER = "translationIdentifier";
```
boubaker:
```suggestion

  public static final String TRANSLATION_IDENTIFIER = "translationIdentifier";
```

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private       TranslationService translationService;
+  private       AttachmentService  attachmentService;
+  private final SecureRandom       random = new SecureRandom();
```
boubaker:
```suggestion
  private       TranslationService translationService;

  private       AttachmentService  attachmentService;

  private  SecureRandomService secureRandomService;
```

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
```diff
@@ -0,0 +1,109 @@
...
+
+  @Override
+  protected void setViewRequestAttributes(String name, RenderRequest request, RenderResponse response) {
+    PortletPreferences preferences = request.getPreferences();
+    String currentTranslationIdentifier = preferences.getValue(TRANSLATION_IDENTIFIER,null);
```
boubaker:
```suggestion
    String currentTranslationIdentifier = preferences.getValue(TRANSLATION_IDENTIFIER, null);
```

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
```diff
@@ -0,0 +1,109 @@
...
+    if (currentTranslationIdentifier == null) {
+      // If translation identifier is not set, generate a new one
+      Long randomIdentifier = random.nextLong() & Long.MAX_VALUE;
+      currentTranslationIdentifier = String.valueOf(randomIdentifier);
+      savePreference(TRANSLATION_IDENTIFIER, currentTranslationIdentifier);
```
boubaker: Why not using SETTING_NAME of the portlet ? Why creating a new Portlet Preference identifier ?

webapp/src/main/webapp/vue-apps/common/components/ExoDrawer.vue
```diff
@@ -417,6 +417,10 @@ export default {
         this.drawer = true;
       }
       this.resetFilter();
+      this.$nextTick(() => {
+        window.setTimeout(() =>
+          this.$el.querySelector('input')?.focus(), 50);
+      });
```
boubaker: Why adding this generic behavior in all drawers ?

```suggestion
```

webapp/src/main/webapp/vue-apps/login-form/components/LoginForm.vue
```diff
@@ -0,0 +1,361 @@
...
+  <v-app class="mx-auto" v-show="this.init" style="max-width: 375px;">
+    <login-form-settings-drawer
+          :translation-identifier="values.translationIdentifier"
+          :register-enabled="values.registerEnabled"
+          :signin-option="signinOption"
+          :display-signin-email-button-icon="displaySigninEmailButtonIcon"
+          :list-external-providers="listExternalProviders"
+          :display-providers-icons="displayProvidersIcons"
+          :display-welcome-message="displayWelcomeMessage"
+          v-if="values.canEdit"/>
```
boubaker: Use `npm run lint` to format Vue files.

Besides, the `v-*` attributes goes all time in the first place
+ Drawers goes at the end of the file to let maintainers focus, easily, on Component content rather than Drawers (which are generally defined in a component apart)

webapp/src/main/webapp/vue-apps/login-form/components/LoginForm.vue
```diff
@@ -0,0 +1,361 @@
...
+<style lang="scss">
+  .loginFormSigninEmailButton {
+    min-height: 36px;
+    height: 100% !important;
+  }
+  .loginFormSigninEmailButton > .v-btn__content {
+    max-width: 100%;
+    white-space: normal;
+  }
+</style>
```
boubaker:
```suggestion
```

Avoid defining inline Style inside the Vue files. This was banned from the beginning of Meeds.
The only usage where it was used, where the CSS is dynamically defined for animations. (Dynamic no Static)

webapp/src/main/webapp/vue-apps/login-form/components/LoginFormSettingsDrawer.vue
```diff
@@ -0,0 +1,393 @@
...
+-->
+<template>
+  <exo-drawer
+    ref="drawer"
+    v-model="drawer"
+    :right="!$vuetify.rtl"
+    :loading="loading"
+    eager
+    @closed="reset">
+    <template slot="title">
```
boubaker: Avoid using `slot` since in Vue3, we will have to upgrade all to use `#` instead

```suggestion
    <template #title>
```

webapp/src/main/webapp/vue-apps/login-form/components/LoginFormSettingsDrawer.vue
```diff
@@ -0,0 +1,393 @@
...
+      this.$refs.drawer.close();
+    },
+    save() {
+      this.loading = true;
+      const promise = [];
+      let firstTranslation = true;
+
+      //if we have saveTranslations request, we need to ensure that the first one is executed before the others
+      //because it create the metadata for the translations
+      //if all saveTranslations are executed at the same time, we will have a "unique constraint violation" error because metadata will be created more than once
```
boubaker, line 272: You can use `async` and `await` in this case which makes the method more readable.

webapp/src/main/webapp/WEB-INF/portlet.xml
```diff
@@ -1131,4 +1131,183 @@
...
+    </supports>
+    <supported-locale>en</supported-locale>
+    <resource-bundle>locale.portlet.SidebarLogin</resource-bundle>
+    <portlet-info>
+      <title>Sidebar Login</title>
+      <keywords>sidebar login</keywords>
+    </portlet-info>
+  </portlet>
+
+
```
boubaker, line 1207:
```suggestion
```

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
rdenarie: I firstly tried with SETTING_NAME as for Image and it was not working.

I don't remember exactly, but there is a case, when you save a page template, or something like that, the setting_name change and we don't go in setViewRequestAttribute. It is not problematic for Image Portlet, but in case of translation, it make it not working as initial translations medatadas are not more associated with the correct identifier.

webapp/src/main/webapp/vue-apps/common/components/ExoDrawer.vue
rdenarie: For accessibility purpose : when you navigate with keyboard, and make an action opening a drawer, the focus is not "in" the drawer, so you cannot continue your navigation inside the drawer
This behavior ensure to put the focus on the first input of the drawer for all drawer of the platform.

In addition, it need to be improved in another task to take care of drawer which have no inputs.

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
boubaker: Page Template Management is different. Thus it should be better to fix the `SETTING_NAME` if there is a bug into it, rather than workaround it by redefining a new Portlet Preference IMO.

webapp/src/main/webapp/vue-apps/common/components/ExoDrawer.vue
boubaker: Each drawer will have to manage this apart. In some drawers, not the first input is `autofocus-able`.

webapp/src/main/webapp/vue-apps/login-form/components/LoginForm.vue
rdenarie: This css was added because Vue do not allow to display correctly a text on a button on 2 lines. 
I didnt manage to make the text wrap correctly without this trick.

Have you suggestions about that ?

webapp/src/main/webapp/WEB-INF/portlet.xml
rdenarie: done

webapp/src/main/webapp/vue-apps/login-form/components/LoginFormSettingsDrawer.vue
rdenarie: done

webapp/src/main/webapp/vue-apps/login-form/components/LoginForm.vue
rdenarie: done

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
rdenarie: done

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
rdenarie: done

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
rdenarie: done

component/core/src/test/java/io/meeds/social/translation/service/TranslationServiceTest.java
rdenarie: done

component/core/src/main/java/org/exoplatform/social/core/attachment/AttachmentServiceImpl.java
rdenarie: done

component/api/src/main/java/io/meeds/social/translation/service/TranslationService.java
rdenarie: done

webapp/src/main/webapp/vue-apps/common/components/ExoDrawer.vue
rdenarie: Have you an example of a drawer in which the first input is not autofocus-able ?

This solution have the advantage to make all drawers a11y compatible in one modification

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
rdenarie: What is the bug for you ? 
I'm not sure to understand

webapp/src/main/webapp/vue-apps/login-form/components/LoginForm.vue
boubaker: yes, you can add the text inside a `div` or even a `<v-card max-width=".." width="..." color="transparent" flat>`:
```html
<v-btn...>
<div class="text-wrap">....</div>
</v-btn>
```

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
boubaker: You said `but there is a case, when you save a page template, or something like that, the setting_name change and we don't go in setViewRequestAttribute`. If this behavior isn't ok and is a bug, thus we may need to fix it globally, isn't it ?

rdenarie: For this point, I need to redig into current behaviour. 
Once done, I'll ping you to speak about that.

webapp/src/main/webapp/vue-apps/common/components/ExoDrawer.vue
boubaker: yes, drawers with `input = "checkbox"` or `input = "radio"` or `RichEditor` as first input. The `autofocus` attribute is added in all inputs Type (either it's of type `<input>` or not). The first assumption, that all inputs are of type `<input>` + the fact that it's available right after opening the Drawer is wrong.

Let's do a compromise, make this behavior optional using a dedicated attribute `autofocusFirstInput` with a `false` value by default. WDYT ?

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
boubaker: Sure :+1:

rdenarie: I remember why I not use CMS_SETTINGS_NAME.

CMS_SETTINGS_NAME is on format "Untitled-603-d6964c9b-9480-4f22-8e56-b114e77b3e63"

But, as I save translations, I use translations services (which use metadata service), which require a long as ID for audience_id.

That's why I created a new settings, randomly generated

boubaker: `audience_id` is a long yes which can be 0. No relation with the ObjectId = String.
In fact, there was a problem before which doesn't let using `TranslationService` with a String `ObjectId`, which has been improved already here: https://github.com/Meeds-io/social/commit/1f0ae4767460570e92478ff92d270a161b1f2aff

> feat: Improve TranslationService API with Object Id of Type String - [MEED-9632](https://builders.meeds.io/portal/meeds/tasks/taskDetail/9632) - https://github.com/Meeds-io/meeds/issues/3608

Thus I think we may now reuse the Setting Name. WDYT ?

rdenarie: I had a version problem on my side.
It is working with the CMS SETTING NAME

Resolved

webapp/src/main/webapp/vue-apps/common/components/ExoDrawer.vue
rdenarie: As discussed together, I'll add a paramter in exoDrawer, autofocusFirstInput, with default value = `true`

boubaker: Yes, this may require some test on existing drawers (especially on mobile for Activity Stream Drawer). I think the focus will jump from the first input into the Rich Editor input. Can you check it please @srenault-meeds  ?

boubaker: I thin even that it would be better to split concerns here and not couple the `Generic Components` accessibility with the current MIP. WDYT ? This way, we can take the time to study its impact.

webapp/src/main/webapp/vue-apps/login-form/main.js
```diff
@@ -0,0 +1,45 @@
...
+ * Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import './initComponents.js';
+
+//getting language of the PLF
+const lang = typeof eXo !== 'undefined' ? eXo.env.portal.language : 'en';
```
boubaker:
```suggestion
const lang = eXo?.env?.portal?.language || 'en';
```

webapp/src/main/webapp/vue-apps/login-internal-onboarding/components/InternalOnboardingExpired.vue
```diff
@@ -19,8 +19,8 @@
 
 -->
 <template>
-  <v-card flat>
-    <v-card-title class="primary--text text-font-size px-0">
+  <v-card flat class="transparent">
```
boubaker: Attrs with attribute value goes before boolean attrs without value (Just a Hint)
+ you can use `color` on `v-card` instead of `class`

```suggestion
  <v-card color="transparent" flat>
```

webapp/src/main/webapp/vue-apps/login-internal-onboarding/components/InternalOnboardingResetForm.vue
```diff
@@ -173,6 +177,33 @@ export default {
     toggleConfirmShow() {
       this.showConfirmPassword = !this.showConfirmPassword;
     },
-  },
+    submitForm() {
+      let url = `${eXo.env.portal.context}/${eXo.env.portal.selectedNodeUri}?`;
+      for (const [key, value] of new URLSearchParams(window.location.search)) {
+        url += `${key}=${value}&`;
+      }
+      const body = `action=resetPassword&username=${encodeURIComponent(this.username)}&password=${encodeURIComponent(this.password)}&password2=${encodeURIComponent(this.confirmPassword)}&captcha=${encodeURIComponent(this.captcha)}`;
```
boubaker: Use `FormData.append`+ `new URLSearchParams(formData).toString()`, see it [here](https://github.com/Meeds-io/social/blob/develop/webapp/src/main/webapp/vue-apps/common/js/SpaceService.js#L193-L194)

webapp/src/main/webapp/vue-apps/login-internal-onboarding/main.js
```diff
@@ -37,16 +35,16 @@ const urls = [
   `/social/i18n/locale.portal.login?lang=${lang}`
 ];
 
-export function init(params) {
+export function init(id,params) {
```
boubaker:
```suggestion
export function init(id, params) {
```

webapp/src/main/webapp/vue-apps/login-form/components/LoginForm.vue
boubaker: Thanks, seems fixed now.

rdenarie: Not completly, I removed the css, but did not update the vue.
I have to finish this one

webapp/src/main/webapp/vue-apps/common/components/ExoDrawer.vue
srenault-meeds: Hello, 

I agree to make the autofocus to first input. It is better for the UX actually. Indeed, it is painful for users to click again and again to finally access the input (for desktop and for mobile too).
And if this is good for accessibility, then let's keep it. Thanks.

Regarding the flag to make the autofocus activated or not, I confirm TRUE must be the default one. And then, it can be reviewed for some drawer in case it needs to be turned to FALSE.

Hope that helps

srenault-meeds: FYI @boubaker we will remove this commit from the MIP and it will be done as a separated maintenance issue

boubaker: ok, Thanks.

webapp/src/main/webapp/vue-apps/login-form/main.js
rdenarie: done

webapp/src/main/webapp/vue-apps/login-internal-onboarding/components/InternalOnboardingExpired.vue
rdenarie: done

webapp/src/main/webapp/vue-apps/login-internal-onboarding/components/InternalOnboardingResetForm.vue
rdenarie: done for this file. Also applied to other concerned vue component

webapp/src/main/webapp/vue-apps/login-internal-onboarding/main.js
rdenarie: done for this file, also applied for other vue components

webapp/src/main/webapp/vue-apps/common/components/ExoDrawer.vue
rdenarie: Commit removed.

webapp/src/main/webapp/vue-apps/login-form/components/LoginFormSettingsDrawer.vue
rdenarie: Can you detail what you mean ? 

In fact, I already use async and await when calling saveTranslationSynchronously to be sure that first saveTranslation request is done before doing the next promises.

boubaker: It was just a proposition to make the code more clear and understandable in fact. But if not used to deal with `async` and `await`, then no problem, I 'll just resolve this remark and let you read more about it for next MIPs.

webapp/src/main/webapp/vue-apps/login-form/components/LoginForm.vue
rdenarie: I tried to replace the `v-btn` by a `v-card`. 
It is not a success. It creates a lot of code to reproduce the button behaviour, hover animation is still missing, it requires to have aria-pressed attributes in addition to role=button to be a11y compatible, and so on.

The current pb here, it that, by convention, Vue do not want to have button labels on 2 lines or more. They argue that button should stay short to preserve ergonomy. I'm completly ok with that point, but, in some language, like DE, short english word is a long german word. 

I'm also completly ok with the fact that we must ban specific css in vue file.

But, for me, it is a very specific case, and I think we should have an exception for LoginForm. I manage to reduce the css to only one instruction : I need to add max-width="100%" on 'v-btn__content'

boubaker: you can add in `vuetify-all.less` ([here](https://github.com/Meeds-io/platform-ui/blob/develop/platform-ui-skin/src/main/webapp/skin/less/vuetify/vuetify-all.less#L280-L284)) any special behavior that you which to change on default `v-btn` coming from Vuetify. We already made some overrides there.

rdenarie: are you sure that we should apply on all v-btn of the platform ?

boubaker: by example:

```css
.v-btn.mutliple-lines {
  .v-btn__content {
    max-width: 100%;
  }
}
```

this way, you can add `class="mutliple-lines"` in your `v-btn` without changing the default behavior on other buttons.

rdenarie: ok, deal. 
I do that.

Thanks

rdenarie: Done

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
```diff
@@ -0,0 +1,712 @@
...
+import java.util.Locale;
+import java.util.Random;
+
+@RestController
+@RequestMapping("/login")
+@Tag(name = "/login", description = "Manage User action around login")
+public class LoginRest {
+
+  @Autowired
+  private PasswordRecoveryService passwordRecoveryService;
```
boubaker: please format the file and use the following order for attributes definition:
1. Static
2. Public
3. Protected
4. Private

+ Attributes before Methods.

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
```diff
@@ -0,0 +1,712 @@
...
+  @Produces("image/png")
+  @Operation(summary = "Create a captcha image", method = "GET")
+  @ApiResponses(value = {
+      @ApiResponse(responseCode = "200", description = "Request fulfilled"),
+  })
+  public ResponseEntity<byte[]> serveCaptcha(HttpServletRequest request,
+                                             @Parameter(description = "Captcha name")
+                                             @RequestParam("name")
+                                             String name) {
+    HttpSession session = request.getSession();
```
boubaker, line 191: Please ensure to split the Business Logic into Service Layer + Rest Layer. The Rest Layer should be kept as much simple as possible to access Contextual information (like Current User) and pass all into a Service Layer where we will have all the Business Logic

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
```diff
@@ -0,0 +1,98 @@
...
+
+  private       TranslationService translationService;
+
+  private       AttachmentService  attachmentService;
+
+  @Override
+  protected void setViewRequestAttributes(String name, RenderRequest request, RenderResponse response) {
+    PortletPreferences preferences = request.getPreferences();
+    String initTranslationIdentifier = preferences.getValue(DATA_INIT_PREFERENCE_NAME,null);
+    if (initTranslationIdentifier != null) {
```
boubaker, line 47: Why creating a new Translation if it's not an `export` operation ? IHO, we should keep it simple like made in other Portlets.

webapp/src/main/webapp/vue-apps/common/js/loginService.js
```diff
@@ -0,0 +1,104 @@
+export function verifyToken(token, tokenType) {
+  let url = `/social/${eXo.env.portal.rest}/login/verify?token=${token}`;
+  if (tokenType) {
+    url = url.concat(`&tokenType=${tokenType}`);
```
boubaker: User `FormData.append` to ensure that the URL is correctly encoded.

webapp/src/main/webapp/vue-apps/common/js/LoginService.js
```diff
@@ -0,0 +1,104 @@
+export function verifyToken(token, tokenType) {
```
boubaker, line 1: Please rename the file to start with Capital Letter

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
rdenarie: To be able to change the translation on a page, without changing the translation of all instances. 

The translation is copied from the initial instance. But at the moment the portlet is added on a page, it have is own translation.

webapp/src/main/webapp/vue-apps/login-external-onboarding/components/ExternalOnboarding.vue
```diff
@@ -19,27 +19,70 @@
...
-    },
-  },
+  data: () => ({
+    email: '',
+    action: '',
+    token: ''
+  }),
   created() {
-    document.title = this.$t('externalOnboarding.pagetitle');
+    this.token = new URLSearchParams(window.location.search).get('token');
```
boubaker:
```suggestion
    const searchParams = new URLSearchParams(window.location.search);
    this.token = searchParams.get('token');
    this.action = searchParams.get('action');
```

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
rdenarie: done

webapp/src/main/webapp/vue-apps/common/js/LoginService.js
rdenarie: done

webapp/src/main/webapp/vue-apps/common/js/loginService.js
rdenarie: done

webapp/src/main/webapp/vue-apps/login-external-onboarding/components/ExternalOnboarding.vue
rdenarie: done

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
rdenarie: This endpoint creates a random captcha image. 

I'm completely OK with the rule about no BusinessLogic in Rest Layer. 

But in this case, I think we should let this function in REST Layer : 

- Firstly, I estimate that there is no real BusinessLogic : no interaction with any other Services, no real condition to create the image or not.
- Secondly : if I create a Service Layer, it will contains only this function.

So on this case, I think I prefer to keep as is.

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
```diff
@@ -0,0 +1,763 @@
...
+                                            String token,
+                                            @Parameter(description = "Token Type")
+                                            @RequestParam(name = "tokenType", required = false, defaultValue = "")
+                                            String tokenType) {
+    String username = passwordRecoveryService.verifyToken(token, tokenType);
+    if (username != null) {
+      try {
+        String email = finishUserCreation(username, request, token, request.getLocale());
+        cleanCapcha(request.getSession(),"email-validation");
+        return ResponseEntity.status(HttpStatus.FOUND).header(LOCATION_HEADER, LOGIN_PATH + "?email="+email).build();
```
boubaker, line 169: Can we avoid redirecting with user Email in HTTP URL (personal data inside the URL will be in clear text, even when using SSL)

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
```diff
@@ -0,0 +1,763 @@
...
+  }
+
+  @PostMapping(path = "/finishRegistration")
+  @Produces(MediaType.APPLICATION_JSON)
+  @Operation(summary = "Finalize external user registration", method = "POST")
+  @ApiResponses(value = {
+      @ApiResponse(responseCode = "200", description = "Request fulfilled"),
+  })
+  public ResponseEntity<String> finishRegistration(HttpServletRequest request,
+                                            @Parameter(description = "Token")
```
boubaker: Code format

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
```diff
@@ -0,0 +1,763 @@
...
+                                                  .put(ERROR_MESSAGE_PARAM, errorMessageName).toString());
+    }
+
+
+    try {
+      HttpSession session = request.getSession();
+      if (!usernameToken.contains("@")) {
+        //username from token is not an email
+        //so we have not verify the email
+        //send accountVerificationEmail
```
boubaker: We may need to explicitly split those two concerns into two dedicated methods to make it clearer, WDYT ?

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
```diff
@@ -0,0 +1,763 @@
...
+                                              @RequestParam("password")
+                                              String password,
+                                              @Parameter(description = "Confirm Password")
+                                              @RequestParam("password2")
+                                              String password2,
+                                              @Parameter(description = "token")
+                                              @RequestParam("token")
+                                              String token) {
+
+    String usernameToken = passwordRecoveryService.verifyToken(token, CookieTokenService.FORGOT_PASSWORD_TOKEN);
```
boubaker, line 492: Need to enhance this with an OTP code some day using `io.meeds.web.security.service.OtpService`. The code may be sent at the same time than the URL. Let's not change it for now, but will have to do it to secure the Reset Password mechanism.

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
```diff
@@ -0,0 +1,763 @@
...
+                                            @RequestParam("password2")
+                                            String password2,
+                                            @Parameter(description = "token")
+                                            @RequestParam("token")
+                                            String token,
+                                            @Parameter(description = "captcha")
+                                            @RequestParam("captcha")
+                                            String captcha) {
+
+    String usernameToken = passwordRecoveryService.verifyToken(token, "onboard");
```
boubaker, line 536: Same remark for OTP code flow usage, but will be treated apart in a dedicated enhancement.

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
boubaker: Yes, for the Captcha no problem (Even I prefer using an Utils Class rather than adding the source code in REST, in order to make it very light). It was a remark about all the `private` methods in the Bottom of the class, which holds some validation Business Logic:
- `findUser`
- `validateUserAndPassword`
- `validateUsernameAndPassword`
- `validateUserFullName`
- `validateEmail`
- `isExistsEmail`
- `sendOnboardingEmail`
- `generateUsername`
- `deleteFromInternalUsersGroup`
- `addToExternalUsersGroup`
- `generateUserDetailCredential`
- `finishUserCreation`
- `generateUserFromCredential`

All of those methods are business logic and should go to a service where we can reuse if we will do the same flow with other endpoints than REST: (Command Line, AI Tool, JSON RPC for MicroServices ...). The REST should get params + context info => pass them as is without thinking nor knowing what is it about + catch exception for response code specificity + format the output in HTTP response.

But no problem, let's keep it as is, we will have to rework some other REST Endpoints already but this one increases a bit the technical debt.

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
rdenarie: The idea is to prefill the login form with the user email.

Before modification, after email validation process, we auto-login the user. 
We the new architecture based on LoginRest, I'm not able to auto-login cleanely, and in addition, I'm not sure that it is a good security practice. 
So I change to redirect on login form, with identifier prefilled.

boubaker: you can use the `sessionStorage` (not `localStorage`) to preserve the username/email to automatically login with in fact.

rdenarie:
> 

    To be able to change the translation on a page, without changing the translation of all instances.
    All instances are using the same CMS Setting Name ?

(I was not able to answer directly to your comment)

When a portlet instance is create, a CMS Setting Name is generated. Translations done in this instance are associated with this CMS Settings Name

When the instance is added in a page, the LoginFormPortletInstancePreferencePlugin.generatePreferences returns the CMS Setting Name as DATA_INIT_PREFERENCE_NAME.
And then, the DATA_INIT_PREFERENCE_NAME is used to get back translations, and copy it to the new CMS Setting Name of the portlet.

This allow to get back existing translation if user add it in the portlet instance, and allow then update the new portlet on a page without modifying the initial instance.

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
rdenarie: done

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
rdenarie: done

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
rdenarie: sessionStorage is client side, right ?

I'm not sure to understand what you have in mean.
Can you explain ?

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
rdenarie: When a portlet instance is create, a CMS Setting Name is generated. Translations done in this instance are associated with this CMS Settings Name

When the instance is added in a page, the LoginFormPortletInstancePreferencePlugin.generatePreferences returns the CMS Setting Name as DATA_INIT_PREFERENCE_NAME.
And then, the DATA_INIT_PREFERENCE_NAME is used to get back translations, and copy it to the new CMS Setting Name of the portlet.

This allow to get back existing translation if user add it in the portlet instance, and allow then update the new portlet on a page without modifying the initial instance.

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
boubaker: When the user registers and then the registration is ok (It's may be not using REST ?) then registration flow returns an error or a success. If error => display the error message right away without refreshing the page. If success => the entered email can be used to automatically send a user login Request using the entered login + password. WDYT ?
In client side, the business logic of navigation flow is controlled by JS Services and not in server side (which was the Stateful mode used before).

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
boubaker: It appears not yet in fact.

```suggestion
  public ResponseEntity<String> finishRegistration(HttpServletRequest request,
                                                                                          @Parameter(description = "Token")
```

rdenarie: Seems ok now on my side :
image

Can you recheck ?

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
rdenarie: to be able to finalize the merge, can we create this as other task and treat it after the merge ?
Thanks

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
boubaker: ok, may be Github doesn't display well spaces, strange because the `verifyToken` is well displayed.
I mark this as resolved.

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
boubaker: I don't think that we can de-prioritize this since it will never be reworked after the merge. This is an unusual flow which may make the maintenance harder. IMHO, we need to make it clearer for maintainers, else this same code will have to be reworked and worked around each time it's unclear for a developer.

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
boubaker: Sorry I always see an if => sendAccountVerificationEmail with else => createUser, switch whether the Token contains an `@` or not.

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
rdenarie: I have few task to treat directly after the merge, before the 7.1 final release. 
So I can take this one during this set of task

That's why I propose to make it after the merge

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
rdenarie: Im sorry but I dont understand what you except.
In your préviens message you require to Split into 2 methods.
I created 2 methods : one for e-mail validation, one for direct création.

As in both case we are in a "register" use case, I prefer to keep only one register method in loginRest

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
boubaker: Issue added here https://github.com/Meeds-io/meeds/issues/3734

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
boubaker: Issue added here Meeds-io/meeds#3734

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
boubaker: We have to differentiate two different flows with two different methods. The outcome of `sendAccountVerificationEmail` is different from `createUser`. Thus, it's unclear why having both different flows in a single method. It's may clear for you, but unfortunately not clear enough for a developer willing to maintain the code. IMO, one REST method = One responsibility, the FrontEnd has to choose either the first method or the second one.

rdenarie: Split done.
There are now 2 methods : 
- register
- VerifyEmail

boubaker: Thanks :+1:

webapp/src/main/java/io/meeds/social/portlet/CmsPortletWithMetadata.java
boubaker: Since the `DATA_INIT_PREFERENCE_NAME` will be used only the first time that a new portlet without `Setting Name` is added into a page. Thus I'm ok with the proposed first solution, until we work on the `export` mechanism.

component/service/src/main/java/io/meeds/social/security/rest/LoginRest.java
boubaker: ok, let's do it that way. Thanks.

boubaker:
All points considered. Thanks.

boubaker (approved):

PR #5041, opened 2025-07-15
fix: Display the total number of comments in activity details and the comment drawer - MEED-9450 - Meeds-io/Meeds#3455

sofyenne:
Prior to this change, the number of activity comments displayed reflected only the top-level comments. This update now displays the total number of comments for an activity, including replies.

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java
@@ -72,6 +72,7 @@ import org.exoplatform.social.core.space.model.Space;
 import org.exoplatform.social.core.space.spi.SpaceService;
 import org.exoplatform.social.rest.api.EntityBuilder;
 import org.exoplatform.social.rest.api.RestUtils;
+import org.exoplatform.social.rest.entity.ActivityCommentCollectionEntity;
 import org.exoplatform.social.rest.entity.ActivityEntity;
 import org.exoplatform.social.rest.entity.ActivitySearchResultEntity;
 import org.exoplatform.social.rest.entity.CollectionEntity;
@@ -513,11 +514,12 @@ public class ActivityRest implements ResourceContainer {
                                                                            sortDescending,
                                                                            offset,
                                                                            limit);
-    CollectionEntity collectionComment = new CollectionEntity(commentsEntity, EntityBuilder.COMMENTS_TYPE, offset, limit);
+    ActivityCommentCollectionEntity collectionComment = new ActivityCommentCollectionEntity(commentsEntity, EntityBuilder.COMMENTS_TYPE, offset, limit);
     if (returnSize) {
       boolean expandSubComments = EntityBuilder.expandSubComments(expand);
       RealtimeListAccess<ExoSocialActivity> listAccess = activityManager.getCommentsWithListAccess(activity, expandSubComments);
       collectionComment.setSize(listAccess.getSize());
+      collectionComment.setTotalCommentsSize(activityManager.getNumberOfAllComments(activityId));
     }
     //
     return EntityBuilder.getResponse(collectionComment, uriInfo, RestUtils.getJsonMediaType(), Response.Status.OK);
--- a/component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java
@@ -89,6 +89,8 @@ import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import io.swagger.v3.oas.annotations.tags.Tag;
 
+import static org.exoplatform.social.rest.api.EntityBuilder.TOTAL_COMMENTS_COUNT;
+
 @Path(VersionResources.VERSION_ONE + "/social/activities")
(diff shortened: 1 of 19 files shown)


component/core/src/main/java/org/exoplatform/social/core/activity/CommentsRealtimeListAccess.java
```diff
@@ -113,6 +113,9 @@ public ExoSocialActivity[] load(int index, int limit) {
    * {@inheritDoc}
    */
   public int getSize() {
+    if (loadSubComments) {
+      return activityStorage.getNumberOfAllComments(existingActivity);
+    }
```
boubaker: It's not the same usage in fact, between loading sub comments of retrieved comment and loading the size of the direct comments of the activity.

Please add a dedicated method in Service layer (without changing this list access) to count all comments of a given Activity + Add a new attribute in `...rest..ActivityEntity` which will be added in `EntityBuilder` only when an `expand` option is present. This new number will be used for displaying the information only.

```suggestion
```

webapp/src/main/webapp/vue-apps/activity-stream/components/comment/list/ActivityCommentsDrawer.vue
```diff
@@ -268,6 +270,7 @@ export default {
         .then(data => {
           let comments = data && data.comments || [];
           this.commentsSize = data && data.size && Number(data.size) || 0;
+          this.totalCommentsSize = this.activity.totalCommentsSize = data?.totalCommentsSize && Number(data.totalCommentsSize) || 0;
```
boubaker: Seprate assignments into two different lines
```suggestion
          this.totalCommentsSize = data?.totalCommentsSize && Number(data.totalCommentsSize) || 0;
          this.activity.totalCommentsSize = this.totalCommentsSize;
```

webapp/src/main/webapp/vue-apps/activity-stream/components/comment/list/ActivityCommentsPreview.vue
```diff
@@ -76,6 +77,7 @@ export default {
       this.$nextTick().then(() => {
         this.comments = this.$activityService.computeParentCommentsList(this.activity.comments) || [];
         this.activity.commentsSize = this.commentsSize = this.activity.commentsCount && Number(this.activity.commentsCount) || 0;
+        this.totalCommentsSize = this.activity.totalCommentsSize = this.activity.totalCommentsCount && Number(this.activity.totalCommentsCount) || 0;
```
boubaker: Seprate assignments into two different lines

boubaker: `this.activity` is a prop which shouldn't be mutated. Handle its update using a generic event triggering using `this.$root.$emit` + `this.$root.$on`

webapp/src/main/webapp/vue-apps/activity-stream/components/comment/list/ActivityCommentsPreview.vue
```diff
@@ -93,6 +95,7 @@ export default {
           this.$nextTick().then(() => {
             this.comments = data && data.comments || [];
             this.activity.commentsSize = this.commentsSize = data && data.size && Number(data.size) || 0;
+            this.totalCommentsSize = this.activity.totalCommentsSize = data?.totalCommentsSize && Number(data.totalCommentsSize) || 0;
```
boubaker: Same here

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java
```diff
@@ -513,11 +514,12 @@ public Response getComments(
                                                                            sortDescending,
                                                                            offset,
                                                                            limit);
-    CollectionEntity collectionComment = new CollectionEntity(commentsEntity, EntityBuilder.COMMENTS_TYPE, offset, limit);
+    ActivityCommentCollectionEntity collectionComment = new ActivityCommentCollectionEntity(commentsEntity, EntityBuilder.COMMENTS_TYPE, offset, limit);
     if (returnSize) {
       boolean expandSubComments = EntityBuilder.expandSubComments(expand);
       RealtimeListAccess<ExoSocialActivity> listAccess = activityManager.getCommentsWithListAccess(activity, expandSubComments);
       collectionComment.setSize(listAccess.getSize());
+      collectionComment.setTotalCommentsSize(activityManager.getNumberOfAllComments(activityId));
```
boubaker, line 526: You may need to add this with a dedicated condition `if (expandFields.contains(TOTAL_COMMENTS_COUNT)) {}` as made for activity.

webapp/src/main/webapp/vue-apps/activity-stream/components/comment/list/ActivityCommentsPreview.vue
sofyenne: done

webapp/src/main/webapp/vue-apps/activity-stream/components/comment/list/ActivityCommentsPreview.vue
sofyenne: done

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java
sofyenne: done

boubaker (approved):

PR #5046, opened 2025-07-17
feat: Add extensions for portlet login form - meeds-io/MIPs#203 - MEED-9363

rdenarie:
This commit add multiproviders and registration to login form


webapp/src/main/webapp/WEB-INF/jsp/portlet/loginForm.jsp
```diff
@@ -2,24 +2,65 @@
...
+                    .forEach(paramsExtension -> {
+                      Map<String, Object> extendedParams = paramsExtension.extendParameters(controllerContext,
+                                                                                            LoginHandler.LOGIN_EXTENSION_NAME);
+                      if (MapUtils.isNotEmpty(extendedParams)) {
+                        extendedParams.forEach((key, value) -> {
+                          try {
+                            params.put(key, value);
+                          } catch (Exception e) {
+                            // Handle potential JSON exceptions
+                          }
```
ahamdi, line 47: is the try/catch needed ?

webapp/src/main/webapp/vue-apps/login/components/provider/LoginProvidersMenu.vue
```diff
@@ -29,9 +29,8 @@
           :aria-label="$t('UILoginForm.label.moreProviders')"
           min-width="auto"
           color="primary"
-          class="pa-4"
-          large
           outlined
+          style="background-color:white"
```
ahamdi: shouldn't this use some CSS class and take a variable for branding ?

webapp/src/main/webapp/WEB-INF/jsp/portlet/loginForm.jsp
rdenarie: yes it is necessary

webapp/src/main/webapp/vue-apps/login/components/provider/LoginProvidersMenu.vue
rdenarie: no. The branding apply a color for the background of the v-card, but the background for the button must be white : 
image

boubaker: yes, you can use `class="white"` to force using white background or better `class="btn"`

rdenarie: I tried proposed solution, but the rendering is not the same : 

with style : 
withStyle

With class btn
withClassBtn

with class white
withClassWhite

boubaker: Yes, that's normal, the `class="btn"` is the main style of normal buttons of the platform. We shouldn't change to make a specific UX for button for logins.
@srenault-meeds WDYT ?

srenault-meeds: Yes I confirm, we should use class="btn"

rdenarie: PR Updated

ahamdi (approved):

PR #5053, opened 2025-07-21
feat: Search activity in multiple spaces - MEED-9331 - Meeds-io/MIPs#211

sofyenne:


component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java
```diff
@@ -999,10 +999,10 @@ public Response searchActivities(
                                    @QueryParam(
                                      "tags"
                                    ) List<String> tagNames,
-                                   @Parameter(description = "Space id used to search activities", required = false)
+                                   @Parameter(description = "spaces used to search activities", required = false)
                                    @QueryParam(
-                                     "spaceId"
-                                   ) String spaceId,
+                                     "spaces"
+                                   ) String spaces,
```
boubaker, line 1005:
```
                                   ) List<String> spaceIds,
```

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java
```diff
@@ -999,10 +999,10 @@ public Response searchActivities(
                                    @QueryParam(
                                      "tags"
                                    ) List<String> tagNames,
-                                   @Parameter(description = "Space id used to search activities", required = false)
+                                   @Parameter(description = "spaces used to search activities", required = false)
                                    @QueryParam(
-                                     "spaceId"
-                                   ) String spaceId,
+                                     "spaces"
```
boubaker, line 1004: Keep the same name please, with each spaceId in a query String apart
```
                                     "spaceId"
```

PR #5065, opened 2025-07-23
Merge Search in multiple spaces Meeds-io/MIPs#211

### LINKED ISSUE #211: Search in multiple spaces
## Rationale

It is now possible to search contents only in one space.
It appears that search in multiple spaces would be more than interesting.
Indeed, when it comes to search for something, end-users reported that they can hesitate between different spaces. 

Hence, to avoid them to search in one space then in another, this MIP will provide a way to search in multiple spaces at once.

#### Option to search in multiple spaces

Given I open the search
When I want to open the spaces selection
Then I can add multiple spaces

Given I choose a space
Then the search results are updated

Given I want to add other spaces or I want to update the spaces selection
Then the search results are updated accordingly

**Precision: searching from a space still means the search is filtered to the current space**

#### Enabling this option for any content, favorite and tags

When searching in multiple spaces,
Then contents are filtered to it 
Including favorite and tags

Image

#### Gamification
NA

#### Notifications
NA

#### Analytics
NA

#### Unified Search
The filter option will be updated to allow users to filter search results by multiple spaces.

- No change on Search app
- No change in stored data

### Security 
The user shouldn't be allowed to search in spaces where they are not identified as a member or a manager.

### Extensibility 
N/A

### Configurability 
N/A

### Upgradability
N/A

### Existing Features
N/A

### Feature Flags

N/A

N/A

### Security

A check on each Search connector has to be made on the list of designated space identifiers to ensure that search results do not include unauthorized elements. This check has to be made in Service Layer.

### Access 
   
The Search application has to be changed in order to allow filtering the search result to multiple spaces

### Services & processing

 Each Search Connector REST and Service layer has to be modified in order to include a Set of Space Ids to filter rather   than a single space Id to use in filter.

### Data and persistence
   The following search connectors has already links to space identifiers persisted in their Elasticsearch data:
- Space
- Notes
- News
- Activity
- Tasks
- Quests
- Profile

sofyenne:
This MIP provides an option for users to filter the search by multiple spaces.
Resolves https://github.com/Meeds-io/MIPs/issues/211


component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java
```diff
@@ -1002,7 +1002,7 @@ public Response searchActivities(
                                    @Parameter(description = "Space id used to search activities", required = false)
                                    @QueryParam(
                                      "spaceId"
-                                   ) String spaceId,
+                                   ) List<String> spaceIds,
```
boubaker:
```suggestion
                                   ) List<Long> spaceIds,
```

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRest.java
```diff
@@ -249,7 +250,7 @@ public Response getUsers(@Context UriInfo uriInfo,
                            @Parameter(description = "User status to filter online users, ex: online") @QueryParam("status") String status,
                            @Parameter(description = "User type to filter, ex: internal, external") @DefaultValue("internal") @QueryParam("userType") String userType,
                            @Parameter(description = "Is connected users") @QueryParam("isConnected") String isConnected,
-                           @Parameter(description = "Space id to filter only its members, ex: 1") @QueryParam("spaceId") String spaceId,
+                           @Parameter(description = "Space id to filter only its members, ex: 1") @QueryParam("spaceId") List<String> spaceIds,
```
boubaker:
```suggestion
                           @Parameter(description = "Space id to filter only its members, ex: 1") @QueryParam("spaceId") List<Long> spaceIds,
```

component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java
sofyenne: done!

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRest.java
sofyenne: done

boubaker (approved):

PR #5068, opened 2025-07-24
fix: multivalue dropdown attribute required but never checked - MEED-9482 - meeds-io/meeds#3485

akhanfir:
before this change, when create new profile attribute dropdown which and has set of drop-down values as required then userX changes his personal information, dropdown attribute isn't marked as required and if dropdown is empty, the profile info drawer is saved. After this change, as required, dropdown attribute a asterisk displayed and the profile info drawer is not saved if dropdown isn't filled + focus on required attribute + display message.


rdenarie:
I think this should not be in the commit message

> To fix this problem, reset minStartTime and minEndTime if the selected date is no longer today

Can you remove it

Jihed525:
> I think this should not be in the commit message
> 
> > To fix this problem, reset minStartTime and minEndTime if the selected date is no longer today
> 
> Can you remove it

I have applied the requested update on original commit description, can you check @rdenarie if it is ok ?

rdenarie (approved):

PR #5069, opened 2025-07-24
feat: Add an event for the creation space from the public site - MEED-9341 - Meeds-io/MIPs#199

MayTekayaa:
This change will  add an event for the creation space from the public site.


boubaker (approved):

boubaker:
@MayTekayaa, this was previously approved can you rebase it and metge into MeedsV2 FB please

PR #5091, opened 2025-08-05
Fix swagger configuration

rdenarie:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <artifactId>maven-parent-pom</artifactId>
     <groupId>io.meeds</groupId>
-    <version>28-M02</version>
+    <version>28-SNAPSHOT</version>
     <relativePath />
   </parent>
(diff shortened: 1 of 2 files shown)


ahamdi (approved):

pom.xml
```diff
@@ -22,7 +22,7 @@
   <parent>
     <artifactId>maven-parent-pom</artifactId>
     <groupId>io.meeds</groupId>
-    <version>28-M02</version>
```
hbenali: We don't use SNAPSHOT versions for parent POMs, as they interfere with the release process. Once the new parent POM version is propagated to these modules, this PR will address the current compatibility issues and can be merged safely. Thus, please cancel this change

rdenarie: Hi,
This PR is still a test, as it need a M03 version for maven-parent-pom.

I prepare it, and let it in a waiting state until my return from holidays. I have PR for all other modules to create.
I put it in draft.

PR #5100, opened 2025-08-12
fix: RTL bad menu display of activity stream post options - MEED-9547 - meeds-io/meeds#3558

akhanfir:
before this change, when access to any post in stream page and click on the three points button on the top left of the page, the menu appears with bad display. To fix this problem, change pl-3 to ps-3. After this change, on RTL mode, the menu option appears with the right display.


azayati (changes requested):
@akhanfir can you modify the title to be more significant ?

Jihed525:
> @akhanfir can you modify the title to be more significant ?

is it ok now @azayati ?

azayati (approved):

PR #5152, opened 2025-09-01
fix: Invite again a user previously invited or requested to join - MEED-9553 - meeds-io/meeds#3564

akhanfir:
Before this change, when inviting registered users or receiving a request from them to join your space, even though the UI should help space admins manage pending invitations/requests, it was not possible to send a new invitation to registered users who had already been invited or had already requested to join the space. To resolve this issue, remove the user type check condition. After this change, as a space admin, I can send an invitation to registered users as long as they have not yet joined the space, even if an invitation or request is still pending.


component/service/src/main/java/org/exoplatform/social/service/rest/PeopleRestService.java
```diff
@@ -1,21 +1,21 @@
...
+ * version 3 of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
```
azayati: keep the old license header

azayati (changes requested):

azayati (approved):

PR #5169, opened 2025-09-05
fix: Update shared activity after updating the original one - MEED-9627 - Meeds-io/meeds#3511 

MayTekayaa:
This change will update shared activity after updating the original one.

Resolves: Meeds-io/meeds#3511

### DIFF
--- a/component/core/src/main/java/io/meeds/social/activity/listener/ActivitySharedActionsListener.java
+++ b/component/core/src/main/java/io/meeds/social/activity/listener/ActivitySharedActionsListener.java
@@ -1,3 +1,21 @@
+/**
+ * This file is part of the Meeds project (https://meeds.io/).
+ *
+ * Copyright (C) 2020 - 2025 Meeds Association contact@meeds.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java
```diff
@@ -1143,4 +1142,15 @@ private long computeLastUpdated(ExoSocialActivity activity) {
...
+  private void updateSharedActionsCacheTime(ExoSocialActivity activity) {
+    Set<ActivityShareAction> sharedActions = activity.getShareActions();
+    for (ActivityShareAction action : sharedActions) {
+      for (Long id : action.getSharedActivityIds()) {
+        ExoSocialActivity sharedactivity =  activityManager.getActivity(id.toString());
+        sharedactivity.setCacheTime(System.currentTimeMillis());
+        activityManager.updateActivity(sharedactivity);
+      }
+    }
+  }
```
boubaker: Move this inside a dedicated listener instead of Adding this Business Logic in REST Layer.

MayTekayaa: done!

component/core/src/main/java/io/meeds/social/activity/listener/ActivitySharedActionsListener.java
```diff
@@ -0,0 +1,40 @@
+package io.meeds.social.activity.listener;
```
boubaker, line 19: please add license header

MayTekayaa: added!

boubaker (approved):

PR #5190, opened 2025-09-11
fix: Fix the display of ordered pinned activities - MEED-9683 - Meeds-io/meeds#3620

MayTekayaa:
Prior to this change, pinned activities are not displayed at the top of the feed. This fix will adjust the order of these activities.


boubaker:
I think the bug isn't in backend, you may need to double check wether the option `showPinned` is `true` or not passed Vue to REST to Service Layer.

boubaker:
Besides, the term `Resolves: Meeds-io/meeds#3620` should be used in description only on **develop PR** which will automatically close the issue once merged. Which is not the case when merging into the FB, the issue has to still `Open` until the final PR to develop is created and then merged which should automatically close the associated issue.

azayati (approved):

azayati (changes requested):
@MayTekayaa please consider @boubaker feedback here https://github.com/Meeds-io/social/pull/5190#issuecomment-3284264190

boubaker (approved):

azayati (approved):

PR #5203, opened 2025-09-16
fix: Fix people and users admin pages - MEED-9690 - Meeds-io/meeds#3627

MayTekayaa:
Prior to this change, the people and users admin page didn't work because of the use of an attribute with multiple values in the profile card . This commit will resolve this issue and disable the save button when changing an attribute  to have multi values used in the first, second and third fields in the profile card.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/profileproperty/ProfilePropertyServiceImpl.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/profileproperty/ProfilePropertyServiceImpl.java
@@ -23,7 +23,9 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.commons.lang3.StringUtils;
 import org.picocontainer.Startable;
@@ -91,6 +93,8 @@ public class ProfilePropertyServiceImpl implements ProfilePropertyService, Start
   private List<String>                               excludedQuickSearchProps               = new ArrayList<>();
 
   private List<String>                               excludedAnalyticsIndexProps            = new ArrayList<>();
+
+  private static final String                        USER_CARD_SETTINGS                     = "UserCardSettings";
 
   public ProfilePropertyServiceImpl(InitParams params,
                                     ProfileSettingStorage profileSettingStorage,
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/vue-apps/profile-settings/components/drawers/ProfileSettingFormDrawer.vue
```diff
@@ -476,6 +485,34 @@ export default {
       if (this.isDropdownList) {
         this.setting.propertyType = this.propertyTypes[0];
       }
+    },
+    'setting.multiValued': function ()  {
```
azayati: we can send the information canEditMultiValued from back directly without calling 3 rest services

azayati (changes requested):

component/core/src/main/java/org/exoplatform/social/core/profileproperty/ProfilePropertyServiceImpl.java
```diff
@@ -205,6 +209,9 @@ public void updatePropertySetting(ProfilePropertySetting profilePropertySetting)
         && getUnhiddenableProfileProperties().contains(profilePropertySetting.getPropertyName())) {
       throw new IllegalArgumentException(String.format("%s cannot be hidden", profilePropertySetting.getPropertyName()));
     }
+    if (profilePropertySetting.isMultiValued() && isUserCardFieldSettings(profilePropertySetting.getPropertyName())) {
+      throw new IllegalArgumentException(String.format("%s cannot be multivalued", profilePropertySetting.getPropertyName()));
+    }
```
azayati, line 214:
```suggestion
     if (!canEditMultiValued(profilePropertySetting.getPropertyName())) {
      throw new IllegalArgumentException(String.format("Mulativalued property %s cannot be edited", profilePropertySetting.getPropertyName()));
    }
```
Then you should intercept this exception in rest layer to block the edit

webapp/src/main/webapp/vue-apps/profile-settings/components/drawers/ProfileSettingFormDrawer.vue
```diff
@@ -364,7 +364,7 @@
           {{ $t('profileSettings.button.cancel') }}
         </v-btn>
         <v-btn
-          :disabled="isSaveButtonDisabled || saving || !valid"
```
azayati, line 367: i think that we should disable the input instead of the save button based on canEditMultiValued information

azayati (changes requested):

azayati (approved):

PR #5231, opened 2025-09-25
feat: Create a new re-usable StickyArrowButton common component - MEED-9473 - Meeds-io/MIPs#198

ahamdi:
Create a new re-usable StickyArrowButton common component


boubaker:
Is the Sticky Arrow Button a reusable component ? Should it be named Sticky as it's not ? (See behavior of [StickyPositionElement.vue](https://github.com/Meeds-io/social/blob/develop/webapp/src/main/webapp/vue-apps/common/components/StickyPositionElement.vue) for stickiness)

hakermi:
> Is the Sticky Arrow Button a reusable component ? Should it be named Sticky as it's not ? (See behavior of [StickyPositionElement.vue](https://github.com/Meeds-io/social/blob/develop/webapp/src/main/webapp/vue-apps/common/components/StickyPositionElement.vue) for stickiness)

Yes its a reusable component, maybe the sticky name is not applicable as it uses a fixed position inside its parent container, and as it floats over content, we may call it floatingArrowButton instead wdyt?.

boubaker:
Yes `FloatingArrowButton` will be ok, by renaming the Vue component as well please `floating-arrow-button`

ahamdi:
> Yes `FloatingArrowButton` will be ok, by renaming the Vue component as well please `floating-arrow-button`

Change is applied

hakermi (approved):

PR #5249, opened 2025-10-02
Merge/task 9786 9785 9758 9757 9756 9638

rdenarie:


mkrout (dismissed):

webapp/src/main/resources/locale/portlet/Login_fr.properties
```diff
@@ -1,5 +1,6 @@
 UILoginForm.msg.Invalid-account=Nom d'utilisateur et/ou mot de passe incorrect ou manquant. Veuillez ré-essayer.
 loginForm.settings.editTooltip=Editer les param\u00E8tres
+loginForm.settings.back=Retour \u00E0 la page pr\u00E9c\u00E9dente
```
azayati, line 3: why do we need to encode the special chars here ?

azayati (changes requested):

ahamdi (dismissed):

webapp/src/main/resources/locale/portlet/Login_fr.properties
rdenarie: When I type a translated string into my IDE, I always use encoded char to prevent file encoding pb : sometimes files are UTF-8 encoded, somethimes ISO-8859-1

So, to be sure I encode.

At the end, when the string is translated in crowdin, the file is always correctly encoded, so it does not really matters.

azayati (dismissed):

azayati (approved):

webapp/src/main/resources/locale/portlet/Login_en.properties
```diff
@@ -95,3 +96,7 @@ generalSettings.login.pageTile.on-boarding.label=Onboarding page when users are
 generalSettings.login.pageTile.register.label=Self-registration page
 generalSettings.login.edit.tooltip=Edit Layout
 generalSettings.login.preview.tooltip=Preview Page
+
+
```
azayati: remove that

azayati (changes requested):

webapp/src/main/resources/locale/portlet/Login_en.properties
rdenarie: done

azayati (approved):

PR #5267, opened 2025-10-20
fix: The copy of an attachment copies only the link - MEED-9802 - meeds-io/meeds#3742

rdenarie:
Before this fix, when an attachment is copied from an object to another one, the attachment linked to the both object is the same. So, if you update the attachment in the target object, the initial attachment is removed and the source object have no more the attachment.

To resolve this problem, we ensure to copy the initial attachment and link the destination object with the new attachment.

Another option is modify the "updateAttachment" behaviour. When changing the attachment of an object, the old attachment is removed. The other option consist in keeping the old attachment, but it can lead to uneeded data retention, so this option was not choosen.

Resolves meeds-io/meeds#3742


azayati (changes requested):
@rdenarie need rebase

rdenarie:
done @azayati

azayati (approved):

PR #5286, opened 2025-10-24
fix: error log when saving loginForm settings first time - MEED-9899 - meeds-io/meeds#3790

rdenarie:
Before this fix, when saving the settings (and specially translations) first time there is an error in server log. This error is due to saveTranslations requests which were send in parallele. But, add the metadata does not exists, both request try to create it, which generate a unicity problem The fix ensure to send the first save translation request synchronously, with async/await, and then launch the others one with promises. As the first already create the metadata, the other requests can be sent in parallel

Resolves meeds-io/meeds#3790


mkrout (approved):
Lgtm

PR #5317, opened 2025-10-31
fix: Fix the display of encoded special characters in titles - MEED-9880 - Meeds-io/meeds#3745

MayTekayaa:
This change will fix the display of encoded special characters in titles.

### DIFF
--- a/component/api/src/main/java/io/meeds/social/translation/service/TranslationService.java
+++ b/component/api/src/main/java/io/meeds/social/translation/service/TranslationService.java
@@ -235,6 +235,29 @@ public interface TranslationService {
   }
 
   /**
+   * Saves sanitized Translation Labels for a given Object's field. This will replace any
+   * existing list of translations
+   *
+   * @param objectType Object type for which the Translation Metadata will be
+   *          attached
+   * @param objectId Object unique identifier
+   * @param fieldName Object field
+   * @param labels {@link Map} of Translations with a corresponding label for
+   *          each {@link Locale}
+   * @param username user name updating the list, used for ACL check
+   * @throws IllegalAccessException When user doesn't have write permission to
+   *           object
+   * @throws ObjectNotFoundException When object wasn't found
+   */
+  default void saveRichTranslationLabels(String objectType,
+                                         long objectId,
+                                         String fieldName,
+                                         Map<Locale, String> labels,
+                                         String username) throws IllegalAccessException, ObjectNotFoundException {
+    saveRichTranslationLabels(objectType, objectId == 0 ? null : String.valueOf(objectId), fieldName, labels, username);
+  }
+
+  /**
    * Saves Translation Labels for a given Object's field. This will replace any
    * existing list of translations
    * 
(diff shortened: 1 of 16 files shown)


rdenarie (commented):
I see that you had add a new method in TranslationRestService, and a new method in TranslationService.

I find that it make a lot of duplicated code, specially in TranslationService.

I'm asking if we should not keep only one function in TranslationService, with a new parameter : isRichText:boolean

So you have a new function 
TranslationService.saveTranslationLabels(......, boolean isRichText) 
In function of isRichText value, you can call or not the processLabelLocale

In existing TranslationService.saveTranslationLabels, which have not the isRichText parameter, you ensure to call TranslationService.saveTranslationLabels(......, boolean isRichText)  with isRichText=false

And, in TranslationRestService, you keep both endpoint, calling the correct TranslationService.saveTranslationLabels(......, boolean isRichText) 

I think it will simplify the code, without having a lot of duplicated lines.

WDYT ?

MayTekayaa:
> I see that you had add a new method in TranslationRestService, and a new method in TranslationService.
> 
> I find that it make a lot of duplicated code, specially in TranslationService.
> 
> I'm asking if we should not keep only one function in TranslationService, with a new parameter : isRichText:boolean
> 
> So you have a new function TranslationService.saveTranslationLabels(......, boolean isRichText) In function of isRichText value, you can call or not the processLabelLocale
> 
> In existing TranslationService.saveTranslationLabels, which have not the isRichText parameter, you ensure to call TranslationService.saveTranslationLabels(......, boolean isRichText) with isRichText=false
> 
> And, in TranslationRestService, you keep both endpoint, calling the correct TranslationService.saveTranslationLabels(......, boolean isRichText)
> 
> I think it will simplify the code, without having a lot of duplicated lines.
> 
> WDYT ?

I completely agree, I thought about it but I was just afraid to do it. (since you have asked to make another different endpoint)
We do it like this.

rdenarie:
> > I see that you had add a new method in TranslationRestService, and a new method in TranslationService.
> > I find that it make a lot of duplicated code, specially in TranslationService.
> > I'm asking if we should not keep only one function in TranslationService, with a new parameter : isRichText:boolean
> > So you have a new function TranslationService.saveTranslationLabels(......, boolean isRichText) In function of isRichText value, you can call or not the processLabelLocale
> > In existing TranslationService.saveTranslationLabels, which have not the isRichText parameter, you ensure to call TranslationService.saveTranslationLabels(......, boolean isRichText) with isRichText=false
> > And, in TranslationRestService, you keep both endpoint, calling the correct TranslationService.saveTranslationLabels(......, boolean isRichText)
> > I think it will simplify the code, without having a lot of duplicated lines.
> > WDYT ?
> 
> I completely agree, I thought about it but I was just afraid to do it. (since you have asked to make another different endpoint) We do it like this.

Just keep the 2 endpoints in TranslationRest :)

component/api/src/main/java/io/meeds/social/translation/service/TranslationService.java
```diff
@@ -265,13 +269,15 @@ void saveTranslationLabels(String objectType,
...
    * @throws ObjectNotFoundException When object wasn't found
    */
   default void saveTranslationLabels(String objectType,
                                      long objectId,
                                      String fieldName,
-                                     Map<Locale, String> labels) throws ObjectNotFoundException {
-    saveTranslationLabels(objectType, objectId == 0 ? null : String.valueOf(objectId), fieldName, labels);
+                                     Map<Locale, String> labels,
+                                     boolean isRichText) throws ObjectNotFoundException {
+    saveTranslationLabels(objectType, objectId == 0 ? null : String.valueOf(objectId), fieldName, labels, isRichText);
```
rdenarie:
```suggestion
                                     Map<Locale, String> labels) throws ObjectNotFoundException {
    saveTranslationLabels(objectType, objectId == 0 ? null : String.valueOf(objectId), fieldName, labels, false);
```

component/api/src/main/java/io/meeds/social/translation/service/TranslationService.java
```diff
@@ -230,8 +231,9 @@ default void saveTranslationLabels(String objectType,
                                      long objectId,
                                      String fieldName,
                                      Map<Locale, String> labels,
-                                     String username) throws IllegalAccessException, ObjectNotFoundException {
-    saveTranslationLabels(objectType, objectId == 0 ? null : String.valueOf(objectId), fieldName, labels, username);
+                                     String username,
+                                     boolean isRichText) throws IllegalAccessException, ObjectNotFoundException {
+    saveTranslationLabels(objectType, objectId == 0 ? null : String.valueOf(objectId), fieldName, labels, username, isRichText);
```
rdenarie:
```suggestion
                                     String username) throws IllegalAccessException, ObjectNotFoundException {
    saveTranslationLabels(objectType, objectId == 0 ? null : String.valueOf(objectId), fieldName, labels, username, false);
```

rdenarie (changes requested):

rdenarie (approved):

PR #5386, opened 2025-11-17
fix: Fix non editable fields after importing users for non-root administrators - MEED-9977 - Meeds-io/meeds#3863

MayTekayaa:
Prior to this change, when an admin, other than the super admin, set a profile property as anon-editable field, after importing users from a CSV file, the values of non-editable fields were neither imported nor displayed on their profile pages. This fix allows all values to be imported and sets the data as uneditable for the non-editable properties.


ahamdi (dismissed):

mkrout (approved):

ahamdi:
@MayTekayaa Could you fix the UT ?

MayTekayaa:
> @MayTekayaa Could you fix the UT ?

@ahamdi unit tests are fixed, Javadoc errors still exist (javadoc build success locally)
can you check please?

PR #5398, opened 2025-11-20
fix: Display the see all buttons when focusing - MEED-9939 - Meeds-io/meeds#3831

MayTekayaa:
Prior to this change, the widget component needed to have better accessibility. Therefore, focusin/focusout events were added to make the component more accessible.


mkrout (dismissed):

rdenarie (commented):
Can you modify the commit message ? 

You should explain WHAT is the problem, and HOW you fixed it.

Thanks

mkrout (approved):

PR #5403, opened 2025-11-20
Merge mip227 devmeed

AzmiTouil:


component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1405,4 +1408,86 @@ private String buildPrettyName(String... names) {
...
+
+    Integer subspacesMaxLimit = parentTemplate.getSubspacesMaxLimit();
+    List<String> allowedSubspaceTemplates = parentTemplate.getAllowedSubspaceTemplates();
+
+    SpaceFilter spaceFilter = new SpaceFilter();
+    spaceFilter.setParentSpaceId(parentSpaceId);
+    ListAccess<Space> existingSubspacesAccess = getAllSpacesByFilter(spaceFilter);
+    Space[] existingSubspaces;
+    try {
+      existingSubspaces = existingSubspacesAccess.load(0, -1);
```
boubaker: Use `getSize` rather thn loading all subspaces

```suggestion
      subspacesCount = existingSubspacesAccess.getSize();
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1405,4 +1408,86 @@ private String buildPrettyName(String... names) {
...
+    } catch (Exception e) {
+      throw new SpaceException(Code.ERROR_DATASTORE, "Failed to load subspaces for parent " + parentSpaceId, e);
```
boubaker:
```suggestion
      throw new SpaceException(Code.ERROR_DATASTORE, "Failed to load subspaces count for parent %s".formatted(parentSpaceId), e);
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1405,4 +1408,86 @@ private String buildPrettyName(String... names) {
...
+    }
+
+    int totalSubspaces = existingSubspaces != null ? existingSubspaces.length : 0;
```
boubaker:
```suggestion
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1405,4 +1408,86 @@ private String buildPrettyName(String... names) {
...
+    }
+    Space parentSpace = getSpaceById(parentSpaceId);
+    SpaceTemplate parentTemplate = spaceTemplateService.getSpaceTemplate(parentSpace.getTemplateId());
+    if (parentTemplate == null) {
+      throw new SpaceException(Code.UNKNOWN_SPACE_TEMPLATE,
+                               String.format("Unknown parent space template for space %s", parentSpace.getDisplayName()));
+    }
+
+    Integer subspacesMaxLimit = parentTemplate.getSubspacesMaxLimit();
+    List<String> allowedSubspaceTemplates = parentTemplate.getAllowedSubspaceTemplates();
```
boubaker: Move this variable down where it's used

```suggestion
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1405,4 +1408,86 @@ private String buildPrettyName(String... names) {
             .orElseThrow();
   }
 
+  private void checkSubspaceCreationAllowed(Space space, long parentSpaceId) throws SpaceException {
+    if (parentSpaceId <= 0) {
+      return;
+    }
+    Space parentSpace = getSpaceById(parentSpaceId);
```
boubaker, line 1416: We should add a condition to check if the user is member of the parent space.

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1405,4 +1408,86 @@ private String buildPrettyName(String... names) {
...
+    }
+
+    if (allowedSubspaceTemplates == null || allowedSubspaceTemplates.isEmpty()) {
+      return;
+    }
+
+    long templateId = space.getTemplateId();
+    String matchedRule = null;
+    for (String rule : allowedSubspaceTemplates) {
+      if (rule != null && rule.startsWith(templateId + ":")) {
```
boubaker: Add the format check here before assigning the `matchedRule`, this way you assign only a valid rule

```suggestion
      if (rule != null && rule.startsWith(templateId + ":") && matchedRule.split(":").length == 2) {
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1405,4 +1408,86 @@ private String buildPrettyName(String... names) {
...
+                                             parentSpace.getDisplayName(),
+                                             subspacesMaxLimit));
+    }
+
+    if (allowedSubspaceTemplates == null || allowedSubspaceTemplates.isEmpty()) {
+      return;
+    }
+
+    long templateId = space.getTemplateId();
+    String matchedRule = null;
```
boubaker:
```suggestion
    String matchedRule = allowedSubspaceTemplates.stream()
      .filter(ruld -> rule != null && rule.startsWith(templateId + ":") && matchedRule.split(":").length == 2)
      .findFirst()
      .orElseThrow(() => new SpaceException(Code.SPACE_PERMISSION,
                               "Subspace template '%s' is not allowed under parent template '%s'".formatted(templateId, parentTemplate.getId())));
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1405,4 +1408,86 @@ private String buildPrettyName(String... names) {
...
+        break;
+      }
+    }
+
+    if (matchedRule == null) {
+      throw new SpaceException(Code.SPACE_PERMISSION,
+                               String.format("Subspace template '%s' is not allowed under parent template '%s'",
+                                             templateId,
+                                             parentTemplate.getId()));
+    }
```
boubaker: Check Moved to the Stream method as described above

```suggestion
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1405,4 +1408,86 @@ private String buildPrettyName(String... names) {
...
+
+    String[] parts = matchedRule.split(":");
+    if (parts.length == 2) {
```
boubaker: Not needed anymore, since the check is made above

```suggestion
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1405,4 +1408,86 @@ private String buildPrettyName(String... names) {
...
+      }
+
+      int countForTemplate = 0;
+      if (existingSubspaces != null) {
+        for (Space existing : existingSubspaces) {
+          if (existing.getTemplateId() > 0 && existing.getTemplateId() == templateId) {
+            countForTemplate++;
+          }
+        }
+      }
```
boubaker: Use a `ListAccess.getSize` using a filter on `parentSpaceId` + `templateId` rather than loading all sub spaces and then counting the associated templates one by one. (Imagine having tens or hundreds of Circle Spaces in a single Parent Space, the memory consumption and CPU will increase considerably)

```suggestion
```

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/entity/SpaceEntity.java
```diff
@@ -177,138 +190,18 @@ public class SpaceEntity implements Serializable {
...
-  }
-
-  public void setCategories(List<SpaceCategoryEntity> categories) {
-    this.categories = categories;
-  }
+  @Setter
+  @Getter
+  @ManyToOne
+  @JoinColumn(name = "PARENT_SPACE_ID")
+  private SpaceEntity                              parentSpaceEntity;
```
boubaker: Do we need to map all the Parent Entity ? IMO, we can map it as an id only, since we don't use it in a JPQL. WDYT ?

```suggestion
  private long                              parentSpaceId;
```

component/core/src/test/java/org/exoplatform/social/core/space/spi/SpaceServiceTest.java
```diff
@@ -2589,6 +2583,40 @@ private Space createMoreSpace(String spaceName) {
     return space2;
   }
 
+  @SneakyThrows
+  public void testCreateSubspace() {
```
boubaker: We should have the following Tests IMO:
1. When the user isn't member of the parent space
2. When the `SubspacesMaxLimit` is reached (the global setting)
3. When the `templateId` is missing or not configured as allowed
4. When the `templateId` isn't permitted (Access Permission to user) for the user
5. When the `templateId` specific max limit is reached

component/core/src/test/java/org/exoplatform/social/core/space/spi/SpaceServiceTest.java
```diff
@@ -2589,6 +2583,40 @@ private Space createMoreSpace(String spaceName) {
...
+    spaceTemplate.setSubspacesMaxLimit(2);
+    spaceTemplateService.updateSpaceTemplate(spaceTemplate);
+
+    Space space = new Space();
+    space.setDisplayName("parentSpace");
+    space.setPrettyName("parentSpace");
+    space.setRegistration(Space.OPEN);
+    space.setDescription("description of space parentSpace");
+    space.setVisibility(Space.PRIVATE);
+    space = spaceService.createSpace(space, ROOT_NAME);
```
boubaker: This should throw an exception when the `templateId` isn't specified, isn't it ?

component/core/pom.xml
```diff
@@ -28,7 +28,7 @@
   <name>Meeds:: PLF:: Social Core Component</name>
   <description>Meeds Social Core Component: People and Space</description>
   <properties>
-    <exo.test.coverage.ratio>0.72</exo.test.coverage.ratio>
+    <exo.test.coverage.ratio>0.71</exo.test.coverage.ratio>
```
boubaker: We may need to increase it

```suggestion
    <exo.test.coverage.ratio>0.72</exo.test.coverage.ratio>
```

component/service/src/main/java/org/exoplatform/social/rest/entity/SpaceEntity.java
```diff
@@ -493,4 +493,30 @@ public SpaceEntity setUnreadItems(Map<String, Long> getUnreadItems) {
     return this;
   }
 
+  public SpaceEntity setParentSpaceId(Long parentSpaceId) {
+    setProperty("parentSpaceId", parentSpaceId);
+    return this;
+  }
+
+  public Long getParentSpaceId() {
```
boubaker, line 501: Use same methods as `public long getTemplateId()`

component/service/src/main/java/org/exoplatform/social/rest/impl/space/SpaceRest.java
```diff
@@ -235,6 +235,9 @@ public Response getSpaces( // NOSONAR
                             @Parameter(description = "Excluded space ids", required = false)
                             @QueryParam("excludedId")
                             List<Long> excludedIds,
+                            @Parameter(description = "Parent space id", required = false)
+                            @QueryParam("parentSpaceId")
+                            String parentSpaceId,
```
boubaker, line 240:
```suggestion
                            Long parentSpaceId,
```

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/SpaceDAO.java
```diff
@@ -394,6 +399,12 @@ private void buildPredicates(XSpaceFilter spaceFilter,
       parameterNames.add(PARAM_USER_ID);
     }
 
+    if (spaceFilter.getParentSpaceId() > 0) {
+      suffixes.add("ParentSpaceId");
+      predicates.add("s.parentSpaceEntity.id = :parentSpaceId");
```
boubaker: After deleting the entity mapping as suggested above
 
```suggestion
      predicates.add("s.parentSpaceId = :parentSpaceId");
```

component/core/src/test/java/org/exoplatform/social/core/space/spi/SpaceServiceTest.java
boubaker: Please add a unit Test for Space Filtering by Parent Space Id

boubaker (approved):

PR #5424, opened 2025-11-28
feat: enable suggesting only parent spaces - MEED-10014 - Meeds-io/MIPs#227

sofyenne:


component/core/src/main/java/org/exoplatform/social/core/space/SpaceUtils.java
```diff
@@ -613,6 +616,16 @@ private static String computeSpacePermissionFromTemplate(String p, String groupI
     return SPACE_ADMIN_REFERENCE_NAME.equals(p) ? MANAGER + ":" + groupId : p;
   }
 
+  public static List<Long> getTemplateIdsAllowingSubspaces() {
```
boubaker, line 618: Maybe defining it in `SpaceTemplateService` WDYT ?

sofyenne: done

component/core/src/main/java/org/exoplatform/social/core/space/SpaceUtils.java
```diff
@@ -22,9 +22,12 @@
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
+import io.meeds.social.space.template.model.SpaceTemplate;
```
boubaker:
```suggestion
```

boubaker (approved):

PR #5431, opened 2025-12-03
fix: Fix non editable fields after importing users for non-root administrators - MEED-9977 - Meeds-io/meeds#3863

MayTekayaa:
Prior to this change, when an admin, other than the super admin, set a profile property as anon-editable and multivalued field, after importing users from a CSV file, the values of non-editable and multivalued fields were neither imported nor displayed on their profile pages. This change will fix this problem.


ahamdi:
@MayTekayaa The possibility to import multi-valued fields from CSV was restricted to the system properties for a reason, please check it and do not remove the code.

PR #5456, opened 2025-12-11
fix: Open the link added to image - MEED-9993 - Meeds-io/MIPs#232

MayTekayaa:
This change will open the link added to image.

### DIFF
--- a/webapp/src/main/webapp/vue-apps/component-image-crop/components/ImageCropDrawer.vue
+++ b/webapp/src/main/webapp/vue-apps/component-image-crop/components/ImageCropDrawer.vue
@@ -495,7 +495,7 @@ export default {
       this.mimetype = imageItem?.mimetype || imageItem?.data &&  this.getBase64Mimetype(imageItem?.data) || null;
       this.alternativeText = imageItem?.altText || null;
       this.linkUrl = imageItem?.linkUrl || null;
-      this.linkTarget = imageItem?.linkTarget || true;
+      this.linkTarget = imageItem?.linkTarget === 'false' ? false : true;
       this.format = imageItem?.format || ((this.useFormat || this.customFormat) && 'custom') || 'landscape';
       this.specificFormatSelected = !!imageItem?.format;
       this.$nextTick().then(() => {
--- a/webapp/src/main/webapp/vue-apps/component-image-crop/components/ImageCropDrawer.vue
+++ b/webapp/src/main/webapp/vue-apps/component-image-crop/components/ImageCropDrawer.vue
@@ -228,6 +228,8 @@
             </div>
             <v-switch
               v-model="linkTarget"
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/vue-apps/component-image-crop/components/ImageCropDrawer.vue
```diff
@@ -495,7 +495,7 @@ export default {
       this.mimetype = imageItem?.mimetype || imageItem?.data &&  this.getBase64Mimetype(imageItem?.data) || null;
       this.alternativeText = imageItem?.altText || null;
       this.linkUrl = imageItem?.linkUrl || null;
-      this.linkTarget = imageItem?.linkTarget || true;
+      this.linkTarget = imageItem?.linkTarget === 'false' ? false : true;
```
boubaker: How can we have a `false` value for target ? is it new or old behavior ?

MayTekayaa: it 's old behavior!

boubaker: ok, no problem, let's not waste time on its migration if it's already handled.

webapp/src/main/webapp/vue-apps/component-image-crop/components/ImageCropDrawer.vue
```diff
@@ -228,6 +228,8 @@
             </div>
             <v-switch
               v-model="linkTarget"
+              :true-value="''"
```
boubaker, line 231:
```suggestion
              :true-value="null"
```

webapp/src/main/webapp/vue-apps/component-image-crop/components/ImageCropDrawer.vue
```diff
@@ -495,7 +497,6 @@ export default {
       this.mimetype = imageItem?.mimetype || imageItem?.data &&  this.getBase64Mimetype(imageItem?.data) || null;
       this.alternativeText = imageItem?.altText || null;
       this.linkUrl = imageItem?.linkUrl || null;
-      this.linkTarget = imageItem?.linkTarget || true;
```
boubaker, line 498: This initialization is important, isn't it ?

webapp/src/main/webapp/vue-apps/component-image-crop/components/ImageCropDrawer.vue
MayTekayaa: i kept the empty string value because veutify doesn't accept null as true-value
and i have to keep also '_blank' as a false-value

boubaker (approved):

PR #5467, opened 2025-12-16
fix: Refactor cropper drawer alt attributes - MEED-10058 - Meeds-io/MIPs#232

MayTekayaa:
This change will refactor the cropper drawer alt attributes.

### DIFF
--- a/webapp/src/main/webapp/vue-apps/component-image-crop/components/ImageCropDrawer.vue
+++ b/webapp/src/main/webapp/vue-apps/component-image-crop/components/ImageCropDrawer.vue
@@ -220,8 +220,7 @@
             type="text"
             outlined
             dense
-            mandatory
-            @input="$emit('link-url', linkUrl)" />
+            mandatory />
           <div class="d-flex mb-2">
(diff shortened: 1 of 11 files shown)


webapp/src/main/webapp/vue-apps/component-image-crop/components/ImageCropDrawer.vue
```diff
@@ -262,7 +259,7 @@
           {{ $t('imageCropDrawer.cancel') }}
         </v-btn>
         <v-btn
-          :disabled="!imageData || !isValidLink"
+          :disabled="!imageData || (link && !isValidLink)"
```
boubaker: It's may be better to combine both conditions `link && !isValidLink`. (`isValidLink() { return !this.link || ...`)
WDYT ?

MayTekayaa: I confirm yes, done!

boubaker (approved):

PR #5468, opened 2025-12-16
feat: Allow setting auto cols in the application header when no center content and not in compact mode  - EXO-82846  - Meeds-io/meeds#3937

mkrout:
This commit allows setting auto cols in the application header when no center content and not in compact mode


azayati (changes requested):
it needs a meeds issue IMO

azayati (approved):

PR #5518, opened 2026-01-09
fix: set correct profile page title for accessibility - MEED-10108 - Meeds-io/meeds#3967

hakermi:
Prior to this change, the profile page title displayed "Profile – Company Name", 
which did not include the user's full name and was not fully suitable for accessibility.

This commit changes the page title to display "Profile of {username} – Company Name",
ensuring it includes the user’s full name and improves accessibility for screen readers.

### DIFF
--- a/webapp/src/main/webapp/vue-apps/profile-header/components/ProfileHeader.vue
+++ b/webapp/src/main/webapp/vue-apps/profile-header/components/ProfileHeader.vue
@@ -169,6 +169,9 @@ export default {
     },
     isAdmin() {
       return this.user?.isAdmin;
+    },
+    isProfilePage() {
+      return eXo?.env?.portal?.selectedNodeUri === 'profile';
     }
   },
   watch: {
@@ -192,6 +195,13 @@ export default {
     });
   },
   methods: {
+    setPageTitle() {
+      if (this.isProfilePage) {
+        const companyName = eXo.env.portal.companyName;
+        const fullName = this.user?.fullname;
+        window.document.title = this.$t('profileHeader.page.title', {0: fullName, 1: companyName});
+      }
+    },
     editAvatar() {
       this.imageType = 'avatar';
       this.$nextTick()
@@ -243,6 +253,7 @@ export default {
       return this.$userService.getUser(eXo.env.portal.profileOwner, 'relationshipStatus,settings')
         .then(user => {
           this.user = user;
+          this.setPageTitle();
           return this.$nextTick();
         })
         .catch((e) => {


rdenarie (changes requested):
Can you improve your commit message ?
Thanks

hakermi:
> Can you improve your commit message ? Thanks

The commit message is already clear and describes the change accurately.
I think it’s sufficient as is.

Let’s focus on the fix itself rather than refining the commit message.

rdenarie:
We have guideline about commit messages : https://developer.exoplatform.org/#id-contributions

These rules were defined for important reasons, including code maintainability.
Today, the fix is ​​fresh in memory, so it's easy to understand why the change was made.

However, in 6 months, 1 year, or 3 years, this memory will no longer exist, and the WHY behind the fix will be lost.

That's why we've defined the following in the guidelines:
- Why is this change needed?
- How does this change address the issue?

The commit message is part of the fix; it's not something to overlook.
Therefore, please update the message to answer the two questions outlined in the guidelines.

Thanks

hakermi:
Why is this change needed?
**To improve accessibility, as already indicated in the original commit message.**

How does this change address the issue?
**By setting a correct and accessible page title, as described in the original commit message.**

Even though I’m not fully convinced this was strictly necessary, I’ve slightly reworded the commit message. Does it look better now?

rdenarie:
No, it is not clear.

This is an example of a clear message :
Why is this change needed?
Prior to this change, the page title display "Profile - Company Name", which is not suitable for accessibility. The page title should contains the user full name : "Profile of {username} - Company Name"

How does this change address the issue?
This commit change the page title to display the company name, and the user fullname, in order to improve accessibility

The commit message is here to understand the fix without having to read the code. 
This message "To improve accessibility"
does not explain WHY the modification is needed, but what is the need of the issue. Yes, fixing this issue improve accessibility. But the accessibility was not good because the title displayed does not contains the user fullname.


Same for the question "how" : "By setting a correct and accessible page title" is not clear. Indeed, you set a correct title, it is the aim of the issue. But what is correct title ? 


Once again, all of this should not be see as a constraint, but as a message for the futur. 
When you read commit list (specially in support or maintenance context), it is easier to work when the commit message gives all infos. Then you have not to browse the code to understand.

Thanks

rdenarie (changes requested):
see my last comment

hakermi:
I still feel this discussion grew longer than necessary, but the changes have been applied.

rdenarie (approved):

PR #5555, opened 2026-01-30
feat: make email, call and messaging properties linkable - MEED-10167, MEED-10168 - Meeds-io/MIPs#231

hakermi:
- Convert email, call and messaging profile properties into clickable links
- Enable opening of dedicated native apps (mail client, phone dialer, messaging app)
- Support service numbers (e.g. *123#,, emergency numbers) for call/messaging types
- Preserve existing autolinker behavior for URLs and normal phone numbers
- Ensure safe parsing (no double-linking, no nested anchors, sanitized output)

### DIFF
--- a/webapp/src/main/webapp/vue-apps/profile-contact-information/components/property-type/action/ProfilePropertyInput.vue
+++ b/webapp/src/main/webapp/vue-apps/profile-contact-information/components/property-type/action/ProfilePropertyInput.vue
@@ -1,0 +1,119 @@
+<!--
+ * This file is part of the Meeds project (https://meeds.io/).
+ *
+ * Copyright (C) 2026 Meeds Association contact@meeds.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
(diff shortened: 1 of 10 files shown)


webapp/src/main/webapp/vue-apps/profile-contact-information/components/property-type/action/ProfilePropertyInput.vue
```diff
@@ -0,0 +1,120 @@
...
+    rulesInput() {
+      const rules = [];
+      if (this.isRequired) {
+        rules.push(v => !!v
+            || this.$t('profileContactInformation.message.field.required'));
+      }
+      switch (this.property?.propertyType) {
+      case 'call':
+      case 'sms':
+        rules.push(v => this.phoneRegex.test(v)
```
boubaker: Maybe use `v && Autolinker.parse(v, {___PARAMS___})?.[0]?.getAnchorHref?.()`. If it returns a value, then it's linkable else not, WDYT ?

hakermi: The problem that it should validate service numbers like *123# or  like198 for emergency numbers, thus i choose a custom regex, because autolinker only parses real phone numbers not local service numbers

boubaker: Thanks for the explanation, yes it's a good idea. But just to be sure that we use the same validation process as when processing the Field Value in `autolinker directive`. We should use the same way of processing in Input Value validation and processing. In other words, if we combine `Autolinker & Regex` in Value display, we should use the same for Value validation. I propose to centralize this in `Utils.js` so that we centralize the Phone, SMS, Email and URL processing algorithm, WDYT ?

hakermi: Yes, Done

boubaker (approved):

PR #5563, opened 2026-02-04
feat: Display of phone & email actions in profile popover - MEED-10175 - Meeds-io/MIPs#231

hakermi:
Display of phone & email actions in profile popover

### DIFF
--- a/webapp/src/main/webapp/vue-apps/popover/components/UserPopoverContent.vue
+++ b/webapp/src/main/webapp/vue-apps/popover/components/UserPopoverContent.vue
@@ -106,6 +106,8 @@ export default {
         identityId: this.identity?.username,
         username: this.identity?.username,
         fullname: this.identity?.fullname,
+        displayedEmail: this.identity?.displayedEmail,
+        displayedPhone: this.identity?.displayedPhone,
         identityEnabled: this.identity?.enabled,
         identityDeleted: this.identity?.deleted,
         compactDisplay: true,
@@ -165,7 +167,7 @@ export default {
           extension.init(container, this.username);
         }
       }
-    },
+    }
   }
 };
 </script>
--- a/webapp/src/main/webapp/vue-apps/profile-settings/extensions/components/EmailPropertyPopoverButton.vue
+++ b/webapp/src/main/webapp/vue-apps/profile-settings/extensions/components/EmailPropertyPopoverButton.vue
@@ -1,0 +1,46 @@
+<!--
+ * This file is part of the Meeds project (https://meeds.io/).
+ *
+ * Copyright (C) 2026 Meeds Association contact@meeds.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+-->
+<template>
+  <v-btn
+    :href="emailUrl"
+    :aria-label="$t('UserProfilePopup.button.sendEmail')"
+    :title="$t('UserProfilePopup.button.sendEmail')"
+    icon>
+    <v-icon size="18">
+      fas fa-envelope
+    </v-icon>
+  </v-btn>
+</template>
+
+<script>
+export default {
+  props: {
+    displayedEmail: {
+      type: String,
+      default: null
+    }
+  },
+  computed: {
+    emailUrl() {
+      return `mailto://${this.displayedEmail}`;
+    }
+  }
+};
+</script>
--- a/webapp/src/main/webapp/vue-apps/profile-settings/extensions/components/PhonePropertyPopoverButton.vue
+++ b/webapp/src/main/webapp/vue-apps/profile-settings/extensions/components/PhonePropertyPopoverButton.vue
@@ -1,0 +1,46 @@
+<!--
+ * This file is part of the Meeds project (https://meeds.io/).
+ *
+ * Copyright (C) 2026 Meeds Association contact@meeds.io
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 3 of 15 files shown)


webapp/src/main/webapp/vue-apps/profile-settings/extensions/components/PhonePropertyPopoverButton.vue
```diff
@@ -0,0 +1,61 @@
...
+  computed: {
+    displayedPhone() {
+      return this.user?.displayedPhone;
+    },
+  },
+  methods: {
+    getUser() {
+      this.$userService.getUser(this.identityId).then(data => {
+        this.user = data;
+      });
```
boubaker: Each REST call is causes a performance regression for popover display. Please consider updating the extension in `UserPopoverContent.vue` rather than retrieving the identity in each button.

```suggestion
```

boubaker: Each REST call is causes a performance regression for popover display. Please consider updating the extension in `UserPopoverContent.vue` rather than retrieving the identity in each button.

```suggestion
```

webapp/src/main/webapp/vue-apps/profile-settings/extensions/components/EmailPropertyPopoverButton.vue
```diff
@@ -0,0 +1,61 @@
...
+  computed: {
+    displayedEmail() {
+      return this.user?.displayedEmail;
+    },
+  },
+  methods: {
+    getUser() {
+      this.$userService.getUser(this.identityId).then(data => {
+        this.user = data;
+      });
```
boubaker: Each REST call is causes a performance regression for popover display. Please consider updating the extension in `UserPopoverContent.vue` rather than retrieving the identity in each button.

```suggestion
```

webapp/src/main/webapp/vue-apps/profile-settings/extensions/components/PhonePropertyPopoverButton.vue
```diff
@@ -0,0 +1,61 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+-->
+<template>
+  <v-btn
+    :href="`tel://${displayedPhone}`"
```
boubaker: Please make the whole value as computed.


```suggestion
    :href="phoneUrl"
```

webapp/src/main/webapp/vue-apps/profile-settings/extensions/components/EmailPropertyPopoverButton.vue
```diff
@@ -0,0 +1,61 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+-->
+<template>
+  <v-btn
+    :href="`mailto://${displayedEmail}`"
```
boubaker: Please make the whole value as computed.

```suggestion
    :href="mailUrl"
```

webapp/src/main/webapp/vue-apps/profile-settings/extensions/main.js
```diff
@@ -0,0 +1,37 @@
...
+  'email-profile-property-popover': EmailPropertyPopoverButton
+};
+
+for (const key in components) {
+  Vue.component(key, components[key]);
+}
+
+export function init() {
+  registerExtensions();
+}
```
boubaker: No need for init method call waiting. 

```suggestion
registerExtensions();
```

webapp/src/main/webapp/vue-apps/profile-settings/extensions/components/EmailPropertyPopoverButton.vue
hakermi: I was waiting this comment to update it :), Done

boubaker: you 're welcome :-)

webapp/src/main/webapp/vue-apps/popover/components/UserPopoverContent.vue
```diff
@@ -166,6 +168,11 @@ export default {
         }
       }
     },
+    getUser() {
+      this.$userService.getUser(this.identityId).then(data => {
+        this.user = data;
+      });
+    }
```
boubaker: We should reuse the `identity` object IMO since it has already the `primaryProperty`, `secondaryProperty`... and if missing, we should add the `emailProperty` and `phoneProperty` values in `EntityBuilder`. This will avoid having to use a second call to `this.$userService.getUser`. WDYT ?

```suggestion
```

hakermi: that object comes from the directive param here:
https://github.com/Meeds-io/social/blob/develop/webapp/src/main/webapp/vue-apps/popover/main.js#L12

and its not complete identity object

hakermi: i will update here then:
https://github.com/Meeds-io/social/blob/develop/webapp/src/main/webapp/vue-apps/common/components/UserAvatar.vue#L434

boubaker: Cool, thanks!
We may miss other places, I let you check the behavior wherever the popover directive is used.

hakermi: Seems the only needed place in case of user popover.
Update done.

boubaker: May be this Method `org.exoplatform.social.core.service.LinkProvider.getProfileLink(String, String)` where it's used for user mention replacements, WDYT ?

hakermi: Updated too

boubaker (approved):

webapp/src/main/webapp/vue-apps/popover/components/UserPopoverContent.vue
boubaker: Thanks :+1:

PR #5598, opened 2026-02-17
feat: Implement space invitation Link generation - MEED-10221 - Meeds-io/MIPs#244

hakermi:
Implement space invitation Link generation


component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -100,36 +97,46 @@
...
 public class SpaceServiceImpl implements SpaceService {
 
-  private static final Log         LOG                   = ExoLogger.getLogger(SpaceServiceImpl.class);
+  private static final Log            LOG                     = ExoLogger.getLogger(SpaceServiceImpl.class);
 
-  private static final int         MAX_SPACE_NAME_LENGTH = 200;
+  private static final int            MAX_SPACE_NAME_LENGTH   = 200;
 
-  private SpaceStorage             spaceStorage;
+  private static final SecureRandom   SECURE_RANDOM           = new SecureRandom();
```
boubaker, line 108: Please use `org.exoplatform.web.security.security.SecureRandomService` instead to get a `SecureRandom` instance
 
```suggestion
  private static final SecureRandom   secureRandom;
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -100,36 +97,46 @@
...
+
+  private SpaceTemplateService        spaceTemplateService;
+
+  private FileService                 fileService;
+
+  private CookieTokenService          cookieTokenService;
+
+  private SpaceLifecycle              spaceLifeCycle         = new SpaceLifecycle();
+
+  private static CodecInitializer     codecInitializer;
```
boubaker, line 138: Since it's a service and normally a `Singleton`, no need to make it static in fact

```suggestion
  private CodecInitializer     codecInitializer;
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -100,36 +97,46 @@
...
+  private static final Log            LOG                     = ExoLogger.getLogger(SpaceServiceImpl.class);
 
-  private static final int         MAX_SPACE_NAME_LENGTH = 200;
+  private static final int            MAX_SPACE_NAME_LENGTH   = 200;
 
-  private SpaceStorage             spaceStorage;
+  private static final SecureRandom   SECURE_RANDOM           = new SecureRandom();
 
-  private SpaceSearchConnector     spaceSearchConnector;
+  private static final int            NONCE_LENGTH            = 12;
```
boubaker, line 110: Please add this parameter provided from XML which should be possible to change by properties as well. When not defined, it can have the default value `12`

```suggestion
  private int            invitationTokenNonceMaxLength;
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1123,6 +1130,23 @@ public void addSpaceListener(SpaceListenerPlugin plugin) {
...
 
+  @Override
+  public String generateInvitationToken(Long spaceId, Long userIdentityId) throws TokenServiceInitializationException {
+    int nonceLength = NONCE_LENGTH;
+    byte[] bytes = new byte[nonceLength];
+    SECURE_RANDOM.nextBytes(bytes);
+    String nonce = URL_ENCODER.encodeToString(bytes).substring(0, nonceLength);
+    String payload = spaceId + ":" + userIdentityId + ":" + nonce;
+    return getCodec().encode(payload);
+  }
```
boubaker, line 1141: May be to make it more secure:
1. add a generated nonce at the beginning
2. make the none length random

```suggestion
  @Override
  public String generateInvitationToken(Long spaceId, Long userIdentityId) throws TokenServiceInitializationException {
    String payload = "%s:%s:%s:%s".formatted(
      generateNonce(),
      spaceId,
      userIdentityId,
      generateNonce());
    return getCodec().encode(payload);
  }

  private String generateNonce() {
    int nonceLength = random.nextInt(1, invitationTokenNonceMaxLength);
    byte[] bytes = new byte[nonceLength];
    SECURE_RANDOM.nextBytes(bytes);
    return URL_ENCODER.encodeToString(bytes).substring(0, nonceLength);
  }
```

component/service/src/main/java/org/exoplatform/social/rest/impl/spacemembership/SpaceMembershipRest.java
```diff
@@ -416,6 +410,46 @@ public Response deleteSpaceMembershipById(
...
+		String authenticatedUser = ConversationState.getCurrent().getIdentity().getUserId();
+
+		Space space = spaceService.getSpaceById(spaceId);
+		if (space == null) {
+			return Response.status(Response.Status.NOT_FOUND).build();
+		}
+
+		if (!spaceService.canManageSpace(space, authenticatedUser)) {
+			return Response.status(Response.Status.UNAUTHORIZED).build();
+		}
```
boubaker, line 437: Please move those checks in Service layer as documented in Best practices guide. The service method can throws `ObjectNotFoundException` and `IllegalAccessException`.

component/service/src/main/java/org/exoplatform/social/rest/impl/spacemembership/SpaceMembershipRest.java
```diff
@@ -416,6 +410,46 @@ public Response deleteSpaceMembershipById(
...
+			String invitationToken = spaceService.generateInvitationToken(Long.parseLong(spaceId),
+					RestUtils.getCurrentUserIdentityId());
+
+			DataEntity response = new DataEntity();
+			response.put("invitationToken", invitationToken);
+
+			return Response.ok(response).build();
+		} catch (Exception e) {
+			return Response.status(Response.Status.INTERNAL_SERVER_ERROR).build();
+		}
```
boubaker, line 449: Please remove the catch of the generic exception in REST endpoint to make it clear and logged what's happening, in logs.

```suggestion
```

component/service/src/main/java/org/exoplatform/social/rest/impl/spacemembership/SpaceMembershipRest.java
```diff
@@ -416,6 +410,46 @@ public Response deleteSpaceMembershipById(
...
+
+		try {
+			String invitationToken = spaceService.generateInvitationToken(Long.parseLong(spaceId),
+					RestUtils.getCurrentUserIdentityId());
+
+			DataEntity response = new DataEntity();
+			response.put("invitationToken", invitationToken);
+
+			return Response.ok(response).build();
```
boubaker, line 446:
```suggestion
			return Response.ok(invitationUrl).build();
```

component/service/src/main/java/org/exoplatform/social/rest/impl/spacemembership/SpaceMembershipRest.java
```diff
@@ -416,6 +410,46 @@ public Response deleteSpaceMembershipById(
...
+			return Response.status(Response.Status.NOT_FOUND).build();
+		}
+
+		if (!spaceService.canManageSpace(space, authenticatedUser)) {
+			return Response.status(Response.Status.UNAUTHORIZED).build();
+		}
+
+		try {
+			String invitationToken = spaceService.generateInvitationToken(Long.parseLong(spaceId),
+					RestUtils.getCurrentUserIdentityId());
```
boubaker, line 441: Please add a method for Full URL generation (including the domain name, using `CommonsUtils.getCurrentDomain()`) rather than just generating the Token. This will help to evolve the Implementation when needed without impacting the API.

```suggestion
			String invitationUrl = spaceService.generateInvitationUrl(spaceId, RestUtils.getCurrentUserIdentityId());
```

component/service/src/main/java/org/exoplatform/social/rest/impl/spacemembership/SpaceMembershipRest.java
```diff
@@ -416,6 +410,46 @@ public Response deleteSpaceMembershipById(
...
+	@GET
+	@Path("/invitationToken")
+	@RolesAllowed("users")
+	@Produces(MediaType.APPLICATION_JSON)
+	@Operation(summary = "Generate invitation token for a space", method = "GET", description = "Generates a secure invitation token for a space.")
+	@ApiResponses(value = { @ApiResponse(responseCode = "200", description = "Invitation token generated"),
+			@ApiResponse(responseCode = "401", description = "User not authorized"),
+			@ApiResponse(responseCode = "404", description = "Space not found") })
+	public Response generateInvitationToken(@Parameter(description = "Space technical identifier", required = true)
+                                          @QueryParam("spaceId") String spaceId) {
```
boubaker, line 422: You can use a `Long` instead of a `String` for a known typed parameter.

```suggestion
                                          @QueryParam("spaceId") Long spaceId) {
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
hakermi: OK, but IMO the min length should be 8 wdyt ?
random.nextInt(8, invitationTokenNonceMaxLength);

boubaker: Good idea! In this case, we need to make sure that upper bound is equal or greater than 12 by example IMO

hakermi:
feedback applied here: https://github.com/Meeds-io/social/pull/5599

boubaker:
Thanks :pray:

PR #5599, opened 2026-02-18
feat: Enhance Space invitation link generation API - MEED-10221 - Meeds-io/MIPs#244

hakermi:
Enhance Space invitation link generation API

### DIFF
--- a/component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
+++ b/component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
@@ -51,12 +51,14 @@ import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 
+import org.exoplatform.commons.exception.ObjectNotFoundException;
 import org.exoplatform.commons.file.model.FileItem;
 import org.exoplatform.commons.file.services.FileService;
 import org.exoplatform.commons.utils.CommonsUtils;
 import org.exoplatform.commons.utils.ListAccess;
 import org.exoplatform.commons.utils.ListAccessImpl;
 import org.exoplatform.container.ExoContainerContext;
+import org.exoplatform.container.xml.InitParams;
 import org.exoplatform.portal.config.UserACL;
 import org.exoplatform.services.log.ExoLogger;
 import org.exoplatform.services.log.Log;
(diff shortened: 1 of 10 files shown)


component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -133,10 +136,14 @@ public class SpaceServiceImpl implements SpaceService {
...
 
-  private SpaceLifecycle              spaceLifeCycle         = new SpaceLifecycle();
+  private SpaceLifecycle              spaceLifeCycle          = new SpaceLifecycle();
 
-  private static CodecInitializer     codecInitializer;
+  private CodecInitializer            codecInitializer;
+  
+  private SecureRandomService         secureRandomService;
 
+  @Value("${space.invitation.token.nonce.maxLength:12}")
```
boubaker: In fact, the Service is still injected using `Kernel` rather than `Spring` (XML)

Image

We will need to add an `<init-params>` in XML definition:

Image

and then consume it in Constructor:

Image

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -1131,16 +1138,40 @@ public void addSpaceListener(SpaceListenerPlugin plugin) {
...
-    String payload = spaceId + ":" + userIdentityId + ":" + nonce;
-    return getCodec().encode(payload);
+  public String generateInvitationLink(Long spaceId, Long userIdentityId) throws Exception {
+    Space space = getSpaceById(spaceId);
+    if (space == null) {
+        throw new ObjectNotFoundException("space not found");
+    }
+
+    if (!canManageSpace(space, identityManager.getIdentity(userIdentityId).getRemoteId())) {
+    throw new IllegalAccessException();
```
boubaker:
```suggestion
      throw new IllegalAccessException();
```

component/service/src/main/java/org/exoplatform/social/rest/impl/spacemembership/SpaceMembershipRest.java
```diff
@@ -410,45 +415,33 @@ public Response deleteSpaceMembershipById(
...
-	@Path("/invitationToken")
-	@RolesAllowed("users")
-	@Produces(MediaType.APPLICATION_JSON)
-	@Operation(summary = "Generate invitation token for a space", method = "GET", description = "Generates a secure invitation token for a space.")
-	@ApiResponses(value = { @ApiResponse(responseCode = "200", description = "Invitation token generated"),
+  @GET
+  @Path("/invitationToken")
+  @RolesAllowed("users")
+  @Produces(MediaType.APPLICATION_JSON)
+  @Operation(summary = "Generate invitation token for a space", method = "GET", description = "Generates a secure invitation token for a space.")
```
boubaker: Please update the Documentation to indicate that the Invitation URL is generated rather than the Token

component/service/src/main/java/org/exoplatform/social/rest/impl/spacemembership/SpaceMembershipRest.java
```diff
@@ -410,45 +415,33 @@ public Response deleteSpaceMembershipById(
...
-			return Response.status(Response.Status.BAD_REQUEST).entity("spaceId is required").build();
-		}
-
-		String authenticatedUser = ConversationState.getCurrent().getIdentity().getUserId();
-
-		Space space = spaceService.getSpaceById(spaceId);
-		if (space == null) {
-			return Response.status(Response.Status.NOT_FOUND).build();
-		}
+  public Response generateInvitationToken(@Parameter(description = "Space technical identifier", required = true)
```
boubaker:
```suggestion
  public Response generateInvitationUrl(@Parameter(description = "Space technical identifier", required = true)
```

component/core/src/main/java/io/meeds/social/space/service/SpaceServiceImpl.java
```diff
@@ -156,6 +165,12 @@ public SpaceServiceImpl(SpaceStorage spaceStorage, // NOSONAR
     this.localeConfigService = localeConfigService;
     this.fileService = fileService;
     this.cookieTokenService = cookieTokenService;
+    String value = initParams.getValueParam(NONCE_MAX_LENGTH_VALUE_PARAM).getValue();
```
boubaker: In fact, the `initParams` is nullable (in Unit Tests by example)

hakermi: Its only in unit tests and its mocked in that case, but no problem i will add a null check

boubaker (approved):

PR #5621, opened 2026-02-23
fix: CSV user import issue if attribute isn't the exact value - MEED-10196 - Meeds-io/meeds#4002

akhanfir:
Before this change, create attribute filed which ID is Phone and import csv having `téléphone` attribute, user attribute never displayed on user profile and the `Téléphone` attribute can never be reimported again. To fix this problem, ensured that profile property validation is performed in a strictly case-sensitive way at the application level, independently of the database collation. Now, the system verifies that the imported field name exactly matches the attribute ID before allowing it. After this change, attribute will not be imported if it doesn't exactly match the attribute ID.

### DIFF
--- a/component/core/src/main/java/io/meeds/social/core/identity/service/UserImportService.java
+++ b/component/core/src/main/java/io/meeds/social/core/identity/service/UserImportService.java
@@ -553,6 +553,9 @@ public class UserImportService {
       if (!STANDARD_FIELDS.contains(field)) {
         if (!field.contains(".")) {
           ProfilePropertySetting propertySetting = profilePropertyService.getProfileSettingByName(field);
+          if (propertySetting != null && !propertySetting.getPropertyName().equals(field)) {
+            propertySetting = null;
+          }
           if (propertySetting == null) {
             userImportResult.addWarnMessage("ALL", "PROFILE_PROPERTY_DOES_NOT_EXIST:" + field);
             unauthorizedFields.add(field);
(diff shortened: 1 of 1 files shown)


mkrout (dismissed):

component/core/src/main/java/io/meeds/social/core/identity/service/UserImportService.java
```diff
@@ -547,39 +548,66 @@ protected void linkMembership(UserImpl user, Group groupObject, MembershipType m
...
+
+          } else if (propertySetting.isMultiValued()
+            && !SYSTEM_PARENT_MULTIVALUED_FIELDS.contains(field)) {
+            userImportResult.addWarnMessage("ALL",
+              "CUSTOM_FIELD_MULTIVALUED:" + field);
             unauthorizedFields.add(field);
           }
+
         } else {
+          // 🔹 gérer les parent.child pareillement (on peut aussi sécuriser ici si besoin)
```
ahamdi: If needed , we should add comments in English

component/core/src/main/java/io/meeds/social/core/identity/service/UserImportService.java
```diff
@@ -547,39 +548,66 @@ protected void linkMembership(UserImpl user, Group groupObject, MembershipType m
...
-          ProfilePropertySetting parentProperty = profilePropertyService.getProfileSettingByName(fieldNames[0]);
           if (fieldNames.length > 2) {
-            userImportResult.addWarnMessage("ALL", "PROPERTY_HAS_MORE_THAN_ONE_PARENT:" + field);
-            unauthorizedFields.add(field);
-          } else if (parentProperty == null) {
-            userImportResult.addWarnMessage("ALL", "PROPERTY_HAS_MISSING_PARENT_PROPERTY:" + field);
-            unauthorizedFields.add(field);
-          } else if (parentProperty.isMultiValued()
-                     && !SYSTEM_PARENT_MULTIVALUED_FIELDS.contains(parentProperty.getPropertyName())) {
-            userImportResult.addWarnMessage("ALL", "CUSTOM_PARENT_FIELD:" + field);
```
ahamdi, line 577: Why is this block removed ?

component/core/src/main/java/io/meeds/social/core/identity/service/UserImportService.java
ahamdi: Comments should be in English

ahamdi (approved):

PR #5659, opened 2026-03-26
feat: Display the number of views of a post - MEED-10147 - Meeds-io/MIPs#239

MayTekayaa:
This change will display the number of views of a post via extension registry.

### DIFF
--- a/component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
+++ b/component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
@@ -228,6 +228,8 @@ public class EntityBuilder {
 
   public static final String              TOTAL_COMMENTS_COUNT                       = "totalCommentsCount";
 
+  public static final String              VIEWERS_COUNT                              = "viewersCount";
+
   private static final int                DEFAULT_LIKERS_LIMIT                       = 4;
 
   private static UserPortalConfigService  userPortalConfigService;
@@ -1298,6 +1300,9 @@ public class EntityBuilder {
     if (expandFields.contains(TOTAL_COMMENTS_COUNT)) {
       activityEntity.setTotalCommentsCount(getActivityManager().getNumberOfAllComments(activity.getId()));
     }
+    if (expandFields.contains(VIEWERS_COUNT)) {
+      activityEntity.setViewersCount(getActivityViewersCount(activity));
+    }
     return activityEntity;
   }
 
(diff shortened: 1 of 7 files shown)


component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
```diff
@@ -1777,6 +1782,36 @@ private static int getEtagValue(String... properties) {
...
+                                                                                                                                                            metadataItem.getObjectId(),
+                                                                                                                                                            metadataItem.getParentObjectId(),
+                                                                                                                                                            metadataItem.getCreatorId(),
+                                                                                                                                                            metadataItem.getMetadata().getAudienceId(),
+                                                                                                                                                            metadataItem.getProperties()))
+                                                                                                                .toList()
```
boubaker: No need to make this Objects switch, isn't it ? Try an algorithm with less memory consumption please.

webapp/src/main/webapp/vue-apps/activity-stream/js/ActivityConstants.js
```diff
@@ -1,9 +1,9 @@
...
   // Attention!!! when changing this, the list of preloaded
   // URLs has to change in JSP as well
-  FULL_ACTIVITY_IDS_EXPAND: 'ids,identity,likes,shared,commentsPreview,subComments,favorite',
-  FULL_ACTIVITY_EXPAND: 'identity,likes,shared,commentsPreview,subComments,favorite,totalCommentsCount',
-  ACTIVITY_EXPAND: 'identity,likes,shared',
-  FULL_COMMENT_EXPAND: 'identity,likes,subComments,totalCommentsCount',
+  FULL_ACTIVITY_IDS_EXPAND: 'ids,identity,likes,shared,commentsPreview,subComments,favorite,viewersCount',
+  FULL_ACTIVITY_EXPAND: 'identity,likes,shared,commentsPreview,subComments,favorite,totalCommentsCount,viewersCount',
+  ACTIVITY_EXPAND: 'identity,likes,shared,viewersCount',
+  FULL_COMMENT_EXPAND: 'identity,likes,subComments,totalCommentsCount,viewersCount',
```
boubaker, line 7: please fix at the same time the preload directive when displaying activity stream (avoid having this warning in console when displaying Activity Stream):

Image

( Check org.exoplatform.social.rest.impl.activity.ActivityRest.addPreloadActivityIds(List<String>, int, String, ResponseBuilder) for further information )

component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
```diff
@@ -1777,6 +1782,21 @@ private static int getEtagValue(String... properties) {
...
+  private static int getActivityViewersCount(ExoSocialActivity activity) {
+    Map<String, List<MetadataItem>> activityMetadatas = activity.getMetadatas();
+    if (MapUtils.isNotEmpty(activityMetadatas) && activityMetadatas.containsKey("viewers")) {
+      List<MetadataItem> viewersMetadata = activityMetadatas.get("viewers");
+      if (viewersMetadata != null && !viewersMetadata.isEmpty()) {
+        MetadataItem viewers = viewersMetadata.get(0);
+        Map<String, String> props = viewers.getProperties();
+        String viewerIdsString = props.get("viewerIds");
+        String[] ids = viewerIdsString.split(",");
+        return (viewerIdsString.isEmpty()) ? 0 : ids.length;
```
boubaker:
```suggestion
        return ids.length;
```

boubaker (approved):

PR #5666, opened 2026-03-30
fix: when login with openid, with a user is deactivated in eXo, the error message is not correct - MEED-10341 - meeds-io/meeds#4151

rdenarie:
Before this fix, if the user is deactivated in eXo, the page displayed is the page to link the account. It should display the login form with the classic error message when the user is deactivated

Resolves meeds-io/meeds#4151


azayati:
@rdenarie can you rebase ?

rdenarie:
> @rdenarie can you rebase ?

Done

azayati (approved):

PR #5678, opened 2026-04-08
fix: Fix the number of views label when there is no analytics portlet - MEED-10365 - Meeds-io/meeds#4168

MayTekayaa:
This change will fix the number of views label when there is no analytics portlet.

### DIFF
--- a/component/service/src/main/resources/locale/social/Webui_en.properties
+++ b/component/service/src/main/resources/locale/social/Webui_en.properties
@@ -124,7 +124,6 @@ UIActivity.label.Show_All_Comments=View all {0} comments.
 UIActivity.label.Hide_All_Comments=Hide all {0} comments.
 UIActivity.label.Show_Previous_Comments=View previous comments.
 UIActivity.label.seeMore=see more
-analytics.activity.views={0} Views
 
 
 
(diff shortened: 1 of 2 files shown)


component/service/src/main/resources/locale/social/Webui_en.properties
```diff
@@ -124,6 +124,7 @@ UIActivity.label.Show_All_Comments=View all {0} comments.
 UIActivity.label.Hide_All_Comments=Hide all {0} comments.
 UIActivity.label.Show_Previous_Comments=View previous comments.
 UIActivity.label.seeMore=see more
+analytics.activity.views={0} Views
```
boubaker: Please add it into `Portlets_en.properties` which is the generic I18N file. In fact, rather than adding it here, please move it from the file `analytics-webapps/src/main/resources/locale/portlet/Analytics_en.properties` into a new file `analytics-webapps/src/main/resources/locale/portlet/Portlets_en.properties` (remove all definitions added into the file `Analytics_***.properties`).

When adding a new I18N file, we will have to update the `crowdin.yml` of analytics project in order to make it visible in Crowdin Meeds Project.

PR #5709, opened 2026-05-07
fix: Fix confirmation popup message XSS vulnerability - EXO-86231 

azayati:
Prior to this change, a JavaScript payload injected into the confirmation popup message could be executed when the popup was displayed, leading to an XSS vulnerability.
After this commit, the confirmation popup message is sanitized using the v-sanitize-html directive instead of v-html, preventing the XSS vulnerability.

Resolves https://github.com/Meeds-io/si/issues/11

### DIFF
--- a/webapp/src/main/webapp/vue-apps/common/components/ConfirmDialog.vue
+++ b/webapp/src/main/webapp/vue-apps/common/components/ConfirmDialog.vue
@@ -16,7 +16,7 @@
         <span class="ignore-vuetify-classes text-title" v-html="title"></span>
       </div>
       <!-- eslint-disable-next-line vue/no-v-html -->
-      <v-card-text v-html="message" />
+      <v-card-text v-sanitized-html="message" />
       <v-card-actions v-if="!hideActions">
         <v-spacer />
         <button


rdenarie (approved):

webapp/src/main/webapp/vue-apps/common/components/ConfirmDialog.vue
```diff
@@ -16,7 +16,7 @@
         <span class="ignore-vuetify-classes text-title" v-html="title"></span>
       </div>
       <!-- eslint-disable-next-line vue/no-v-html -->
```
boubaker, line 18: Delete this if not needed anymore, but IMO it was added on purpose to not be sanitized here but sanitized when the html content is provided. To check where it's used and whether there is an impact on existing features using it or not.

```suggestion
```

azayati: @boubaker I have verified, this has no effect when adding or removing it, seems that this ESLint rule is not activated at all. So I have [removed it](https://github.com/Meeds-io/social/commit/6011aae17e76040804a293a497d34bac33533882#diff-182b7415fc810396c2063b94cd44d7af8f25c133f254647ea7b190b67f4bb3f4L18-L19) since anyway sanitization is recommended to be done only on display.

boubaker: The deletion comment was about this
image
In fact, using `<!-- eslint-disable-next-line vue/no-v-html -->` will disable eslint check. If the `v-html` is replaced by `v-sanitized-html`, then the comment isn't used anymore.

azayati: @boubaker already deleted directly in ft/maintenance see https://github.com/Meeds-io/social/commit/6011aae17e76040804a293a497d34bac33533882#diff-182b7415fc810396c2063b94cd44d7af8f25c133f254647ea7b190b67f4bb3f4L18-L19

boubaker: cool

PR #5781, opened 2026-06-08
fix: Tooltip displayed on feed after scroll - EXO-87368

sofyenne:
Prior to this change, in mobile view, long-pressing the feed action buttons would trigger the tooltip, which remained visible even while scrolling the screen. , this change disables the tooltip in mobile view.

### DIFF
--- a/webapp/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityStreamToolbar.vue
+++ b/webapp/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityStreamToolbar.vue
@@ -69,8 +69,8 @@
             </v-tooltip>
             <v-tooltip
               v-if="$root.canEdit"
-              bottom
-              :disabled="isMobile">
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/vue-apps/activity-stream/components/toolbar/ActivityStreamToolbar.vue
```diff
@@ -67,7 +67,10 @@
                 {{ $t('activity.filter.button.markAllAsRead') }}
               </span>
             </v-tooltip>
-            <v-tooltip v-if="$root.canEdit" bottom>
+            <v-tooltip
+              v-if="$root.canEdit"
+              bottom
+              :disabled="isMobile">
```
boubaker: variabilized pros before constant

```suggestion
              :disabled="isMobile"
              bottom>
```

boubaker (approved):

PR #5782, opened 2026-06-08
fix: Display space description as plain text in the muted spaces list  - EXO-97380

sofyenne:
Prior to this change, when a space description contained HTML, it was displayed as HTML in the muted spaces list drawer. This change ensures that the space description is displayed as plain text.

### DIFF
--- a/webapp/src/main/webapp/vue-apps/notification-user-settings/components/UserSettingNotificationMuteSpacesDrawer.vue
+++ b/webapp/src/main/webapp/vue-apps/notification-user-settings/components/UserSettingNotificationMuteSpacesDrawer.vue
@@ -75,7 +75,7 @@
               </v-list-item-title>
               <v-list-item-subtitle
                 v-if="space.description"
-                v-sanitized-html="space.description"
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/vue-apps/notification-user-settings/components/UserSettingNotificationMuteSpacesDrawer.vue
```diff
@@ -75,7 +75,7 @@
               </v-list-item-title>
               <v-list-item-subtitle
                 v-if="space.description"
-                v-sanitized-html="space.description"
+                v-sanitized-html="getSpaceDescriptionText(space.description)"
                 class="caption text-truncate" />
```
boubaker: If it becomes text, it would be better to display it as a text field rather than HTML.

```suggestion
                class="caption text-truncate">
                {{ getSpaceDescriptionText(space.description) }}
</v-list-item-subtitle>
```

boubaker (approved):

PR #5789, opened 2026-06-09
fix: Fix the count of the manager's subordinates - EXO-87491 

MayTekayaa:
This change will fix the count of the manager's subordinates by setting wildCardSearch to false to filter by exact value match.

### DIFF
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
@@ -87,7 +87,6 @@ public class ProfileSearchConnector {
   public int count(Identity identity,
                    ProfileFilter filter,
                    Type type) {
-    filter.setWildcardSearch(false);
     String esQuery = buildQueryStatement(identity, filter, type, 0, 1);
     String jsonResponse = this.client.sendRequest(esQuery, this.index);
     return getCount(jsonResponse);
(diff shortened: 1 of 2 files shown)


component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java
```diff
@@ -87,6 +87,7 @@ public List<String> search(Identity identity,
   public int count(Identity identity,
                    ProfileFilter filter,
                    Type type) {
+    filter.setWildcardSearch(false);
```
boubaker: Do you mean the default value should be false [here](https://github.com/Meeds-io/social/blob/ea05b0e412014fdf827fa896c32376e8b9a4934d/component/api/src/main/java/org/exoplatform/social/core/profile/ProfileFilter.java#L86) ? This may cause regressions in search and user suggester especially when we search a part a word which is in the middel, exemple: "ames" to search for "James"

MayTekayaa: No, we should keep it to “true” by default,
and then set it to “false” so that it is only used for the “count managers” function.

boubaker: So for you, the global `count` method shouldn't give the same result as `search` method and it's used for `count managers` only ? Can you double check its usage please.

MayTekayaa: You are right, thanks for you feedback
moved when building ManagedUsersCount

boubaker: Thanks

boubaker (approved):

PR #5814, opened 2026-06-18
fix: sanitize rich editor content on initialization to prevent XSS attacks - EXO-87658

sofyenne:
Prior to this change, the content passed to the editor was loaded directly without sanitization during initialization. This could allow XSS attacks. This change ensures that editor content is sanitized during initialization.


boubaker (commented):
Thanks @sofyenne 

Please update the PR title and description to reflect the modification rather than Task description.

boubaker (approved):

PR #5819, opened 2026-06-19
fix: error when requesting access to a hidden space via a referral link - EXO-87778

sofyenne:
Prior to this change, when attempting to access an unlisted space via a referral link, an Unauthorized exception was thrown when submitting an access request (registration by validation). This change adds a check to allow valid referral/invitation access to hidden spaces, resolving the issue.


component/service/src/main/java/org/exoplatform/social/rest/impl/spacemembership/SpaceMembershipRest.java
```diff
@@ -493,4 +493,10 @@ private void handleSpaceInvitation(String invitationToken, String user) {
       }
     }
   }
+
+  private boolean isValidInvitationToHiddenSpace(Space space, String invitationToken) {
+    return Space.HIDDEN.equals(space.getVisibility()) 
+        && !Space.CLOSED.equals(space.getRegistration()) 
+        && StringUtils.isNotBlank(invitationToken);
```
boubaker: Need to check either this is sufficient ti give access or not because the invitation token isn't meant to change permission to access hidden spaces IMO.

boubaker (approved):

PR #5820, opened 2026-06-19
fix: Disallow editing a post comment without any changes - EXO-87782

azayati:
Prior to this change, when editing a post comment containing a mention of someone or everyone then save directly without any changes, the saved comment has display problems for the mentions. After this commit, in order to avoid such mentions display problem, we will disallow to edit a post comment without any changes.

### DIFF
--- a/webapp/src/main/webapp/vue-apps/activity-stream/components/comment/content/ActivityCommentRichText.vue
+++ b/webapp/src/main/webapp/vue-apps/activity-stream/components/comment/content/ActivityCommentRichText.vue
@@ -99,7 +99,8 @@ export default {
     avatarSize: '33px',
     files: [],
     attachments: null,
-    comment: null
+    comment: null,
+    commentBodyEdited: false,
   }),
   computed: {
     avatarUrl() {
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/vue-apps/activity-stream/components/comment/content/ActivityCommentRichText.vue
```diff
@@ -152,11 +153,18 @@ export default {
   },
   mounted() {
     this.init();
-  },
+  },
```
boubaker, line 156:
```suggestion
  },
```

boubaker (approved):
Thanks, just a useless space added please

webapp/src/main/webapp/vue-apps/activity-stream/components/comment/content/ActivityCommentRichText.vue
azayati: @boubaker fixed directly in the feature/maintenance [commit ](https://github.com/Meeds-io/social/commit/3c159bc683617b238c2a7496af241fb5c0ffec76)

boubaker: cool, Thanks.

PR #5836, opened 2026-06-29
fix: When a text content is domPurify, mailto is incorrectly displayed as 2 links - EXO-88030

rdenarie:
Before this fix, the domPurify does not generates a correct link for mailto link (or any protocol different from mailto with email after) It generates currently 2 links : one before the @, one after, instead of a single link for all the protocol

The commit ensure to manage theses links in DomPurify

### DIFF
--- a/webapp/src/main/webapp/js/ExtendedDomPurify.js
+++ b/webapp/src/main/webapp/js/ExtendedDomPurify.js
@@ -34,6 +34,7 @@
     content = decodeEntitiesInUrls(content);
     content = content.replace(/<div> <\/div>/g, '<div><br><\/div>');
     content = content.trim().replace(/>[ \n]+</g, '> <').replace(/  /g, '&nbsp;&nbsp;');
+    content = content.replace(/\b[a-zA-Z][a-zA-Z0-9+\-.]*:\/?\/?[^\s<>"'@&]+(?:@|&#64;)[^\s<>"'<>]+/gi, url => `<a href="${url}">${url}</a>`);
     const pureHtml = DOMPurify.sanitize(Autolinker.link(content, {
       email: false,
       replaceFn : function (match) {
--- a/webapp/src/main/webapp/js/ExtendedDomPurify.js
+++ b/webapp/src/main/webapp/js/ExtendedDomPurify.js
@@ -34,6 +34,7 @@
     content = decodeEntitiesInUrls(content);
     content = content.replace(/<div> <\/div>/g, '<div><br><\/div>');
     content = content.trim().replace(/>[ \n]+</g, '> <').replace(/  /g, '&nbsp;&nbsp;');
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/js/ExtendedDomPurify.js
```diff
@@ -34,6 +34,7 @@
     content = decodeEntitiesInUrls(content);
     content = content.replace(/<div> <\/div>/g, '<div><br><\/div>');
     content = content.trim().replace(/>[ \n]+</g, '> <').replace(/  /g, '&nbsp;&nbsp;');
+    content = content.replace(/\b[a-zA-Z][a-zA-Z0-9+\-.]*:\/?\/?[^\s<>"'@&]+(?:@|&#64;)[^\s<>"'<>]+/gi, url => `<a href="${url}">${url}</a>`);
```
boubaker, line 38: Thanks. Have you checked if the upgrade from `2.3.8` to `3.4.11` can fix the prb ?

rdenarie: In fact, there are 2 steps : 
The DomPurify library ensure that there is no problematic attributes in the DOM, and remove it if necessary.

The current problem is text transformation in links.
The Autolinker lib transform this `https://google.fr` into `<a href='https://google.fr'>https://google.fr</a>`

For text like `mailto://rdenarie#64;exoplatform.com`, it is transformed in `<a href='mailto://rdenarie'>mailto://rdenarie</a>@<a href='exoplatform.com'>exoplatform.com</a>`
instead of 
`<a href='mailto://rdenarie@exoplatform.com'>mailto://rdenarie@exoplatform.com</a>` 

So the problem is not in DomPurify but in Autolinker, and there is an [opened issue](github.com/gregjacobs/Autolinker.js/issues/393)

So waiting the fix in autolinker, I propose this fix.

boubaker: cool, thanks for the clarification, may be adding the link of opened issue in a comment with this:

```suggestion
    // FIXME a workaround for https://github.com/Meeds-io/social/pull/github.com/gregjacobs/Autolinker.js/issues/393
    content = content.replace(/\b[a-zA-Z][a-zA-Z0-9+\-.]*:\/?\/?[^\s<>"'@&]+(?:@|&#64;)[^\s<>"'<>]+/gi, url => `<a href="${url}">${url}</a>`);
```

WDYT ?

rdenarie: done

boubaker (dismissed):

rdenarie:
@boubaker can you revalidate ? I rebased the branch

boubaker (approved):

PR #5837, opened 2026-06-29
fix: disable link option in crop drawer when opened from activity composer - EXO-88016

hakermi:
disable link option in crop drawer when opened from activity composer

### DIFF
--- a/webapp/src/main/webapp/vue-apps/attach-image/components/form/ImageAttachmentCropDrawer.vue
+++ b/webapp/src/main/webapp/vue-apps/attach-image/components/form/ImageAttachmentCropDrawer.vue
@@ -27,8 +27,8 @@
     :back-icon="backIcon"
     :use-format="useFormat"
     :custom-format="!useFormat"
+    :link="!noLink"
     alt
-    link
     @apply="forwardUpdate" />
 </template>
 <script>
@@ -61,6 +61,7 @@ export default {
   data: () => ({
     imageItem: null,
     maxFileSize: 20971520,
+    noLink: false,
   }),
   computed: {
     imageCropperSrc() {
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/vue-apps/attach-image/components/form/ImageAttachmentCropDrawer.vue
```diff
@@ -79,10 +80,14 @@ export default {
...
-      this.open(event?.detail);
+      const detail = event?.detail || {};
+      const imageItem = detail.imageItem || detail;
+      const allowLink = detail.allowLink;
+      this.open(imageItem, allowLink);
     },
-    open(imageItem) {
+    open(imageItem, allowLink) {
       this.imageItem = imageItem;
+      this.allowLink = allowLink !== false;
```
boubaker: It's already falsy and truthy, but if you would like to change it to be a boolean, you can simply add `!!`, WDYT ?

```suggestion
      this.allowLink = !!allowLink;
```

hakermi: I want the undefined to be considered as true, because its a reusable component, and used in imagePorlet without passing that param, so will keep work as expected, thus i used it that way

boubaker: ok, thus may be change the param to `noLink` rather than `allowLink`. So that if it's not present, it's automatically considered as "with a link"

hakermi: Done

boubaker (dismissed):

boubaker (approved):

PR #5889, opened 2026-07-17
Merge exip7.3.0.1

AzmiTouil:

### DIFF
--- a/component/api/src/main/java/io/meeds/social/category/model/CategoryEntryItem.java
+++ b/component/api/src/main/java/io/meeds/social/category/model/CategoryEntryItem.java
@@ -46,8 +46,8 @@ public class CategoryEntryItem {
   private String     objectType;
 
   /**
-   * Fontawesome Icon identifier used switch a corner badge to identify the
-   * entry type. Supplied by the owning plugin, never guessed switch the
+   * Fontawesome Icon identifier used as a corner badge to identify the
+   * entry type. Supplied by the owning plugin, never guessed from the
    * objectType value.
    */
   private String     icon;
--- a/component/api/src/main/java/io/meeds/social/category/model/CategoryEntryItem.java
+++ b/component/api/src/main/java/io/meeds/social/category/model/CategoryEntryItem.java
@@ -1,0 +1,81 @@
+/**
+ * This file is part of the Meeds project (https://meeds.io/).
+ *
+ * Copyright (C) 2020 - 2026 Meeds Association contact@meeds.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package io.meeds.social.category.model;
+
+import java.util.Date;
+import java.util.List;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * A generic, object-type agnostic preview of an entry linked to a Category,
+ * supplied by the {@link io.meeds.social.category.plugin.CategoryPlugin}
+ * managing the designated objectType.
+ */
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class CategoryEntryItem {
+
+  /**
+   * Object technical identifier
+   */
+  private String     id;
+
+  /**
+   * Object type
+   */
+  private String     objectType;
+
+  /**
+   * Fontawesome Icon identifier used switch a corner badge to identify the
+   * entry type. Supplied by the owning plugin, never guessed switch the
+   * objectType value.
+   */
+  private String     icon;
+
+  private String     title;
+
+  private String     summary;
+
+  private String     illustrationUrl;
+
+  private String     url;
+
+  private String     authorDisplayName;
+
+  private String     authorAvatarUrl;
+
+  private String     spaceDisplayName;
+
+  private String     spaceAvatarUrl;
+
+  private Date       date;
+
+  private long       likesCount;
+
+  private long       commentsCount;
+
+  private long       viewsCount;
+
+  private List<Long> categoryIds;
+
+}
--- a/component/api/src/main/java/io/meeds/social/category/model/CategoryEntryList.java
+++ b/component/api/src/main/java/io/meeds/social/category/model/CategoryEntryList.java
@@ -1,0 +1,40 @@
+/**
+ * This file is part of the Meeds project (https://meeds.io/).
+ *
+ * Copyright (C) 2020 - 2026 Meeds Association contact@meeds.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+package io.meeds.social.category.model;
+
+import java.util.List;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class CategoryEntryList {
+
+  private List<CategoryEntryItem> items;
+
+  private long                      offset;
+
+  private long                      limit;
+
+  private boolean                   hasMore;
+
+}
--- a/component/api/src/main/java/io/meeds/social/category/model/CategoryEntryList.java
+++ b/component/api/src/main/java/io/meeds/social/category/model/CategoryEntryList.java
@@ -37,4 +37,16 @@ public class CategoryEntryList {
 
   private boolean                   hasMore;
 
+  /**
+   * Raw storage offset to use for the next page request (i.e. "load more"),
+   * as opposed to {@link #offset}, which is the storage offset used to build
+   * this page. Since the number of items returned to the caller can differ
+   * from the number of raw storage rows consumed to build them (some rows are
+   * dropped by de-duplication and by the accessibility check), the caller
+   * must not simply advance {@link #offset} by {@link #limit} when loading
+   * the next page - doing so misaligns with the underlying storage and can
+   * skip or repeat entries.
+   */
+  private long                      nextOffset;
+
 }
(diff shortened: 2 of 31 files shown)


boubaker (commented):
## Review: Content Categorization (eXIP — Tribe activity 473782 / news 46633)

Cross-checked this PR against the two companion PRs (`content#832`, `notes#1728`) and the eXIP spec. Two issues here:

### 1. `GET /categories/{id}/entries` will miss most real News/Notes content
`CategoryServiceImpl.getCategoryEntries()` → `CategoryStorage.getLinkedItems()` filters strictly on `mi.objectType IN (:objectTypes)`, and every frontend call site (`ExoNewsDetailsToolBar.vue`, `NotePage.vue`) hardcodes `types=news,notes` — never `activity`.

The problem: once a News/Note is posted to the activity stream, its category link is persisted as `CategoryObject("activity", activityId, ...)` (see `NewsService.linkActivityCategories()` in content, and `NotePage.vue`'s `updateCategories()` in notes, which both fall back to the activity-typed link once `activityId` exists). Since that's the common case for published content, the new "browse content by category" drawer will come back empty/incomplete for exactly the items whose own category chip the user just clicked — contradicting the "Filter to categories" user story in the spec, which expects a News/Notes-level link to be the source of truth with the Activity link merely kept in sync (via a listener that doesn't exist in any of the three PRs).

### 2. `getCategoryEntries` pagination is unsound across "Load more" pages
```java
long fetchLimit = limit * 3;
List<CategoryObject> linkedObjects = categoryStorage.getLinkedItems(categoryId, objectTypes, (int) offset, (int) fetchLimit);
```
`offset` here is the raw storage-row skip, but `EntryListDrawer.vue`'s `loadMore()` advances `offset += limit`, i.e. "items already returned to the user" — not "raw rows consumed". Because dedup + `canAccess` filtering + null-preview drops happen *after* this fetch, the number of raw rows consumed to produce a page of `limit` items varies and isn't tracked or returned. Any time filtering drops candidates (private space, duplicate cross-type link, no resolvable preview — all things the spec explicitly requires), the next page's storage offset will land in the wrong place, silently skipping or repeating entries.

---
Reviewed by Claude

AzmiTouil:
> ## Review: Content Categorization (eXIP — Tribe activity 473782 / news 46633)
> Cross-checked this PR against the two companion PRs (`content#832`, `notes#1728`) and the eXIP spec. Two issues here:
> 
> ### 1. `GET /categories/{id}/entries` will miss most real News/Notes content
> `CategoryServiceImpl.getCategoryEntries()` → `CategoryStorage.getLinkedItems()` filters strictly on `mi.objectType IN (:objectTypes)`, and every frontend call site (`ExoNewsDetailsToolBar.vue`, `NotePage.vue`) hardcodes `types=news,notes` — never `activity`.
> 
> The problem: once a News/Note is posted to the activity stream, its category link is persisted as `CategoryObject("activity", activityId, ...)` (see `NewsService.linkActivityCategories()` in content, and `NotePage.vue`'s `updateCategories()` in notes, which both fall back to the activity-typed link once `activityId` exists). Since that's the common case for published content, the new "browse content by category" drawer will come back empty/incomplete for exactly the items whose own category chip the user just clicked — contradicting the "Filter to categories" user story in the spec, which expects a News/Notes-level link to be the source of truth with the Activity link merely kept in sync (via a listener that doesn't exist in any of the three PRs).
> 
> ### 2. `getCategoryEntries` pagination is unsound across "Load more" pages
> ```java
> long fetchLimit = limit * 3;
> List<CategoryObject> linkedObjects = categoryStorage.getLinkedItems(categoryId, objectTypes, (int) offset, (int) fetchLimit);
> ```
> 
> `offset` here is the raw storage-row skip, but `EntryListDrawer.vue`'s `loadMore()` advances `offset += limit`, i.e. "items already returned to the user" — not "raw rows consumed". Because dedup + `canAccess` filtering + null-preview drops happen _after_ this fetch, the number of raw rows consumed to produce a page of `limit` items varies and isn't tracked or returned. Any time filtering drops candidates (private space, duplicate cross-type link, no resolvable preview — all things the spec explicitly requires), the next page's storage offset will land in the wrong place, silently skipping or repeating entries.
> 
> Reviewed by Claude


▎ Thanks for the review — both issues are now addressed:
▎
▎ 1. Activity-type filtering / empty results for published content: this was fixed by the last commit already in this PR (EXO-88497) — ActivityCategoryPlugin/SpaceCategoryPlugin weren't self-registering with CategoryPluginServiceImpl, so the "activity" → actual content-type transformation silently fell through to a no-op default. With that registered correctly, category links on posted content now resolve to news/notes as expected, matching the drawer's filter.
▎
▎ 2. Pagination offset bug: confirmed and fixed. getCategoryEntries() was advancing offset by the number of visible items returned, not by the number of raw storage rows actually consumed — since de-duplication and the accessibility check can drop a variable number of candidates, this could skip or repeat entries across "load more" pages. Added a nextOffset field to CategoryEntryList that precisely points to the raw row the next page must resume from, and updated EntryListDrawer.vue to use it instead of self-incrementing by limit. Added a dedicated test (testGetCategoryEntriesPaginationAcrossPagesDoesNotSkipOrRepeat) reproducing the exact skip/duplicate scenario across two consecutive pages.

boubaker (commented):
## Follow-up review

Verified the pagination fix — this is correct. `CategoryEntryList.nextOffset` now tracks the actual raw storage row consumed (via `lastConsumedIndex` in `buildPagedEntryList`), and `EntryListDrawer.vue`'s `loadEntries()` resumes from `data.nextOffset` instead of `offset + limit`. That fully addresses the skip/duplicate risk I flagged. Nice fix, and good that it's documented in both the Java field javadoc and the Vue comment.

The other issue from my earlier comment is still open: `getCategoryEntries()`/`CategoryStorage.getLinkedItems()` still filter strictly on the caller-supplied `objectTypes`, and there's still no path from an `"activity"`-typed link back to the News/Note it actually represents (`ActivityCategoryPlugin` wasn't given a `getEntryItem()` override, and no listener syncs a direct news/notes-level link). Since `content#832` and `notes` still resolve categories through the Activity link once an item is posted (the common case), the `/categories/{id}/entries?types=news,notes` endpoint will keep missing that content. Not something this PR alone can fix, but flagging since it's the core gap in the feature as a whole — I think it needs a decision (either `ActivityCategoryPlugin.getEntryItem()` delegates to the underlying content's plugin via its `metadataObjectType`/`metadataObjectId`, or the storage query resolves through the activity link) before this is feature-complete.

Not approving yet given that gap, but the pagination work here is solid.

---
Reviewed by Claude

boubaker (commented):
## Follow-up review (full fresh re-pass against current HEAD)

🔴 **Critical** — Browse-by-category still can't resolve posted News/Notes, and the fix is simpler than it first looked
[`CategoryServiceImpl.java:311-377`](https://github.com/Meeds-io/social/blob/d29992ad16e5d0e330ac45f455c74871fa79d93a/component/core/src/main/java/io/meeds/social/category/service/CategoryServiceImpl.java#L311-L377)
```java
private CategoryEntryList buildPagedEntryList(List<CategoryObject> linkedObjects, ...) {
  ...
  for (Map.Entry<String, CategoryObject> entry : deduplicated.entrySet()) {
    CategoryObject object = entry.getValue();               // raw object, never resolved
    if (... || !categoryPluginService.canAccess(object.getType(), object.getId(), username)) {
```
Once a News/Note is posted, its link is stored as `CategoryObject("activity", activityId, ...)`, and every call site queries `types=news,notes` only, so these rows are never even fetched. `ActivityCategoryPlugin` already has exactly the resolution hook needed — [`ActivityCategoryPlugin.java:87-91`](https://github.com/Meeds-io/social/blob/d29992ad16e5d0e330ac45f455c74871fa79d93a/component/core/src/main/java/io/meeds/social/activity/plugin/ActivityCategoryPlugin.java#L87-L91): `getObject()` resolves an activity's `CategoryObject` to its real `metadataObjectType`/`metadataObjectId` (verified: content's `NewsUtils.NEWS_METADATA_OBJECT_TYPE` = `"news"`, matching `NewsCategoryPlugin.OBJECT_TYPE` exactly). Nothing in `buildPagedEntryList` ever calls `categoryPluginService.getObject(object)` to normalize a raw linked object before deduping/filtering it.

**Fix:** include `"activity"` in the queried `objectTypes`, and call `categoryPluginService.getObject(object)` on each raw result in `buildPagedEntryList` before dedup/`canAccess`/`getEntryItem` — this resolves an activity-linked row back to its real News/Note automatically, using machinery that already exists.

---

🟡 **Medium** — Cross-type duplicate entries can reappear on a later "Load more" page
[`CategoryServiceImpl.java:337-361`](https://github.com/Meeds-io/social/blob/d29992ad16e5d0e330ac45f455c74871fa79d93a/component/core/src/main/java/io/meeds/social/category/service/CategoryServiceImpl.java#L337-L361)
```java
if (existing == null) {
  deduplicated.put(object.getId(), object);
  firstIndexById.put(object.getId(), i);              // only ever set on FIRST occurrence
} else if (objectTypes.indexOf(object.getType()) < objectTypes.indexOf(existing.getType())) {
  deduplicated.put(object.getId(), object);            // value updated, but firstIndexById is NOT
}
...
lastConsumedIndex = Math.max(lastConsumedIndex, firstIndexById.get(entry.getKey()));
```
When the same id is linked under two objectTypes and the *more specific* (winning) duplicate appears at a **later** raw index than the first (losing) occurrence, `firstIndexById` still points at the earlier index. If the page fills up at that entry, `nextOffset` is computed from the wrong (too-low) index, under-counting how much of the raw batch was actually consumed. The next "Load more" call re-fetches the raw row that produced the winning duplicate, and that entry gets shown again. Traced with concrete indices: raw `[(B,"1"), (A,"2"), (A,"1")]` with `objectTypes=[A,B]`, `limit=2` → both entries fill the page using raw indices 0-2, but `nextOffset` computes to `offset+2` instead of `offset+3`, re-serving row 2 on the next page. The new `testGetCategoryEntriesPaginationAcrossPagesDoesNotSkipOrRepeat` test only covers the ACL-filtered-drop case, not this ordering — this exact scenario is untested.

**Fix:** track the *max* raw index seen for each id (update it on every occurrence, not just the first) rather than only the first-seen index.

---

🟡 **Medium** — `categoryPlugins` list isn't thread-safe against concurrent cross-WAR self-registration
[`CategoryPluginServiceImpl.java:42-49`](https://github.com/Meeds-io/social/blob/d29992ad16e5d0e330ac45f455c74871fa79d93a/component/core/src/main/java/io/meeds/social/category/service/CategoryPluginServiceImpl.java#L42-L49)
```java
private List<CategoryPlugin> categoryPlugins = new ArrayList<>();
...
public void addPlugin(CategoryPlugin categoryPlugin) {
  categoryPlugins.add(categoryPlugin);        // plain ArrayList, mutated from each WAR's own @PostConstruct
}
```
Every plugin (social's own two, plus content's `NewsCategoryPlugin`, notes' `NoteCategoryPlugin`, and any future addon) calls `addPlugin()` from its own `@PostConstruct` as its WAR boots, while `getCategoryPlugin()` concurrently `.stream()`s the same list from request-handling threads. A plain `ArrayList` isn't safe under concurrent mutation/iteration — this used to be a Spring-populated, effectively-final `@Autowired List`, and the move to manual registration dropped that safety net.

**Fix:** use `CopyOnWriteArrayList` (cheap here — writes are boot-time-only, reads are hot).

---

🟢 **Nit — a few small ones**
- Recurring typo across new javadocs/Swagger text: "filter switch", "de-duplicated ... switch several objectTypes", "ordered switch last modification date" (should be "by") — looks like a bad find/replace. Appears in `CategoryLinkService.java`, `CategoryService.java`, and `CategoryRest.java`'s `@Operation` description (the last one is user-facing in generated Swagger docs).

What's good: the pagination fix (`nextOffset`) is a real, well-reasoned correction with a dedicated regression test and a clear doc comment explaining *why* naive `offset+limit` was wrong. The self-registration fix for `ActivityCategoryPlugin`/`SpaceCategoryPlugin` is correctly scoped and tested (`CategoryPluginServiceImplTest`).

---
Reviewed by Claude

boubaker (approved):
## Final review — all findings resolved

Re-verified against the latest commits. Everything from my prior rounds is now fixed, each with a dedicated regression test:
- Browse-by-category now resolves activity-linked News/Notes via `categoryPluginService.getObject(...)`, with `"activity"` auto-included in the storage query (`testGetCategoryEntriesResolvesActivityLinkedContent`).
- The cross-type duplicate pagination bug is fixed (`lastIndexById` tracks every occurrence, not just the first) and covered by `testGetCategoryEntriesDoesNotRepeatWinningDuplicateAcrossPages`, which reproduces the exact scenario I traced.
- `categoryPlugins` is now a `CopyOnWriteArrayList`.
- The recurring "switch" typos in javadoc/Swagger text are cleaned up.

Also reviewed the bundled `SpaceCategoryPlugin.getObject()` space-id normalization fix — correct and well-tested (`SpaceCategoryPluginUnitTest`).

Approving.

---
Reviewed by Claude

PR #5897, opened 2026-07-20
fix: Add accessible name to emojis - EXO-87562

akhanfir:
Rendered emojis had no accessible name, making them invisible to screen readers.

### DIFF
--- a/webapp/src/main/webapp/js/EmojiAccessibility.js
+++ b/webapp/src/main/webapp/js/EmojiAccessibility.js
@@ -1,0 +1,125 @@
+/**
+ * This file is part of the Meeds project (https://meeds.io/).
+ *
+ * Copyright (C) 2020 - 2026 Meeds Association contact@meeds.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+(function() {
+  // Accessibility (#4246): expose emojis to screen readers by wrapping
+  // them in a <span role="img" aria-label="..."> once their names are loaded.
+  let emojiNameByChar = null;
+  let emojiPattern = null;
+
+  function escapeRegExp(value) {
+    return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+  }
+
+  function buildEmojiIndex(emojiBank) {
+    const map = new Map();
+    (emojiBank.categories || []).forEach(category => {
+      (category.emojis || []).forEach(item => {
+        if (item.emoji && item.name && !map.has(item.emoji)) {
+          map.set(item.emoji, item.name);
+        }
+      });
+    });
+    emojiNameByChar = map;
+    const chars = Array.from(map.keys()).sort((a, b) => b.length - a.length);
+    emojiPattern = chars.length && new RegExp(chars.map(escapeRegExp).join('|'), 'g');
+  }
+
+  let ready = false;
+  const pendingCallbacks = new Map();
+  const readyPromise = fetch('/social/json/emojiBank.json?v=1')
+    .then(resp => resp.ok && resp.json())
+    .then(emojiBank => emojiBank && buildEmojiIndex(emojiBank))
+    .catch(() => {
+      // Emoji names not available: emojis will remain unlabeled.
+    })
+    .then(() => {
+      ready = true;
+      pendingCallbacks.forEach(callback => callback());
+      pendingCallbacks.clear();
+    });
+
+  function onReady(key, callback) {
+    if (ready) {
+      callback();
+    } else {
+      pendingCallbacks.set(key, callback);
+    }
+  }
+
+  function addAccessibleNameToEmojis(html) {
+    if (!emojiPattern) {
+      return html;
+    }
+    const container = document.createElement('div');
+    container.innerHTML = html;
+    const walker = document.createTreeWalker(container, NodeFilter.SHOW_TEXT);
+    const textNodes = [];
+    let node;
+    while ((node = walker.nextNode())) {
+      emojiPattern.lastIndex = 0;
+      if (!node.parentElement?.closest('[role="img"]') && emojiPattern.test(node.nodeValue)) {
+        textNodes.push(node);
+      }
+    }
+    textNodes.forEach(textNode => {
+      const fragment = document.createDocumentFragment();
+      let lastIndex = 0;
+      let match;
+      emojiPattern.lastIndex = 0;
+      while ((match = emojiPattern.exec(textNode.nodeValue))) {
+        if (match.index > lastIndex) {
+          fragment.appendChild(document.createTextNode(textNode.nodeValue.slice(lastIndex, match.index)));
+        }
+        const span = document.createElement('span');
+        span.setAttribute('role', 'img');
+        span.setAttribute('aria-label', emojiNameByChar.get(match[0]));
+        span.textContent = match[0];
+        fragment.appendChild(span);
+        lastIndex = match.index + match[0].length;
+      }
+      fragment.appendChild(document.createTextNode(textNode.nodeValue.slice(lastIndex)));
+      textNode.parentNode.replaceChild(fragment, textNode);
+    });
+    return container.innerHTML;
+  }
+
+  const readyMixin = {
+    data() {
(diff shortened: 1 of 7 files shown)


ahamdi (changes requested):
This file is for purifying the content, the emoji names handling should not go here

ahamdi (commented):
Automated JS review (Claude Code) appended below — correctness, performance, and maintainability findings on the emoji-accessibility change, alongside the earlier manual comments.

webapp/src/main/webapp/js/EmojiAccessibility.js
```diff
@@ -0,0 +1,96 @@
+/**
+ * This file is part of the Meeds project (https://meeds.io/).
+ *
+ * Copyright (C) 2020 - 2022 Meeds Association contact@meeds.io
```
ahamdi, line 4:
```suggestion
 * Copyright (C) 2020 - 2026 Meeds Association contact@meeds.io
```

webapp/src/main/webapp/js/sanitize-html-directive.js
```diff
@@ -17,6 +17,14 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 (function() {
+  function setSanitizedHtml(el, html) {
+    el.innerHTML = html && EmojiAccessibility.addAccessibleNameToEmojis(html) || '';
+    if (html && !EmojiAccessibility.isReady()) {
+      EmojiAccessibility.ready.then(() => {
+        el.innerHTML = EmojiAccessibility.addAccessibleNameToEmojis(html);
+      });
```
ahamdi, line 25: why should we add the content to el.innerHTML twice ?

webapp/src/main/webapp/js/EmojiAccessibility.js
```diff
@@ -0,0 +1,96 @@
...
+      return html;
+    }
+    const container = document.createElement('div');
+    container.innerHTML = html;
+    const walker = document.createTreeWalker(container, NodeFilter.SHOW_TEXT);
+    const textNodes = [];
+    let node;
+    while ((node = walker.nextNode())) {
+      emojiPattern.lastIndex = 0;
+      if (node.parentElement?.getAttribute('role') !== 'img' && emojiPattern.test(node.nodeValue)) {
```
ahamdi, line 65: **Correctness:** the re-wrap guard only checks the *immediate* parent's `role` attribute (`node.parentElement?.getAttribute('role') !== 'img'`), not the closest ancestor. If an emoji sits inside a nested inline element under an already-labelled `role="img"` ancestor (e.g. `<span role="img" aria-label="Party"><b>🎉</b></span>` from rich content or an oEmbed snippet), it gets wrapped a second time — a screen reader announces the container's label followed by a duplicate label for the same glyph.

Suggested fix: `!node.parentElement?.closest('[role="img"]')` instead of the direct-parent check.

webapp/src/main/webapp/vue-apps/activity-stream/components/activity/content/ActivityLink.vue
```diff
@@ -304,13 +305,15 @@ export default {
       return this.summary && this.$utils.htmlToText(this.summary) || '';
     },
     summaryElement() {
+      const purifiedHtml = this.summary && ExtendedDomPurify.purify(`<div>${this.summary}</div>`) || '';
       return {
```
ahamdi, line 309: **Correctness/perf:** `summaryElement`/`titleElement` return a brand-new object literal every time `emojiBankReady` flips, and that object is passed as the `child` prop into `DynamicHTMLElement`, which does `createElement(this.child)`. Vue treats the new object reference as a new component definition, so the moment `/social/json/emojiBank.json` resolves, every summary/title on the page gets destroyed and remounted — a visible flicker even when the activity itself hasn't changed.

Suggested fix: memoize and only replace the returned object when the computed `template` string actually differs.

webapp/src/main/webapp/vue-apps/activity-stream/components/activity/content/ActivityLink.vue
```diff
@@ -370,6 +373,9 @@ export default {
   created() {
     this.retrieveActivityProperties();
     window.addEventListener('resize', this.displayReadMore);
+    if (!this.emojiBankReady) {
+      EmojiAccessibility.ready.then(() => this.emojiBankReady = true);
+    }
   },
   mounted() {
```
ahamdi, line 380: **Correctness:** `displayReadMore()` is only recomputed on mount and on `window` resize. It isn't re-run when `emojiBankReady` flips and the summary/title re-render with wrapped emoji spans, unlike the sibling `ActivityBody.vue`, which added a `ResizeObserver` for the same kind of async content change. If emoji wrapping nudges the rendered height across the truncation threshold after the emoji index loads, the "read more" button state goes stale until the next window resize.

webapp/src/main/webapp/js/sanitize-html-directive.js
ahamdi, line 25: **Correctness (relates to the comment above on this file):** `setSanitizedHtml()` now does a second, asynchronous full `el.innerHTML` replacement once `EmojiAccessibility.ready` resolves, on top of the initial synchronous assignment, across every `v-sanitized-html`/`v-sanitized-html-no-embed` binding (27+ templates). Any code that wires up behavior on descendant nodes right after the initial render (attaching a handler, caching a child-node reference) will silently lose it when this deferred re-render swaps in new nodes, with no error raised.

webapp/src/main/webapp/vue-apps/activity-stream/components/activity/content/ActivityEmbeddedHTML.vue
```diff
@@ -89,7 +95,8 @@ export default {
   methods: {
     computeEmbeddedHTML(htmlElement) {
       const tempdiv = document.createElement('div');
-      tempdiv.innerHTML = ExtendedDomPurify.purify(htmlElement);
+      const purifiedHtml = ExtendedDomPurify.purify(htmlElement);
```
ahamdi, line 98: **Perf/cleanup:** `computeEmbeddedHTML()` re-runs the full `ExtendedDomPurify.purify()` pass a second time purely to re-apply emoji wrapping once `EmojiAccessibility` becomes ready, instead of caching the already-purified HTML. Since `embeddedHTML` depends on both `activity` and `emojiBankReady`, every mounted embedded-HTML card on the page re-sanitizes simultaneously the instant the shared readiness promise resolves — even cards with no emoji at all. Caching the purify result keyed by source HTML and only re-running the emoji wrap would avoid the duplicate work.

webapp/src/main/webapp/js/sanitize-html-directive.js
```diff
@@ -17,6 +17,14 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 (function() {
+  function setSanitizedHtml(el, html) {
+    el.innerHTML = html && EmojiAccessibility.addAccessibleNameToEmojis(html) || '';
+    if (html && !EmojiAccessibility.isReady()) {
```
ahamdi, line 22: **Cleanup:** `setSanitizedHtml()` registers a new `.then()` on the shared `EmojiAccessibility.ready` promise on every invocation while not yet ready, with no de-duplication. If an element is rebound multiple times before the emoji bank loads (e.g. during pagination), each rebind stacks another closure; all of them fire when the promise resolves and redundantly rebuild the DOM, with only the last one's result actually surviving. A per-element pending-value map (keyed by `el`) would collapse this to one scheduled callback per element.

webapp/src/main/webapp/vue-apps/common/components/DynamicHTMLElement.vue
```diff
@@ -1,8 +1,17 @@
 <script>
 export default {
+  data: () => ({
+    emojiBankReady: EmojiAccessibility.isReady(),
```
ahamdi, line 4: **Maintainability:** the "wait for `EmojiAccessibility` to become ready" boilerplate (`emojiBankReady` data flag + `created()` hook) is copy-pasted verbatim into three components (`DynamicHTMLElement.vue`, `ActivityEmbeddedHTML.vue`, `ActivityLink.vue`), while `sanitize-html-directive.js` implements yet another separate version of the same thing. Worth centralizing (e.g. a small Vue mixin exported alongside `EmojiAccessibility`) so a future fix to this pattern doesn't need to be applied in four places independently.

webapp/src/main/webapp/js/EmojiAccessibility.js
```diff
@@ -0,0 +1,96 @@
...
+        lastIndex = match.index + match[0].length;
+      }
+      fragment.appendChild(document.createTextNode(textNode.nodeValue.slice(lastIndex)));
+      textNode.parentNode.replaceChild(fragment, textNode);
+    });
+    return container.innerHTML;
+  }
+
+  window.EmojiAccessibility = {
+    isReady: () => ready,
```
ahamdi, line 92: **Maintainability:** the module only exposes `isReady()`/`ready` (a raw promise), pushing the "poll flag + then + re-render" responsibility onto every consumer instead of owning it once. This is the root cause of the duplication noted on `DynamicHTMLElement.vue` — a shared `readyMixin` (or equivalent helper) exported from this module would let every consumer opt in with one line instead of hand-rolling the pattern.

ahamdi (approved):

PR #5908, opened 2026-07-22
Fix: URL redirection not ok - EXO-88286

akhanfir:
Before this change, when sending a URL in chat or tasks and the link is displayed correctly, then clicking on the link, the URL opened is one that doesn't exist and a 404 page is displayed. To fix this problem, the encodeURI call is removed; the raw URL, already correctly encoded as returned by match.getUrl(), is used, consistent with Utils.js:toLinkUrl, which already works without this additional encoding. After this change, the link opens the correct URL.

### DIFF
--- a/webapp/src/main/webapp/js/ExtendedDomPurify.js
+++ b/webapp/src/main/webapp/js/ExtendedDomPurify.js
@@ -43,7 +43,7 @@
           case 'url' :
             const url = match.getUrl();
             const tag = match.buildTag();
-            tag.setAttr('href', encodeURI(url));
+            tag.setAttr('href', url);
             if (match.getUrl().indexOf('/') === 0 || match.getUrl().indexOf(window.location.origin) === 0) {
               return true;
             } else {


ahamdi (changes requested):
the title of the PR seems not to be related to the fix.
Besides we need to check why this encoding was there and what If it may cause regression on different modules using it

akhanfir:
> the title of the PR seems not to be related to the fix. Besides we need to check why this encoding was there and what If it may cause regression on different modules using it

ncodeURI was added as a side change alongside the apostrophe-decoding fix — not strictly needed for it. It only ever affected external links (Autolinker rebuilds the tag from scratch when replaceFn returns true, discarding it for internal links).

No regression risk: Utils.js:toLinkUrl builds hrefs the same way without encodeURI and already works fine (activity links). ExtendedDomPurify.purify() is the single shared path for all ~30 consumers (activity, comments, space descriptions, notifications, etc.), so the fix applies uniformly — it removes a double-encoding bug rather than introducing a new risk.

ahamdi (approved):

PR #5962, opened 2026-08-06
Merge mip 237

sofyenne:


boubaker:
## Review — MIP-237 (part of the core#138 · portal#1300 · social#5962 · meeds#4273 set)

Reviewed against the spec in [Meeds-io/MIPs#237](https://github.com/Meeds-io/MIPs/issues/237). Findings verified against head commit `1528190`.

---

🔴 **Critical** — Any authenticated user can enumerate (and CSV-export) the members of any group
[`UserRest.java:395-401`](https://github.com/Meeds-io/social/blob/15281904bbbc939d88a3e326f1d96a1b48d23915/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRest.java#L395-L401)

```java
if (!userACL.getSuperUser().equals(username)
    && !RestUtils.isMemberOfAdminGroup()
    && (CollectionUtils.isNotEmpty(groupIds) && !RestUtils.isOrganizationalUnitsManager(groupIds))
    && userType != null                    // <-- one big AND: the OU check is skipped when userType is null
    && !userType.equals(INTERNAL)) {
  throw new WebApplicationException(Response.Status.FORBIDDEN);
}
```

The new `listMembers` ACL check was merged by `&&` into the pre-existing "non-admins can't list external users" condition. With the default `userType=null` (or `userType=internal`), the whole conjunction is false and **no check runs**: any user in `/platform/users` can call `GET /rest/v1/social/users?groupId=/platform/administrators&returnSize=true` and get the full member list of any group — including per-user `isManager`/`isRedactor`/`isPublisher` flags added at [`EntityBuilder.java:413-417`](https://github.com/Meeds-io/social/blob/15281904bbbc939d88a3e326f1d96a1b48d23915/component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java#L413-L417). The `exportFile` branch ([`UserRest.java:432-449`](https://github.com/Meeds-io/social/blob/15281904bbbc939d88a3e326f1d96a1b48d23915/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRest.java#L432-L449)) passes `groupIds` into `UserExportService` with **no** check at all — and the old delegated-admin filtering inside `UserExportService.getUsers(...)` was removed, so the leak extends to CSV export with each user's group memberships. Contrast with the portal-side sibling [`UserRestResourcesV1.getUserMemberships`](https://github.com/Meeds-io/portal/blob/94b79f1d362669006d54ff7f65b3969c7e42f186/component/portal/src/main/java/org/exoplatform/portal/rest/UserRestResourcesV1.java#L568-L571), which does it correctly (`!isAdmin && !canListGroupMemberships(groupId)` → 401). The deleted `testGetUsersByDelegatedAdmin` was not replaced by an OU-manager equivalent, so no test covers this path.

**Fix:** make the group-scope check independent: `if (CollectionUtils.isNotEmpty(groupIds) && !isSuperUser && !RestUtils.isMemberOfAdminGroup() && !RestUtils.isOrganizationalUnitsManager(groupIds)) throw FORBIDDEN;` — keep the `userType` check as a separate condition, and apply the same guard before the `exportFile` branch. Add a REST test for a plain user passing `groupId`.

---

🟡 **Medium** — `ProfileFilter.spaceIdentityIds` is now silently ignored by profile search, without deprecation
[`ProfileSearchConnector.java:190`](https://github.com/Meeds-io/social/blob/15281904bbbc939d88a3e326f1d96a1b48d23915/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ProfileSearchConnector.java#L190)

```java
if (CollectionUtils.isNotEmpty(filter.getGroupIds())) {   // spaceIdentityIds no longer consulted
  String membershipPermissionsQuery = buildMembershipPermissionsExpression(filter);
```

The connector's permission clause switched from `filter.getSpaceIdentityIds()` to `filter.getGroupIds()`, but `ProfileFilter.spaceIdentityIds` (public API, used by addons) still exists and is now dead for user search: an external caller that sets it and searches gets **unfiltered** results — a scope widening that fails open. In-repo callers were all migrated, but the field carries no `@Deprecated` and no bridge.

**Fix:** in `buildQueryStatement`, keep honoring `spaceIdentityIds` (translate space identity ids → group tokens, or AND both clauses), and mark the setter `@Deprecated` pointing at `setGroupIds`.

---

🟡 **Medium** — Membership-type list cached forever in `UserPermissionService`
[`UserPermissionService.java:93-100`](https://github.com/Meeds-io/social/blob/15281904bbbc939d88a3e326f1d96a1b48d23915/component/core/src/main/java/io/meeds/social/identity/permission/service/UserPermissionService.java#L93-L100)

```java
if (membershipTypes == null) {
  membershipTypes = organizationService.getMembershipTypeHandler()
                                       .findMembershipTypes()...  // never invalidated
```

Admins can create membership types at runtime. The ES query expansion for a blank `membershipType` (`buildMembershipPermissionsExpression`) iterates this stale list, so users whose only membership in the OU uses a newly created type silently vanish from keyword search results, until restart. Also not thread-safe.

**Fix:** invalidate the cache via a `MembershipTypeEventListener` (or just don't cache — `findMembershipTypes` is cheap and cached upstream).

---

🟡 **Medium** — Per-row ACL identity resolution on a list endpoint (N+1)
[`EntityBuilder.java:413-417`](https://github.com/Meeds-io/social/blob/15281904bbbc939d88a3e326f1d96a1b48d23915/component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java#L413-L417)

```java
userEntity.setIsManager(userACL.isMemberOf(userACL.getUserIdentity(...remoteId), MANAGER, groupIds.getFirst()));
userEntity.getDataEntity().put("isRedactor", userACL.isMemberOf(userACL.getUserIdentity(...), ...));  // resolved again
userEntity.getDataEntity().put("isPublisher", userACL.isMemberOf(userACL.getUserIdentity(...), ...)); // and again
```

Every row of `GET /users?groupId=...` triggers three `userACL.getUserIdentity(remoteId)` resolutions (authenticator round-trips) — 300 for a 100-row page.

**Fix:** call `getUserIdentity` once per profile and reuse it for the three `isMemberOf` checks; better, derive the flags from the already-loaded `SOC_USER_PERMISSION` rows.

---

🟢 **Nit — a few small ones**
- [`UserPermissionBackfillUpgradePlugin.java`](https://github.com/Meeds-io/social/blob/15281904bbbc939d88a3e326f1d96a1b48d23915/component/core/src/main/java/io/meeds/social/identity/permission/upgrade/UserPermissionBackfillUpgradePlugin.java) — typo `upgradeSacceeded` → `upgradeSucceeded`.
- `UserPermissionStorage` — every write method uses `@CacheEvict(allEntries = true)` on a cache keyed by `userName`; during the backfill (250-user batches × N memberships) this thrashes the whole cache. `@CacheEvict(key = "#userPermission.userName")`-style eviction would do.

---

**What this PR does well:** the `SOC_USER_PERMISSION` denormalization is a sound design — it turns the previous per-group full-membership loads in `ProfileFilterListAccess` into a single indexed `EXISTS` (DB) / keyword-terms (ES) filter, kept in sync by properly thin listeners that delegate to the Service layer, with a resumable, batched backfill plugin. The ACL surface is correctly centralized in a `GroupAclPlugin` per the plugin-SPI pattern, cycle guards are present in both nested-group traversals, and test coverage on the new listeners/storage/upgrade plugins is genuinely thorough.

⚠️ **Classification: N1 (socle).** This diff touches ACL/permission checks, DB schema/Liquibase, JPA entities, and identity resolution — it requires **a priori human validation by an Archi/Dev, no auto-merge**, and the Critical finding above should block merge until fixed.

sofyenne:
Thanks for the detailed review — all findings are addressed. Two of them I
solved differently than proposed; rationale below.

### 🔴 Critical — group members enumeration & CSV export

Fixed. The group-scope check is now independent of the `userType` one:

- `groupIds` non-empty + not super user + not admin + not OU-manager of **all**
  of them → 403, evaluated before any branch, so the `exportFile` path is
  covered by the same guard.
- The pre-existing "non-admins can't list external users" check is kept as its
  own condition. Its old `/platform/delegated` bypass is replaced by the
  OU-manager bypass, so a manager keeps the capability delegated admins had —
  upstream semantics unchanged, only the ACL source swapped.

Added `UserRestResourcesTest#testGetUsersByGroupId` covering: plain user → 403
on both list and export, OU manager → 403 on a group they don't manage / 200 on
one they do, admin → 200 on any group.

### 🟡 `ProfileFilter.spaceIdentityIds` — removed instead of bridged

I went further than proposed and **deleted the field and its accessors** rather
than translating + deprecating.

The field has zero readers: nothing in this repo reads it since the connector
moved to `groupIds`, and across all sibling addons the only `Sp
hits are `SpaceUtils.getSpaceIdentityIds` (a different, still-u
agenda's own filter. Since the field is already broken in the f
direction, `@Deprecated` would only document the trap — removin
silent scope-widening into a compile error, and the migration i
`setGroupIds` + `SpaceUtils.getSpaceGroupIds` already used by `
`UserExportService`. Happy to fall back to deprecate-for-one-re
know of a consumer built outside the distribution BOM.

### 🟡 Membership type list cached forever — cache dropped entirely

`getMembershipTypes()` now fetches on demand; no field, no list

One correction to the finding: `findMembershipTypes` is **not**
`CacheableMembershipTypeHandlerImpl.findMembershipTypes()` just
`super`, which queries the IDM store. But the call site is only reached on the
OU member-listing search when no `membershipType` filter is set
with a handful of rows, so it isn't a hot path and the cache wa
much. Removing it kills the staleness bug and the thread-safety

### 🟡 Per-row ACL identity resolution (N+1)

Fixed — `getUserIdentity` is resolved once per profile and reus
`isMemberOf` checks, so 3 authenticator round-trips per row bec

### 🟢 Nits

- `upgradeSacceeded` → `upgradeSucceeded`.
- `@CacheEvict` now keyed (`#userName` / `#userPermission.userName`) on the five
  per-user writes; `deleteByGroupId` keeps `allEntries` since i
  member of the group.
- Also dropped a now-unused `NestedMembership` import in `UserE
  that checkstyle flags.

### Tests

`UserRestResourcesTest` 38/38, `ProfileSearchConnectorTest` 8/8
`UserExportServiceTest` 12/12 and the `UserPermission*` suites
full `component/core` suite runs 1025 tests with one failure,
`SpaceCategorySidebarPluginTest#testResolveProperties` — pre-ex
unrelated: it reproduces identically on this branch with these
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

boubaker:
## Re-review — head `d90f490`

All findings from the [initial review](https://github.com/Meeds-io/social/pull/5962#issuecomment-5268221776) were addressed:

| # | Finding | Status |
|---|---|---|
| 🔴 | Any authenticated user can enumerate/export any group's members (`UserRest.getUsers`) | ✅ **Fixed** — group-scope check is now independent and placed at the top of the method ([`UserRest.java:395-400`](https://github.com/Meeds-io/social/blob/d90f49095044801b6d9797f07614a9566189d768/component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRest.java#L395-L400)), covering both the listing and the `exportFile` branch. Regression test `testGetUsersByGroupId` covers plain user → 403, OU manager on foreign group → 403, own group → 200, admin → 200, export → 403. |
| 🟡 | `ProfileFilter.spaceIdentityIds` silently ignored by profile search | ✅ **Fixed** — field removed outright (fail-loud instead of fail-open). Verified by org-wide code search: no other Meeds-io repo references `ProfileFilter.get/setSpaceIdentityIds` (notes/content use the static `SpaceUtils.getSpaceIdentityIds`, which is kept), so no downstream compile break. Stricter than the §6 deprecation process, but defensible with zero external consumers. |
| 🟡 | Membership-type list cached forever in `UserPermissionService` | ✅ **Fixed** — cache removed, javadoc explains why. |
| 🟡 | Per-row triple ACL identity resolution in `EntityBuilder` (N+1) | ✅ **Fixed** — single `getUserIdentity` resolution per row. |
| 🟢 | Typo `upgradeSacceeded` | ✅ **Fixed** |
| 🟢 | `@CacheEvict(allEntries = true)` on per-user writes | ✅ **Fixed** — keyed by `userName`, with `allEntries` correctly kept (and commented) for `deleteByGroupId` only. |

### New items (minor)

- 🟢 **Nit** — [`UserRestResourcesTest.testGetUsersByGroupId`](https://github.com/Meeds-io/social/blob/d90f49095044801b6d9797f07614a9566189d768/component/service/src/test/java/org/exoplatform/social/rest/impl/users/UserRestResourcesTest.java#L1500-L1582): when no previous `group` ACL plugin existed, the test's mock plugin stays registered after the test (`UserACL` has no remove API and the `finally` only restores a non-null previous plugin). Harmless today, but a potential test-pollution source for future tests in the suite — consider registering a no-op plugin in the null case.

No other new issues found in the fix commit. The N1 classification stands: this remains an ACL + schema diff requiring a priori human validation, but the previously blocking finding is resolved.

boubaker (approved):

boubaker (approved):

PR #5974, opened 2026-08-11
Merge eXIP 7.3.0.13: Page searching

mkrout:
Resolves https://community.exoplatform.com/portal/g/:spaces:exo_product_management/home/notes/48136


boubaker:
## Review — eXIP 7.3.0.13 "Page Searching"

Cross-repo review of the three PRs implementing [the specification](https://community.exoplatform.com/portal/dw/activity?id=474261):

- **Meeds-io/social#5974** (this PR) — generic `page` connector, SPI, search + favorites
- Meeds-io/layout#562 — `PageUrlResolver` implementation + page lifecycle events
- Meeds-io/notes#1747 — `NotePageContentBlockPlugin` + reindex listener

Findings below are ordered most-severe first and cover all three repos.

---

🔴 **Critical** — `getById` runs an unfiltered ES query, leaking any indexed page's content to any user

[`PageContentSearchConnector.java:146-160`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentSearchConnector.java#L146-L160) · [`PageSearchRest.java:72-86`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/service/src/main/java/io/meeds/social/cms/rest/PageSearchRest.java#L72-L86)

```java
  public PageSearchResult getById(String id, Locale locale) {
    ...
    String esQuery = """
        {
          "query": {
            "terms": {"_id": ["%s"]}   // no buildPermissionFilter() anywhere in this query
          }
        }
        """.formatted(id);
```

`search()` composes `buildPermissionFilter()` into every query; `getById()` composes nothing. The `GET /social/rest/pages/{id}` endpoint that calls it has no `@Secured` either — unlike every sibling controller in `io.meeds.social.*.rest`, which all carry `@Secured("users")`. So the page title, `pagePath`, author and a 150-char content excerpt of *any* indexed block are readable by any caller who has the id, and a user who bookmarked a page and later lost access keeps seeing its content in the favorites drawer. The spec's Security section is explicit that the indexed ACL field is what gates reads — that gate is simply absent on this path.

**Fix:** build the query as a `bool` with `must: {terms: {_id: [...]}}` and `filter: [buildPermissionFilter()]`, exactly as `search()` does, and add `@Secured("users")` to `PageSearchRest`.

---

🔴 **Critical** — Raw path variable spliced into the ES query JSON (query injection)

[`PageContentSearchConnector.java:151-157`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentSearchConnector.java#L151-L157)

```java
        """.formatted(id);   // id is an unvalidated @PathVariable
```

Compare [line 128](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentSearchConnector.java#L128), where `search()` passes the term through `escape()` before templating. `getById` skips that entirely, so a percent-encoded `"` in the path segment (Tomcat decodes `%22` and hands it to `@PathVariable` intact) closes the JSON string and lets the caller replace the whole query body — e.g. a `match_all` that dumps `page_content_alias`. Combined with the missing permission filter above, that is a full read of every indexed page's content.

**Fix:** don't template the id into JSON text — validate it against the known id shape (`page_<digits>(_<hex>)?`) and reject otherwise, or JSON-escape it before substitution.

---

🟠 **High** — Nothing ever unindexes: deleted pages and detached blocks stay searchable

[`PageContentBlockIndexingListener.java:99-109`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/cms/listener/PageContentBlockIndexingListener.java#L99-L109)

```java
  /**
   * Note: this doesn't unindex a block that was just detached from the
   * page — the caller that removes a {@link CMSSetting} binding is
   * responsible for unindexing that specific block's own id.
   */
```

`grep -n "unindex"` across all three PRs hits only test assertions (`verify(indexingService, never()).unindex(...)`) — the responsible caller the javadoc defers to does not exist in this changeset. Delete a page, remove the Single Note View from it, or delete the note, and the ES document survives with its *old* permissions, so it keeps appearing in results (and stays reachable through `getById`). The spec calls this out — *"Existing page lifecycle events (LayoutService) trigger unindex on page deletion"* — and all three sibling connectors implement it ([`SpaceIndexingListenerImpl.java:71`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/search/listener/SpaceIndexingListenerImpl.java#L71), `CategoryModifiedIndexingListener`, `ActivityESListener`).

**Fix:** compare the block ids currently bound to the page against those previously indexed and `indexingService.unindex(TYPE, id)` the difference; add a page-deletion listener in Layout broadcasting the event Social needs.

---

🟠 **High** — Layout and Notes PR builds are red; Notes reaches into Social's `component-core`

[`NoteContentIndexingListener.java:36`](https://github.com/Meeds-io/notes/blob/6dce77671e16c92c3b9115f346b20eb3f4ca5988/notes-service/src/main/java/io/meeds/notes/listener/NoteContentIndexingListener.java#L36)

```java
import io.meeds.social.cms.storage.elasticsearch.PageContentIndexingConnector;
// notes/CI: "package io.meeds.social.cms.storage.elasticsearch does not exist"
```

Both PR builds fail to compile (layout: 4 × `cannot find symbol PageUrlResolver/PageUrlResolverService`; notes: the above plus 6 more). Part of that is just merge ordering — Social must publish first. But the Notes half is structural: the SPI contract lives in `component/api`, while `PageContentIndexingConnector.TYPE` and `buildBlockId(...)` live in `component/core`'s ES storage package. Our extensibility norm is explicit that plugins may call the Service layer only, never DAO/Storage; here Notes has to know Social's Elasticsearch document-id format to do its job.

**Fix:** expose a Service method on `PageContentBlockPluginService` (component/api) such as `reindexContentBlock(String contentType, String settingName)` that resolves the page and id internally, and let both listeners call that instead of composing ids themselves.

---

🟠 **High** — Indexing is a full-table scan per document (O(N²) reindex)

[`PageContentIndexingConnector.java:141-155`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentIndexingConnector.java#L141-L155) · [`227-244`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentIndexingConnector.java#L227-L244) · [`CMSStorage.java:51-56`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/cms/storage/CMSStorage.java#L51-L56)

```java
  public List<String> getAllIds(int offset, int limit) {
    return pluginService.getContentTypes().stream()
        .flatMap(type -> cmsService.getSettingsByType(type).stream()   // whole table, no Pageable
            ...
            .map(s -> buildBlockId(type, s)))                          // layoutService.getPage() per setting
        .distinct().skip(offset).limit(limit)                          // paging done in memory
```

`getSettingsByType` has no `Pageable` and loads every metadata item of the type; `buildBlockId` then hits `layoutService.getPage()` once per setting. Because `skip/limit` are applied afterwards, a paged full reindex repeats that entire scan for *every* batch, and `create(id)` repeats it again per document via `findContentBlock`. The sibling [`CategoryIndexingConnector.getAllIds`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/category/storage/elasticsearch/CategoryIndexingConnector.java#L137) pushes offset/limit down to the storage instead. Our backend norm states almost every query must take a `Pageable` and must not fetch a whole table.

**Fix:** add a paginated `getSettingsByType(type, offset, limit)` on `CMSStorage`/`CMSService`, and give the connector a direct `(contentType, settingName) → CMSSetting` lookup so `create()` resolves one row instead of scanning.

---

🟡 **Medium** — Only 2 of Layout's ~7 page-mutating paths broadcast

[`PageLayoutService.java:363`](https://github.com/Meeds-io/layout/blob/2a3df7670572352e59d070a1fc1f93ac55dce0da/layout-service/src/main/java/io/meeds/layout/service/PageLayoutService.java#L363) · [`437`](https://github.com/Meeds-io/layout/blob/2a3df7670572352e59d070a1fc1f93ac55dce0da/layout-service/src/main/java/io/meeds/layout/service/PageLayoutService.java#L437)

```java
    layoutService.save(existingPage);
    listenerService.broadcast(PAGE_UPDATED_EVENT, username, pageKey.format());
```

`createPage` (:220), `clonePage` (:256), `updatePageApplicationPreferences` (:296/:303) and `restorePageLayout` (:378) all persist page state without broadcasting. `updatePageApplicationPreferences` is the sharpest: the Single Note View portlet preference is what names the `CMSSetting`, so changing it swaps which note the page shows while the index keeps serving the old content.

**Fix:** broadcast `PAGE_UPDATED_EVENT` from every method that persists a page, `updatePageApplicationPreferences` and `restorePageLayout` first.

---

🟡 **Medium** — `canCreateFavorite` throws `NumberFormatException` on a malformed object id

[`PageFavoriteACLPlugin.java:48-51`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/org/exoplatform/social/core/plugin/PageFavoriteACLPlugin.java#L48-L51)

```java
  public boolean canCreateFavorite(Identity userIdentity, String objectId) {
    Page page = layoutService.getPage(PageContentIndexingConnector.parsePageId(objectId));
```

`parsePageId` ends in `Long.parseLong(...)` with no guard. `objectId` arrives from the generic favorites REST API and is fully caller-controlled, so `POST` a favorite of type `page` with id `abc` and the plugin throws out of the ACL check — a 500 where the contract wants `false` (or a 400). `PageFavoriteACLPluginTest` covers page-not-found and both access outcomes, but never a non-numeric id.

**Fix:** catch/validate in `canCreateFavorite` and return `false` for an unparseable id; add the missing test case.

---

🟡 **Medium** — Per-result favorite lookup is an N+1 on every search page

[`PageContentSearchConnector.java:205-215`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentSearchConnector.java#L205-L215)

```java
  private boolean isFavorite(String id) {
    ...
    Identity socialIdentity = identityManager.getOrCreateUserIdentity(username);  // per result
    return favoriteService.isFavorite(new Favorite(..., id, ...));                // per result
```

`buildResult` calls this for each hit, so a 20-result page costs 20 identity resolutions plus 20 favorite queries. This is the list-endpoint N+1 pattern the N2 checklist asks to be checked for.

**Fix:** resolve the identity once per `search()` call and batch the favorite check across the hit ids.

---

🟡 **Medium** — Excerpt language matching can silently miss on region-qualified locales

[`PageContentSearchConnector.java:283-290`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentSearchConnector.java#L283-L290)

```java
    String langField = PageContentIndexingConnector.contentFieldName(locale.toLanguageTag());
```

Index-side field names come from `localeConfigService.getLocalConfigs()` and from Notes' `getPageAvailableTranslationLanguages` — both bare codes like `fr`. Query-side, `locale` is `request.getLocale()`. If that resolves to `fr-FR` rather than `fr`, `langField` is `content-fr-FR`, `hasTranslation` is false, and the user silently gets the default-language excerpt instead of their translation. Worth confirming what the request-locale filter actually yields before merge; the test suite only exercises `Locale.FRENCH`, which is exactly the case that works.

**Fix:** normalize the incoming locale through `LocaleConfigService` before building the field name, and add a `Locale("fr","FR")` test.

---

🟢 **Nit — a few small ones**

- [`PageContentIndexingConnector.java:265-267`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentIndexingConnector.java#L265-L267) — `Integer.toHexString(Objects.hash(...))` is a 32-bit hash; two blocks colliding on the same page silently share one document. A truncated SHA-256 costs nothing here.
- [`component/core/src/test/resources/page-search-query.json`](https://github.com/Meeds-io/social/blob/b8eba4c74db7e6e8dd4c9edf1c27882cd83be554/component/core/src/test/resources/page-search-query.json) omits `@space_filter@`, which the production template has — the test resource can't exercise space filtering.
- `pagePath` is frozen at index time; renaming or moving the navigation node leaves search results pointing at a dead URL, with no reindex trigger.
- `SearchPageCard.vue` declares a `term` prop that is never used.
- `PageSearchResult.siteName` actually carries the *resolved label*, not the site name — worth renaming while the DTO is still new.
- The `component-plugins-configuration.xml` diff is mostly CRLF→LF churn unrelated to the change; reverting it would make the real 9-line addition visible.
- Sonar's quality gate is red on this PR — worth clearing rather than merging over.

---

The extension design is the strong part here: `PageContentBlockPlugin` + `PageUrlResolver` keep Social genuinely free of any Notes or Layout compile dependency, and the `PageUrlResolver` indirection is a neat answer to the real constraint that routing primitives live in `webui`. The multi-language excerpt handling is more careful than it needed to be — `matchedContentInAnyLanguage` correctly distinguishes "matched via title, no excerpt" from "matched only in a language we won't show you", and each branch has a test. The `EXO-89077` wildcard-membership regexp fix carries a comment that explains *why* the escaping is doubled, which is exactly the kind of note that saves the next reader an hour. Test coverage is broad across all three repos.

**Classification: N1.** The diff touches ACL/permission filtering, a trust boundary (client-controlled path variable reaching an ES query), and a new REST surface — this needs an Architect/Dev review before merging and must not be auto-approved on AI review alone. The two Critical findings and the missing unindex path should block.

mkrout:
**Fixed and verified today** (2 Criticals, 3 High, 3 Medium — 57/57 tests pass in social alone, plus 4/9 in notes/layout confirming those builds are clean):

- Critical — unfiltered getById → permission-filtered.
- Critical — query injection → id shape validated, @Secured("users") added.
- High — nothing ever unindexes → detached blocks now cleaned up via findIndexedBlockIds diff.
- High — Notes reaching into Social's storage package → new PageContentBlockPluginService.reindexContentBlock() API on the proper component/api boundary; NoteContentIndexingListener simplified to a one-line delegate.
- Confirmed "Layout/Notes builds are red" was pure merge-ordering — both build clean against the fixed Social jars, nothing to change there.
- Medium — malformed favorite id → returns false instead of throwing.
- Medium — N+1 favorite lookup → batched into one getFavoriteItemsByCreatorAndType call per search(), checked via Set.contains per result.
- Medium — locale mismatch (fr-FR vs fr) → new resolveContentLanguageTag() matches the request locale's language against configured locales before building the field name, so a region-qualified locale still finds its translation.

**Still open, deliberately left alone:**

- High — O(N²) reindex scan (needs a new paginated MetadataService API + safe spaceId handling — a real design task, not a quick fix).
- Nits: 32-bit hash collision risk, stale pagePath after a page move, unused Vue term prop, PageSearchResult.siteName actually holding a resolved label, CRLF diff noise.

boubaker:
## Review round 2 — eXIP 7.3.0.13 "Page Searching"

Re-reviewed at `3efe7a4` (social), `2a3df76` (layout, unchanged), `14462ab` (notes), against [round 1](https://github.com/Meeds-io/social/pull/5974#issuecomment-5256465388).

Both Critical findings are properly closed, and social is now green on build **and** Sonar. Notes is genuinely decoupled — its remaining CI errors no longer mention `io.meeds.social.cms.storage.elasticsearch` at all, only the unpublished `component/api` snapshot, which is pure merge ordering.

### Round 1 status

| # | Sev | Finding | Status |
|---|---|---|---|
| 1 | 🔴 | `getById` ran an unfiltered ES query (any user could read any page's content) | ✅ **Fixed** — `bool`/`filter` with `buildPermissionFilter()`, `@Secured("users")` |
| 2 | 🔴 | Raw `@PathVariable` spliced into the ES query JSON (injection) | ✅ **Fixed** — `BLOCK_ID_PATTERN` guard, returns `null` → 404 rather than throwing (better: can't probe which ids exist) |
| 3 | 🟠 | Nothing ever unindexes | ⚠️ **Partial** — detached-block reconcile added but inert, see **A**; page deletion still open, see **D** |
| 4 | 🟠 | Notes reaching into Social's `component/core`; both builds red | ✅ **Fixed** structurally — Notes now calls `reindexContentBlock` only. Builds still red on merge ordering alone |
| 5 | 🟠 | Indexing is a full-table scan per document (O(N²)) | ❌ **Open** — see **C** |
| 6 | 🟡 | Only 2 of ~7 Layout page-mutating paths broadcast | ❌ **Open** — and it now blocks #3, see **B** |
| 7 | 🟡 | `canCreateFavorite` threw `NumberFormatException` on a malformed id | ✅ **Fixed** |
| 8 | 🟡 | Per-result favorite lookup was an N+1 | ✅ **Fixed** — `resolveFavoriteIds()` resolves once per call, `Set` lookup per hit (better than proposed) |
| 9 | 🟡 | Region-qualified locales missed the translated excerpt | ✅ **Fixed** — `resolveContentLanguageTag`, with a `fr-FR` test |
| 10 | 🟢 | `blockHash` is a 32-bit `Objects.hash` | ❌ **Open** |
| 11 | 🟢 | Test query template diverged from production | ✅ **Fixed** — unified on `"@filters@"` |
| 12 | 🟢 | `pagePath` frozen at index time, stale on navigation rename/move | ❌ **Open** |
| 13 | 🟢 | `SearchPageCard.vue` declares an unused `term` prop | ⛔ **Withdrawn** — sibling search cards do the same; it's the card contract, not dead code |
| 14 | 🟢 | `PageSearchResult.siteName` actually carries the resolved *label* | ❌ **Open** |
| 15 | 🟢 | CRLF churn in `component-plugins-configuration.xml` | ❌ **Open** — still 33 changed lines for a 9-line addition |
| 16 | 🟢 | Sonar quality gate red | ✅ **Fixed** — green |

---

### A. 🟠 High — The new orphan-block cleanup silently does nothing: Elasticsearch rejects `wildcard` on `_id`

[`PageContentSearchConnector.java:207-215`](https://github.com/Meeds-io/social/blob/3efe7a4ac37a97ee4b7336ff2aa7f66d81ff3c86/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentSearchConnector.java#L207-L215)

```java
    String esQuery = """
        {
          "query": {"wildcard": {"_id": "%s_*"}},   // ES: query_shard_exception
```

`_id` is a metadata field mapped with `TextSearchInfo.SIMPLE_MATCH_ONLY`, so ES answers `"Can only use wildcard queries on keyword, text and wildcard fields - not on [_id] which is of type [_id]"` — in both 7.x and 8.x. The failure is invisible: `ElasticSearchingClient.sendRequest` logs a rejected-search audit entry and **returns the error body** instead of throwing, so `findIndexedBlockIds` parses that JSON, finds no `hits` key, and returns an empty list. Net effect — `unindexDetachedBlocks` never unindexes anything, and every page-layout save quietly logs a rejected ES search. `shouldFindIndexedBlockIdsMatchingThePageStorageId` can't catch this: it stubs the client and asserts on the query string.

**Fix:** index the page's storage id as a real `keyword` field (add `"pageStorageId": {"type": "keyword"}` to `ES_MAPPING` and `fields.put("pageStorageId", page.getStorageId())` in `create`), then query `{"term": {"pageStorageId": "..."}}`.

---

### B. 🟠 High — The detached-block path this revision adds is unreachable for the case its javadoc names

[`PageContentBlockIndexingListener.java:128-143`](https://github.com/Meeds-io/social/blob/3efe7a4ac37a97ee4b7336ff2aa7f66d81ff3c86/component/core/src/main/java/io/meeds/social/cms/listener/PageContentBlockIndexingListener.java#L128-L143) · [`PageLayoutService.java:296-332`](https://github.com/Meeds-io/layout/blob/2a3df7670572352e59d070a1fc1f93ac55dce0da/layout-service/src/main/java/io/meeds/layout/service/PageLayoutService.java#L296-L332)

```java
   * A content block can be detached from a page (removed from its layout,
   * or its content-block portlet preference repointed elsewhere) without   // <- never fires
```

Layout still broadcasts from only two methods (`updatePageLayout`, `updatePagePermissions`). `updatePageApplicationPreferences` — the method that repoints a Single Note View's portlet preference, i.e. exactly the case the javadoc claims to handle — persists via `layoutService.save(application.getState(), portletPreferences)` and broadcasts nothing. `createPage` and `restorePageLayout` are likewise silent, so a page created from a template that already carries a block, or restored to a shipped layout, never enters the index.

**Fix:** broadcast `PAGE_UPDATED_EVENT` from `updatePageApplicationPreferences`, `createPage` and `restorePageLayout` in Meeds-io/layout#562 — or drop the preference-repointing claim from the javadoc.

---

### C. 🟠 High — Indexing is still a full-table scan per document

[`PageContentIndexingConnector.java:154-161`](https://github.com/Meeds-io/social/blob/3efe7a4ac37a97ee4b7336ff2aa7f66d81ff3c86/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentIndexingConnector.java#L154-L161) · [`:247`](https://github.com/Meeds-io/social/blob/3efe7a4ac37a97ee4b7336ff2aa7f66d81ff3c86/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentIndexingConnector.java#L247)

```java
        .flatMap(type -> cmsService.getSettingsByType(type)   // whole table, no Pageable
            ...
            .map(s -> buildBlockId(type, s)))                 // getPage() per setting
        .distinct().skip(offset).limit(limit)                 // paged only afterwards
```

Unchanged since round 1. `findContentBlock` (line 247) still re-scans every setting of the content type once per indexed document. An indexed path is already available — a `CMSSetting`'s page reference is stored as the `MetadataItem`'s metadata *name*, so `getMetadataItemsByMetadataNameAndTypeAndObject(pageReference, METADATA_TYPE, type, 0, -1)` gives a page-scoped lookup. Moving `skip`/`limit` ahead of the `getPage()` resolution would also stop a full reindex re-resolving every page on every batch.

**Fix:** add `getSettingsByTypeAndPageReference(type, pageReference)` on `CMSStorage`/`CMSService`, use it in `findContentBlock`, and page the settings before resolving pages in `getAllIds`.

---

### D. 🟡 Medium — A deleted page's blocks are still never unindexed, though the portal event for it exists

[`PageContentBlockIndexingListener.java:47-51`](https://github.com/Meeds-io/social/blob/3efe7a4ac37a97ee4b7336ff2aa7f66d81ff3c86/component/core/src/main/java/io/meeds/social/cms/listener/PageContentBlockIndexingListener.java#L47-L51)

```java
 * Note: a page that is deleted outright never fires either event (Layout
 * has no page-deletion broadcast today), so its previously-indexed blocks
 * are only cleaned up once Layout adds one
```

Honest documentation, but the premise is avoidable: the portal already broadcasts `LayoutService.PAGE_REMOVED` (`org.exoplatform.portal.config.DataStorage.pageRemoved`), which covers every deletion path including the bulk `removePages(siteKey)` that `SiteLayoutService.deleteSite` performs. No Layout change needed. Until then a deleted page keeps a live document, still matching searches under the ACL it carried when last indexed.

**Fix:** add a listener on `LayoutService.PAGE_REMOVED` in Social, take the removed `Page`'s storage id from the event payload, and unindex its blocks.

---

### E. 🟡 Medium — `@Secured("users")` landed on `getById` but not on `search`

[`PageSearchRest.java:53`](https://github.com/Meeds-io/social/blob/3efe7a4ac37a97ee4b7336ff2aa7f66d81ff3c86/component/service/src/main/java/io/meeds/social/cms/rest/PageSearchRest.java#L53) vs [`:74-75`](https://github.com/Meeds-io/social/blob/3efe7a4ac37a97ee4b7336ff2aa7f66d81ff3c86/component/service/src/main/java/io/meeds/social/cms/rest/PageSearchRest.java#L74-L75)

```java
  @GetMapping(value = "search", produces = MediaType.APPLICATION_JSON_VALUE)
  // no @Secured — the sibling endpoint below has one
```

Anonymous callers fall through `buildPermissionFilter()`'s no-`ConversationState` branch to an `Everyone`-only filter, so nothing leaks — but every other controller under `io.meeds.social.*.rest` carries `@Secured("users")`, and the two endpoints on this one now disagree. If the asymmetry is deliberate (public-site search), it deserves a comment; otherwise it's an oversight.

**Fix:** add `@Secured("users")` to `search`, or comment why anonymous search is intended.

---

Round two is a real improvement. The ACL and injection holes are properly closed; `resolveFavoriteIds()` is a better answer to the N+1 than the one I proposed; and the Notes decoupling is structural rather than cosmetic. The added tests track the fixes closely, including the region-qualified-locale case that the original code silently got wrong.

**Classification stays N1.** The two remaining High findings are both "implemented but inert" — the orphan cleanup cannot run at all, and the Layout event that would drive it isn't broadcast — so the stale-document behaviour flagged in round 1 is still live in practice despite the code now being present.

mkrout:
## Fixes applied — response to round 2 review

Pushed at `e88c409f32`. Covers every High/Medium finding plus the actionable nits from [round 2](https://github.com/Meeds-io/social/pull/5974#issuecomment-5266617972).

---

### A. 🟠 `findIndexedBlockIds` was inert (ES rejects `wildcard` on `_id`) — ✅ Fixed

Added a real `pageStorageId` keyword field to the ES mapping (`PageContentIndexingConnector`), populated at index time from `page.getStorageId()`. `findIndexedBlockIds` now queries `{"term": {"pageStorageId": "..."}}` instead of the wildcard ES was silently rejecting. The orphan-block cleanup added last round now actually runs.

### C. 🟠 O(N²) indexing — ✅ Partially fixed

Added `CMSStorage`/`CMSService#getSettingsByTypeAndPageReference(type, pageReference)`, exploiting the fact that the page reference is itself the metadata name (`getMetadataItemsByMetadataNameAndTypeAndObject`). `PageContentIndexingConnector#findContentBlock` (one call per reindexed document) and `PageContentBlockIndexingListener#findContentBlockIds` (one call per page save) now use it instead of scanning every setting of the content type.

Left open: `getAllIds()` still does a full-table scan before paging — a correct fix needs cross-type pagination on `MetadataService` that doesn't exist yet. Judged too large to fold into this round.

### D. 🟡 Deleted page never unindexed — ✅ Fixed

Turned out `LayoutService.PAGE_REMOVED` was declared but never actually broadcast by `LayoutServiceImpl` — fixed upstream (see [gatein-portal EXO-88570](https://github.com/Meeds-io/gatein-portal/commit/dbf969b02e469b06bc2143da584109f9fc839461), branch `EXO-88570`, not yet a PR). On the social side, new `PageRemovedIndexingListener` reacts to that event and unindexes every block still indexed under the removed page's storage id.

### E. 🟡 `@Secured("users")` missing on `search` — ✅ Fixed

Added, matching the platform's generic `SearchRest` endpoint which already requires `@RolesAllowed("users")` for the same unified search bar.

### Nits

| # | Finding | Status |
|---|---|---|
| 10 | 32-bit `blockHash` collision risk | ✅ Fixed — truncated 64-bit SHA-256 digest instead of `Integer.toHexString(Objects.hash(...))` |
| 11 | Test template diverged from production | ✅ Fixed last round (`"@filters@"` placeholder unified) |
| 12 | `pagePath` frozen at index time | ❌ Still open — needs a listener on navigation rename/move, out of scope for now |
| 14 | `PageSearchResult.siteName` actually held the resolved label | ✅ Fixed — renamed to `siteLabel` (DTO, tests, `SearchPageCard.vue`, `PageFavoriteItem.vue`) |
| 15 | CRLF churn in `component-plugins-configuration.xml` | ✅ Fixed — diff is now a clean +9/-0 instead of 57 changed lines |
| 16 | Sonar quality gate red | ✅ Fixed last round |

### Tests

`component/core`: 1014/1015 — the one remaining failure (`SpaceCategorySidebarPluginTest`) is pre-existing and unrelated, confirmed flaky across multiple runs, not touched by this diff.

boubaker:
## Review round 3 — eXIP 7.3.0.13 "Page Searching"

Re-reviewed at `e88c409` (social), `b1f3d5a` (layout), `14462ab` (notes, unchanged), plus the new Meeds-io/portal#1303 at `dbf969b`.

Previous rounds: [round 1](https://github.com/Meeds-io/social/pull/5974#issuecomment-5256465388) · [round 2](https://github.com/Meeds-io/social/pull/5974#issuecomment-5266617972)

Every round-2 finding is addressed except one partial. Two new issues, both in the same file, and one of them is what's holding the Sonar gate red.

### Round 2 status

| # | Sev | Finding | Status |
|---|---|---|---|
| A | 🟠 | Orphan cleanup inert — ES rejects `wildcard` on `_id` | ✅ **Fixed** — `"pageStorageId": {"type": "keyword"}` added to the mapping, set in `create`, query is now `{"term": {"pageStorageId": …}}` |
| B | 🟠 | Detached-block path unreachable — Layout didn't broadcast | ✅ **Fixed** — `createPage`, `updatePageApplicationPreferences` and `restorePageLayout` all broadcast now, each with a test |
| C | 🟠 | Indexing is a full-table scan per document | ⚠️ **Partial** — `findContentBlock` now uses `getSettingsByTypeAndPageReference` (the per-document quadratic is gone). `getAllIds` unchanged, see **G** |
| D | 🟡 | Deleted page's blocks never unindexed | ✅ **Fixed** — portal#1303 + `PageRemovedIndexingListener`. The insertion point is right: `removePages(SiteKey)` → `remove(PageKey)` → `remove(Page)`, so one broadcast covers every deletion path |
| E | 🟡 | `@Secured("users")` on `getById` but not `search` | ✅ **Fixed** |

### Round 1 leftovers

| # | Sev | Finding | Status |
|---|---|---|---|
| 10 | 🟢 | `blockHash` was a 32-bit `Objects.hash` | ✅ **Fixed** — SHA-256 truncated to 16 hex chars |
| 15 | 🟢 | CRLF churn in `component-plugins-configuration.xml` | ✅ **Fixed** — clean 9-line addition now |
| 12 | 🟢 | `pagePath` frozen at index time, stale on navigation rename/move | ❌ Open |
| 14 | 🟢 | `PageSearchResult.siteName` carries the resolved *label* | ❌ Open |

---

### F. 🟠 High — A raw NUL byte is embedded in the source as a character literal

[`PageContentIndexingConnector.java:294-300`](https://github.com/Meeds-io/social/blob/e88c409f3293ea313b70e730b2a102d7e0497ebe/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentIndexingConnector.java#L294-L300)

```java
      byte[] hash = digest.digest((contentType + '<0x00>' + settingName).getBytes(StandardCharsets.UTF_8));
      //                                         ^ a literal 0x00 byte in the file, not the escape '\0'
```

The separator is written as an actual NUL byte between the quotes rather than as `'\0'`. Consequences, in order of how much they'd hurt:

1. **It's load-bearing and invisible.** It's the separator that keeps `("note", "pageX")` from hashing the same as `("notep", "ageX")`. Any editor, formatter, `sed` pass or transfer that normalizes control characters drops it silently — and every document id in the index changes, orphaning the whole `page_content` index with no error anywhere.
2. **Tooling treats the file as binary.** `file` reports `data`, `grep` suppresses matches, and `git diff` can't be trusted on it. I only found this by dumping bytes.
3. **It's the red gate.** SonarCloud reports it as CRITICAL at line 297 — *"Remove the non-escaped  character from this literal"* — and it is the only blocker on an otherwise-green analysis.

**Fix:** write `'\0'` (or `''`). Identical bytes at runtime, plain-ASCII source, gate goes green.

---

### G. 🟠 High — Line endings of the whole connector flipped to CRLF, hiding the real change

[`PageContentIndexingConnector.java`](https://github.com/Meeds-io/social/blob/e88c409f3293ea313b70e730b2a102d7e0497ebe/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentIndexingConnector.java) — 345/345 lines now CRLF; it was 0/345 at `3efe7a4`.

```
 .../PageContentIndexingConnector.java | 672 +++++++++++----------
```

The genuine change in this file is small — a `pageStorageId` mapping entry, a `fields.put`, the page-scoped lookup in `findContentBlock`, and the new SHA-256 `blockHash`. It renders as 672 changed lines, so a reviewer diffing this revision sees the entire file rewritten and has no realistic way to spot **F** inside it. This is the same class of problem as round 1's XML churn, which was correctly cleaned up in this very revision.

**Fix:** convert the file back to LF. Worth adding `*.java text eol=lf` to `.gitattributes` so it can't recur.

---

### H. 🟡 Medium — `getAllIds` still resolves a page per setting, on every batch

[`PageContentIndexingConnector.java:155-168`](https://github.com/Meeds-io/social/blob/e88c409f3293ea313b70e730b2a102d7e0497ebe/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentIndexingConnector.java#L155-L168)

```java
        .flatMap(type -> cmsService.getSettingsByType(type)   // whole content type
            ...
            .map(s -> buildBlockId(type, s)))                 // layoutService.getPage() per setting
        .distinct().skip(offset).limit(limit)                 // …only paged afterwards
```

The half of finding C that matters per-document is fixed; this half isn't. A full reindex calls `getAllIds` once per batch, and each call resolves *every* page in the platform to build ids it then discards. Moving `skip`/`limit` above the `buildBlockId` map fixes it, but the stream is unordered — `getSettingsByType` carries no `ORDER BY` and `getContentTypes()` is a `HashMap` key set — so paging needs a deterministic sort to avoid skipping or repeating blocks across batches.

**Fix:** sort by (contentType, settingName), apply `skip`/`limit`, then resolve pages.

---

### I. 🟢 Nit — `PageRemovedIndexingListener` uses the legacy listener base

[`PageRemovedIndexingListener.java:43`](https://github.com/Meeds-io/social/blob/e88c409f3293ea313b70e730b2a102d7e0497ebe/component/core/src/main/java/io/meeds/social/cms/listener/PageRemovedIndexingListener.java#L43)

```java
public class PageRemovedIndexingListener extends Listener<Object, Page> {
```

`Listener` is an abstract class extending `BaseComponentPlugin` (the Kernel plugin base) — needless inheritance for a Spring `@Component`. Its sibling in the same package, `PageContentBlockIndexingListener`, uses `implements ListenerBase<String, String>`.

**Fix:** `implements ListenerBase<Object, Page>`, for consistency with the sibling.

---

### On portal#1303

One line, in the right place. Because `removePages(SiteKey)` delegates to `remove(PageKey)` which delegates to `remove(Page)`, the single broadcast covers site deletion, navigation-node deletion and direct page removal alike — and it fires *after* `destroyPage` while the `Page` object is still in hand, so `getStorageId()` is available to listeners. `broadcastEvent` passes the service as source and the page as data, which is what `PageRemovedIndexingListener.onEvent` reads. Build and Sonar are green there.

Worth noting it has no test, and it touches the portal core — see the classification note below.

---

CI right now: **social** build green, Sonar red on **F** alone. **portal** fully green. **layout** and **notes** still red purely on merge ordering — layout's four errors are all `PageUrlResolver`/`PageUrlResolverService` and notes' are all `component/api` classes, both awaiting social's snapshot.

This revision closed four of five round-2 findings including both High ones, and the page-deletion solution is better than what I suggested — routing it through the portal's own `remove(Page)` covers paths a Layout-side broadcast would have missed. **F** is a one-character fix that also clears the gate.

**Classification stays N1**, now also because portal#1303 modifies core platform event broadcasting. Note the changeset now spans four repos with a strict merge order: **portal → social → layout/notes**.

mkrout:
## Fixes applied — response to round 3 review

Pushed at `dd1b76c`. Addresses every finding from [round 3](https://github.com/Meeds-io/social/pull/5974#issuecomment-5269172482). No changes needed in `layout`/`notes`/`portal` this round — everything flagged was in this repo.

### F. 🟠 High (Sonar blocker) — literal NUL byte in `blockHash`'s separator — ✅ Fixed

The separator between `contentType` and `settingName` was an actual `0x00` byte typed into the source instead of the escape `'\0'` or a plain `' '`. Replaced with a real ASCII space. Confirmed via byte-level inspection: exactly one `0x00` in the file, now zero; `file` no longer reports it as `data` (binary).

### G. 🟠 High — whole file flipped to CRLF, burying F inside a 690-line diff — ✅ Fixed

`PageContentIndexingConnector.java` was 345/345 lines CRLF (0/345 at the previous clean revision). Converted back to LF byte-for-byte. Same root cause as round 1's XML churn — some tool in the local edit/rebase loop normalizing line endings on save; worth the `.gitattributes` `*.java text eol=lf` suggestion to stop it recurring.

### H. 🟡 Medium — `getAllIds` still resolving every page before paging — ✅ Fixed

Content types and settings are now sorted deterministically (content types via natural `String` order, settings via `CMSSetting::getName`) before `skip`/`limit` is applied to the flattened `(type, setting)` pairs. `buildBlockId` — the call that triggers `layoutService.getPage()` — now only runs on the settings that survive paging, not on every setting in the platform on every batch.

### I. 🟢 Nit — `PageRemovedIndexingListener` used `Listener` instead of `ListenerBase` — ✅ Fixed

Switched to `implements ListenerBase<Object, Page>`, matching its sibling `PageContentBlockIndexingListener`.

### Tests

`component/core`: 1014/1015 — same pre-existing, unrelated `SpaceCategorySidebarPluginTest` flake as every prior round, confirmed again not touched by this diff.

boubaker:
## Review round 4 — eXIP 7.3.0.13 "Page Searching"

Re-reviewed at `b4c634e` (social), `7b09d4b` (layout), `02584c6` (notes), `dbf969b` (portal, unchanged).

Previous rounds: [1](https://github.com/Meeds-io/social/pull/5974#issuecomment-5256465388) · [2](https://github.com/Meeds-io/social/pull/5974#issuecomment-5266617972) · [3](https://github.com/Meeds-io/social/pull/5974#issuecomment-5269172482)

**Every open finding is closed, and social is green on build and Sonar.** Nothing blocking left from my side. Note social's branch has since absorbed a large unrelated merge (MIPs#237 group management / organizational units) — this review covers only the eXIP files.

### Round 3 status

| # | Sev | Finding | Status |
|---|---|---|---|
| F | 🟠 | Raw NUL byte embedded as a character literal; Sonar CRITICAL, red gate | ✅ **Fixed** — separator is now `' '`; Sonar green |
| G | 🟠 | Whole connector flipped to CRLF, hiding the real change | ✅ **Fixed** — file is LF again (427 lines, 0 CR) |
| H | 🟡 | `getAllIds` resolved a page per setting on every batch | ✅ **Fixed** — sorted for deterministic paging, all disqualifying filters evaluated *before* `skip`/`limit`, and a per-call `Map<String, ResolvedPage>` memo. The comment on why a short batch would end `reindexAll` early is a sharper catch than my finding was |
| I | 🟢 | `PageRemovedIndexingListener` extended the legacy `Listener` | ✅ **Fixed** — `implements ListenerBase` |

### Older leftovers

| # | Sev | Finding | Status |
|---|---|---|---|
| R1-12 | 🟢 | `pagePath` frozen at index time, stale after a navigation rename/move | ✅ **Fixed** — `NavigationLayoutService` now broadcasts `PAGE_UPDATED_EVENT` for the whole affected subtree on `updateNode` and `moveNode` |
| R1-14 | 🟢 | `PageSearchResult.siteName` carries the resolved *label*, not the name | ❌ Open (cosmetic) |
| R3-portal | 🟢 | No test on the new `PAGE_REMOVED` broadcast | ❌ Open (portal#1303 unchanged) |

### Changes I didn't ask for, and that are the strongest part of this revision

Several real gaps were found and closed without prompting:

- **The `favorites` query parameter was being silently dropped.** The unified search bar appends it for any connector declaring `favoritesEnabled`, so the spec's *"once bookmarked, it is searchable (combining search + favorite filter)"* simply didn't work. Now filtered server-side, with an early return when the user has no favourites, and — good instinct — `buildFavoriteFilter` matches *nothing* rather than degrading to "no filter at all" when no favourite id is well-shaped.
- **`unindexContentBlock` + `note.deleted`.** The observation behind it is the sharp one: a deleted note leaves its `CMSSetting` behind, so a plain re-index still resolves a document id, and `create()` returning `null` makes the framework *skip* the operation rather than delete the document. The stale document would have lingered forever. That's precisely the failure mode I flagged in round 2 for the wildcard query, found here independently.
- **`PageContentBlockUtils.collectWidgetSettingNames`** closes the "`CMSSetting` outlives its widget" hole at the root, and — critically — applies the same rule in `getAllIds` *and* `create()`, so a full reindex can't resurrect exactly the blocks the incremental listener just removed. The javadoc explains why content availability alone can't distinguish the two cases.
- **`_source` now excludes `content`/`content-*`**, with a new `contentLanguages` keyword field answering the has-a-translation question instead. Search hits stop shipping every language's full page text back just to pick an excerpt field.
- **`PAGE_UNREACHABLE_EVENT` is deliberately *not* `PAGE_REMOVED`**, with the reasoning spelled out: a listener acting on "page data is gone" is entitled to destroy everything attached to the page, which would be wrong for a page that still exists. That distinction is easy to get wrong and was worth making explicit.

I verified the two things those changes depend on: `NavigationService.loadNode(SiteKey)` does use `Scope.ALL`, so the subtree walk really sees descendants; and `CMSPortlet.NAME` is indeed `"name"`, so the duplicated preference literal is correct.

---

### 🟡 Medium — a full navigation-tree load on every node property change

[`NavigationLayoutService.java:207`](https://github.com/Meeds-io/layout/blob/7b09d4bd1739633985dcdec84d2f355340202147/layout-service/src/main/java/io/meeds/layout/service/NavigationLayoutService.java#L207) · [`:463`](https://github.com/Meeds-io/layout/blob/7b09d4bd1739633985dcdec84d2f355340202147/layout-service/src/main/java/io/meeds/layout/service/NavigationLayoutService.java#L463)

```java
    navigationService.updateNode(nodeId, nodeState);
    broadcastPagesUpdated(nodeData.getSiteKey(), nodeId, username);   // unconditional
```

`broadcastPagesUpdated` walks `loadNode(siteKey)` — the whole site tree, `Scope.ALL` — and then re-indexes every page in the node's subtree. `updateNode` fires for *any* node property: label, visibility, start/end date. Only a rename actually changes the page URI, so a visibility toggle on a section now costs a full tree load plus a reindex of every descendant page for no benefit.

**Fix:** compare the node's name (and parent, for `moveNode`) against the incoming model and skip the broadcast when the URI can't have changed.

---

### 🟢 Nit — a few small ones

- [`PageContentIndexingConnector.java:379`](https://github.com/Meeds-io/social/blob/b4c634e7a4b574f0bffe481a3eb2be17fc4ffd57/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentIndexingConnector.java#L379) — the hash separator is now a space, which is unambiguous only because no content type contains one. That holds today (they're plugin-declared constants like `notePage`) but it's an unwritten invariant; `'\0'` written as the escape would have been unconditionally safe, and is what I'd have suggested instead of the raw byte.
- Three string contracts are now duplicated across repos with a "can't depend on that module" comment each: `layout.page.updated` / `layout.page.permissions.updated` (Layout ↔ Social), `layout.page.unreachable` (Layout ↔ Social), and `"name"` (`CMSPortlet.NAME` ↔ `PageContentBlockUtils`). Layout and Notes both already depend on `social-component-api` — a small constants holder there would make these compile-time contracts instead of comments.
- `PageSearchResult.siteName` still holds the resolved label; worth renaming while the DTO is new.

---

CI: **social** and **portal** fully green. **layout** and **notes** remain red purely on merge ordering — layout's four errors are all `PageUrlResolver`/`PageUrlResolverService`, notes' are all `component/api` classes, both awaiting social's snapshot.

Across four rounds this went from two Critical ACL/injection holes and an index that never cleaned up after itself, to a connector that reconciles on layout change, node deletion, page deletion, content deletion and node rename — with the reconciliation rule applied identically in the incremental and full-reindex paths. The remaining items are one Medium efficiency point in Layout and three nits.

**Classification stays N1** (ACL filtering, a trust boundary, and core platform event broadcasting in portal#1303). Merge order: **portal#1303 → social#5974 → layout#562 / notes#1747**.

boubaker:
## Review round 5 — eXIP 7.3.0.13 "Page Searching"

Re-reviewed at `cb0edd6` (social), `fd265f4` (layout); notes `02584c6` and portal `dbf969b` unchanged.

Previous rounds: [1](https://github.com/Meeds-io/social/pull/5974#issuecomment-5256465388) · [2](https://github.com/Meeds-io/social/pull/5974#issuecomment-5266617972) · [3](https://github.com/Meeds-io/social/pull/5974#issuecomment-5269172482) · [4](https://github.com/Meeds-io/social/pull/5974#issuecomment-5289745779)

**All round-4 findings closed.** One new item below, and it's a "verify on a running server", not a defect I can demonstrate.

### Round 4 status

| # | Sev | Finding | Status |
|---|---|---|---|
| 1 | 🟡 | Full navigation-tree load on every node property change | ✅ **Fixed — better than proposed.** Rather than guarding the broadcast, `NavigationUpdateModel` was checked and found to carry **no node name**, so `updateNode` can't move any URI at all; the subtree reindex was simply removed. What replaced it handles a case I hadn't spotted — re-pointing a node at a different page now reindexes the new page and unindexes the previous one if it just lost its last node |
| 2 | 🟢 | `PageSearchResult.siteName` carries the resolved label | ❌ Open (cosmetic) |
| 3 | 🟢 | Cross-repo string contracts duplicated | ❌ Open |
| 4 | 🟢 | portal#1303 has no test | ❌ Open (unchanged) |

### Again, the unprompted fixes are the strongest part

- **`moveNode` looked the moved subtree up in the wrong site.** A cross-site move is allowed — the destination parent's site is what the ACL check runs against — so after the move the node is only findable in the destination tree. The old code used the *source* site key and would have silently left every page in the moved subtree indexed with its pre-move URI. Fixed with a dedicated test asserting the source site is never loaded.
- **`createNode` didn't reindex the page it points at.** This closed a real one-way door: once a page's blocks were unindexed because their last node was deleted, creating a new node for that page would never have brought them back. Draft nodes are correctly skipped.
- **`escape()` didn't strip backslashes.** A term ending in `\` would escape the query template's own closing quote and make the request unparsable — an ES 400 surfacing as a 500. I checked the new pattern: `\Q`-quoting terminates on the *second* backslash's `\E`, so exactly one backslash lands in the character class. Correct.
- **`{"ids": {"values": []}}` → `{"match_none": {}}`** for the empty-favorites case.
- **`findContentBlock` inverted its two checks** — match the hash first, then load only that one widget's preferences instead of every widget's on the page. The reasoning holds: the hash is derived from `contentType + settingName`, so a match identifies that exact block and no other content type could supply it.

---

### 🟡 Medium — the new ACL dependency closes a construction loop that nothing tests

[`PageFavoriteACLPlugin.java:50`](https://github.com/Meeds-io/social/blob/cb0edd61bfc99c6f3980a69774546e99f1c1a4a6/component/core/src/main/java/org/exoplatform/social/core/plugin/PageFavoriteACLPlugin.java#L50) · [`PageContentSearchConnector.java:134`](https://github.com/Meeds-io/social/blob/cb0edd61bfc99c6f3980a69774546e99f1c1a4a6/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentSearchConnector.java#L134)

```java
public PageFavoriteACLPlugin(LayoutService layoutService, UserACL userACL, PageContentSearchConnector searchConnector)
```

The hardening itself is right — checking access on the page alone let any hash be bookmarked under a readable page, producing favourites rows that resolve to nothing on hydration, and denying when the index is unreachable is the safe direction. What's worth a look is the wiring it introduces:

`FavoriteService` → its component-plugin `PageFavoriteACLPlugin` → `PageContentSearchConnector` → whose own constructor takes `FavoriteService`.

That should resolve — Kernel instantiates a component before its plugins, so `FavoriteService` already exists when the connector is built — but nothing here proves it. `PageFavoriteACLPlugin` is declared **only** in `webapp/.../component-plugins-configuration.xml`, never in a test container config, and `PageFavoriteACLPluginTest` constructs the plugin directly with mocks. So no test in this PR boots a container containing both.

**Fix:** no code change needed if it boots — just confirm on a running server that the container starts and that bookmarking a page from a search result still works. If it does cycle, declaring the connector dependency lazily (container lookup on first use, as `CategoryIndexingConnector` does for `CategoryStorage`) breaks it.

---

CI: **social** and **portal** green. **layout** and **notes** red purely on merge ordering — layout's four errors are still `PageUrlResolver`/`PageUrlResolverService`, notes' are `component/api` classes, both awaiting social's snapshot.

Five rounds in, this is in good shape. Nothing I raised is outstanding beyond three cosmetic nits, and for the last three rounds the changes found more than the review did — the wrong-site subtree lookup, the unescaped backslash and the never-reindexed new node were all found here, not by me.

**Classification stays N1.** Merge order: **portal#1303 → social#5974 → layout#562 / notes#1747**.

component/core/src/main/java/io/meeds/social/cms/listener/PageContentBlockIndexingListener.java
```diff
@@ -0,0 +1,157 @@
...
+ * currently carries a content block backed by any registered
+ * {@link io.meeds.social.cms.plugin.PageContentBlockPlugin}.
+ * <p>
+ * A page that is deleted outright never fires either event — that case is
+ * covered separately by {@link PageRemovedIndexingListener}, since by the
+ * time it would fire, {@link LayoutService#getPage} can no longer resolve
+ * the page. This listener only detects a block detached from a page that
+ * still exists.
+ */
+@Component
```
boubaker, line 58: please make this `Asynchronous` and `ContainerTransactional` to improve editing performances

component/core/src/main/java/io/meeds/social/cms/listener/PageRemovedIndexingListener.java
```diff
@@ -0,0 +1,92 @@
...
+ * their broadcasters:
+ * <ul>
+ * <li>{@link LayoutService#PAGE_REMOVED} — the page's own data is gone.</li>
+ * <li>{@value #PAGE_UNREACHABLE_EVENT} — the page still exists but no
+ * navigation node leads to it anymore (Layout broadcasts it after deleting
+ * the last node pointing to it), so it can't be reached from a search result
+ * either.</li>
+ * </ul>
+ */
+@Component
```
boubaker, line 54: please make this `Asynchronous` and `ContainerTransactional` to make the operation non-blocker since not pertinent to be sync

component/core/src/main/java/org/exoplatform/social/core/plugin/PageFavoriteACLPlugin.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+    Page page = layoutService.getPage(pageId);
+    if (page == null || !userACL.hasAccessPermission(page, userIdentity)) {
+      return false;
+    }
+    // A favorite is stored per content block, not per page: the page id is only
+    // the leading portion of the id, so checking access on it alone would let
+    // any hash be bookmarked under a readable page — rows that resolve to
+    // nothing when the favorites list tries to hydrate them.
+    return isIndexedBlockOfPage(page.getStorageId(), objectId);
```
boubaker: This operation performance is very expensive, is it a way to not have it here ? IMO, we should be able to add a page as favorite without having it indexed, isn't it ?

component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentIndexingConnector.java
```diff
@@ -0,0 +1,432 @@
...
+ * View blocks placed in different sections of the same page); each one is
+ * indexed as its own document, so that unified search returns one result
+ * per block rather than a single blended excerpt for the whole page.
+ * <p>
+ * The document id is {@code <page's numeric storage id>_<block hash>} (not
+ * {@link PageKey#format()}, which can exceed the 50-character limit of the
+ * {@code ES_INDEXING_QUEUE.ENTITY_ID} column, nor the setting's own name,
+ * which can be arbitrarily long) — the page id portion stays stable across
+ * page/site renames, unlike the formatted key.
+ */
```
boubaker, line 71: Just a note, nothing else for now. We should have used Spring ES repositories instead of legacy way to request ES data (Manually built queries and indexation).

webapp/src/main/webapp/WEB-INF/conf/social-extension/social/search-configuration.xml
```diff
@@ -63,6 +63,16 @@
     </init-params>
   </component>
 
+  <component>
+    <type>io.meeds.social.cms.storage.elasticsearch.PageContentSearchConnector</type>
```
boubaker, line 67: Just a note: we want to get rid of XML defined beans, you should add this as a `@Service` with `@Value` for properties.

webapp/src/main/webapp/WEB-INF/conf/social-extension/social/component-plugins-configuration.xml
```diff
@@ -158,6 +158,15 @@
     </component-plugin>
   </external-component-plugins>
 
+  <external-component-plugins>
+    <target-component>org.exoplatform.social.metadata.favorite.FavoriteService</target-component>
```
boubaker, line 162: Just a note: we want to get rid of XML defined beans, you should add this as a `@Component` and inject the plugin using the Service method in `@PostConstrcut`.

boubaker (approved):

PR #5976, opened 2026-08-13
Merge App Badges Notification

boubaker:


boubaker:
## AI pre-review — App Badges Notification (eXIP 48140)

Reviewed against the technical specification (note 48140) at head `69e97d14`. **No findings.**

Verified against the spec's pin-restriction requirements:
- Both hunks (`search.jsp:80-83`, `topbarNotification.jsp:26-29`) use the defensive create-or-append idiom (`eXo.env.portal.topbarDisplayedApps = eXo.env.portal.topbarDisplayedApps || []`), so JSP execution order is irrelevant and neither writer needs to know the other exists.
- The `includes()` guard makes a double render idempotent.
- The scripts are plain inline `<script>` blocks (no `defer`/`module`), so they execute during HTML parsing, strictly before any AMD-mounted Vue application — the ordering guarantee the spec's design relies on.
- Being inside the portlet JSPs, they run only when the portlet actually renders: an excluded or disabled topbar item never registers, leaving its pin available — exactly the "reflects what is actually displayed" rule.
- The registered keys match the App Center binding contract: `applications.json` `"url": "search"` and `"url": "notifications"` in app-center.

Eight lines, zero dependencies — exactly the below-App-Center declaration channel the spec required.

**Classification:** additive frontend (N3-shaped) but riding in the N1 eXIP 48140 delivery — merge together with the rest after Architect validation.

sofyenne (approved):

PR #6006, opened 2026-08-21
Merge mips exip7.3.0.16

AzmiTouil:


webapp/src/main/webapp/vue-apps/common/js/ReactionService.js
```diff
@@ -0,0 +1,86 @@
...
+  return reactionOptions && filterReactionOptions(reactionOptions, objectType) || null;
+}
+
+function filterReactionOptions(options, objectType) {
+  return objectType && options.filter(option => !option.objectTypes?.length || option.objectTypes.includes(objectType)) || options;
+}
+
+export function getReactions(objectType, objectId, reactionId, offset, limit) {
+  const reactionFilter = reactionId && `&reactionId=${reactionId}` || '';
+  return fetch(`/social/rest/reactions/${objectType}/${objectId}?offset=${offset || 0}&limit=${limit || 0}${reactionFilter}`, {
```
boubaker: 🟠 **High — the drawer filter by the `#️⃣` keycap reaction always returns an empty list**

```js
const reactionFilter = reactionId && `&reactionId=${reactionId}` || '';
```

`reactionId` is interpolated raw, and a custom reaction id *is* the emoji character. `#️⃣` (`0023-FE0F-20E3`) ships in `emojiBank.json` and is accepted by the server — `isCustomEmojiReaction` reaches `isKeycapBase('#')` because U+20E3 is present. Its literal `#` starts the URL fragment, so the server receives `reactionId=""`, and `getReactionItemsByOption("")` matches no metadata name. The chip renders a count (those come from the grouped DB query) next to an empty reactor list.

**Fix:** `encodeURIComponent(reactionId)` in the filter, and wrap `objectType`/`objectId` the same way.

webapp/src/main/webapp/vue-apps/activity-reactions/components/ActivityReactionsList.vue
```diff
@@ -0,0 +1,229 @@
...
+        detail: {
+          numberOfReactions: this.totalCount,
+          type: 'like'
+        }
+      }));
+    },
+    loadMore() {
+      this.limit = this.limit + 20;
+      this.retrieveReactors();
+    }
```
boubaker, line 247: 🟠 **High — beyond 100 reactors, the drawer shows other people's reactions as plain 👍**

```js
loadMore() {
  this.limit = this.limit + 20;   // unbounded, and refetches from offset 0
  this.retrieveReactors();
},
```

`retrieveAllReactors` passes this same growing `limit` to two endpoints with different ceilings: `$activityService.getActivityLikers` is uncapped, while `ReactionRest.getReactions` clamps to `MAX_LIMIT = 100`. From the 6th "Load more" on, likers 101+ arrive with no entry in `reactionIdsByReactor` and fall through to `'like'` at line 88 (`this.reactionIdsByReactor[liker.id] || 'like'`) — someone who reacted ❤️ is displayed with 👍. `hasMoreReactors` (`selectedCount > this.limit`) also stays true indefinitely, so the button keeps offering a page that adds nothing.

**Fix:** clamp `this.limit` to 100 client-side, or switch to `offset`-based paging that appends (`offset = this.likers.length`) instead of refetching from 0 with a growing limit.

boubaker (commented):
## AI review — Round #1

eXip **"React to post using emojis"** — single-repo delivery on `social`, targeting `feature/mips`.
Reviewed at head `71b3c01` · 41 files · +3165 / −211 · verified against the Tech Spec (community activity 475149).

**7 findings inline** — 3 High, 4 Medium. Nits, verified-conform notes and the classification below.

---

### 🟢 Nits

- **`ReactionChooser.vue:160`** — `popoverWidth = this.options.length * 40 + 8` omits the "+" button (36px + 4px margin), so the right-edge clamp under-measures by 40px and the popover can still overflow near the viewport edge. → `(this.options.length + 1) * 40 + 8`.
- **`Webui_en.properties:136`** — `UIActivity.reaction.active.applause=Applausing` isn't a word; "Applauding". Worth fixing before Crowdin propagates it to ~40 locales.
- **`metadata-plugins-configuration.xml:327`** — `<long>-1122997398</long>` is `"reactions".hashCode()`; the neighbouring types use plain literals (`54175`, `579645`), so a bare negative magic number reads like a typo. One comment line fixes it.
- **`ReactionServiceImpl.java:212,220`** — fully-qualified `java.util.Arrays.stream(...)` twice, while the rest of `java.util` is imported normally.
- **`MetadataServiceImpl.java:486`** — `validateAndGetMetadataType(metadataTypeName)`'s result is discarded and the *name* passed down, so `MetadataStorage` resolves the type a second time; every sibling method passes `metadataType.getId()`.
- **`EmojiPickerQuickEmojis.vue:23-33`** — the new `emojis` prop has no consumer in this PR (the chooser renders its own row). Backward-compatible, but it is added surface on a component Matrix also consumes, with no user.
- **`.reaction-emoji`** is used in three components (`ActivityReactionsList.vue`, `ReactionChooser.vue`, `ActivityLikeAction.vue`) but defined **nowhere** — not in social's skin, not in platform-ui. `ActivityLikeAction.vue` compensates with an inline `:style` for font-size/height/line-height, exactly the anti-pattern the centralized-skin rule targets. Either define it once (carrying the sizing) or drop it.
- **`capReactionItems`** carries the pre-existing `authentiatedUserId` typo into a new signature and javadoc — the one place it could have been left behind.

---

### ✅ Verified conform

Checked and correct, so reviewers don't re-litigate:

- **ACL.** `getActivityWithCheck` resolves the viewer via `IdentityRegistry.getIdentity(username)` and gates on `activityManager.isActivityViewable(activity, viewer)` — the same check the legacy like resource uses. The reactor identity always comes from `request.getRemoteUser()`, never from the path or body. All three endpoints carry `@Secured("users")`, and the controller performs no ACL of its own.
- **Scheduled-activity gate.** `getPublicationStartTime() != null` is the right idiom: the publication named query sets the column to `NULL` (`ActivityEntity.java:73`) and `ActivityESListener.java:95` tests it identically — so previously-scheduled, now-published activities stay reactable.
- **Trust boundary.** An ASCII non-registered id cannot pass `isCustomEmojiReaction` (the keycap bases are only reachable when U+20E3 is present), so free text and HTML never reach `Metadata.name`. The encoded name is bounded at ~102 chars against `NVARCHAR(255)`, and `decodeReactionId` catches both `NumberFormatException` and `IllegalArgumentException`. The **16-char cap rejects 0 of the 2066 bank entries**, 4-person ZWJ families included — I checked all of them.
- **Empty `IN (:creatorIds)`.** Guarded in `MetadataServiceImpl` (`CollectionUtils.isEmpty → emptyList`), so the new JPQL never executes with an empty collection — and the path *is* reachable, since `pagedLikerIds` is empty on a zero-liker activity.
- **No cache corruption from the decode.** `retrieveMetadataItems` maps to *fresh* `MetadataItemEntity` instances before `setName` mutates them; the cached `MetadataItem` list is untouched.
- **Write ordering matches §4.** Like before typed item in `setReaction`; item before unlike in `deleteReaction`, so `ReactionLikeDeletedListener` finds nothing and cannot double-broadcast. `saveLike` is skipped when the caller is already a liker, so changing a reaction fires no notification / gamification / stream lifecycle — as promised.
- **Upgradability claim holds.** No Liquibase changeset, no new `@Entity`, no `jpa-entities.idx` line — the "zero migration" design is real.
- **i18n.** All 20 keys `DefaultReactionOptionsPlugin` derives (`UIActivity.reaction[.selected|.active].<id>` × 6, plus `selectAnother` and `selected.custom`) are present; `activity.reactions.seeMore` already exists in every locale.
- **Wiring.** Drawer extension contract preserved (`id: 'like'`), `activity-likes-list` removed with no dangling reference anywhere, `$reactionService` registered in `common/main.js`, `activityReactions → emojiPicker` declared in `gatein-resources.xml`, and all three new test classes registered in `NoContainerTestSuite` — they actually run.
- **Emoji picker contract.** `EmojiPickerList` emits the emoji *character* via `getEmojiChar`, which is exactly what `selectCustomEmoji` assumes; the chat default list is unchanged.
- **Colour/utility classes are platform-ui helpers, not new CSS.** `light-grey-background`, `grey-lighten1-background`, `icon-default-color`, `baseline-vertical-align` and `overflow-x-auto` all resolve in `platform-ui/core/helpers.less` — no new colour class, no new skin variable, platform-ui genuinely untouched on that axis (see the Medium finding on `Style.less` for the motion axis).

---

### What this PR does well

The load-bearing decision — a reaction *is* a like, with the type as a metadata decoration — is the reason this ships with no Liquibase changeset, no migration, and every downstream consumer (notifications, gamification, analytics, websocket, counts) untouched; the rejected-alternatives section earns that choice rather than asserting it. Storing the id in `Metadata.name` and ASCII-encoding custom emoji keeps counts and filters as single indexed queries. And the test suite is unusually honest for a feature this size: 47 unit tests covering the awkward paths — stale typed items clamped, broadcast failure not breaking the write, keycap emoji accepted, over-long sequences rejected, legacy-unlike cleanup, and the like-before-item ordering asserted explicitly.

---

### Classification: N1

The deciding surfaces are `getActivityWithCheck` (all reaction ACL, newly reachable from three new endpoints) and `isCustomEmojiReaction` (the only barrier between a client-supplied string and a persisted metadata name). Max-severity aggregation makes the whole PR N1 regardless of the ~1800 lines of Vue.

**This PR must be validated by an Architect/Senior Developer before merging into `feature/mips` — please make sure it isn't auto-approved on AI review alone.** Author ≠ approver: @AzmiTouil authored it and cannot be the approver.

webapp/src/main/webapp/vue-apps/activity-stream/components/activity/footer/actions/ActivityLikeAction.vue
```diff
@@ -82,38 +140,79 @@ export default {
...
+      this.morphEnabled = true;
+      return this.applyReaction(option.id);
     },
-    likeActivity() {
+    applyReaction(reactionId) {
       this.changingLike = true;
-      return this.$activityService.likeActivity(this.activityId)
-        .then(data => {
+      return this.$reactionService.setReaction('activity', this.activityId, reactionId)
+        .then(() => {
```
boubaker, line 175: 🟠 **High — the reaction write path has no `.catch`, and the emoji bank offers 11 emoji the server rejects**

`setReaction` throws on any non-2xx, but this chain is `.then(…).finally(…)` with no `.catch`.

Running `isCustomEmojiReaction`'s logic over all 2066 entries of the shipped `emojiBank.json`, **11 are rejected with 400**: the GB subdivision flags 🏴󠁧󠁢󠁥󠁮󠁧󠁿 🏴󠁧󠁢󠁳󠁣󠁴󠁿 🏴󠁧󠁢󠁷󠁬󠁳󠁿 (tag characters U+E0067… fall outside every whitelisted range), U+1FBC5–1FBC9 (🯅🯆🯇🯈🯉), and ℗ ℠ ⊜. The user picks one from the "+" bank, the spinner stops, nothing changes, and the only trace is an unhandled rejection in the console. The same silence covers a 403 on a scheduled activity and any transient 500.

The Tech Spec's assumption ledger accepts that fringe symbols are rejected — it does not accept that the rejection is invisible.

**Fix:** add `.catch` here and on `unlikeActivity` (plus the `ActivityCommentLikeAction.vue` equivalents) surfacing a message; then either extend `isEmojiBase` with U+1FB00–1FBFF and treat U+E0020–E007F as modifiers, or filter those 11 entries out of the bank so the UI never offers what the server refuses.

webapp/src/main/webapp/vue-apps/activity-reactions/components/ActivityReactionsList.vue
```diff
@@ -0,0 +1,229 @@
...
+          this.counts = reactionsData.counts || {};
+          const reactions = reactionsData.reactions || [];
+          const reactionIdsByReactor = {};
+          reactions.forEach(reaction => reactionIdsByReactor[`${reaction.reactorIdentityId}`] = reaction.reactionId);
+          this.reactionIdsByReactor = reactionIdsByReactor;
+          return Promise.all(reactions.map(reaction =>
+            this.$identityService.getIdentityById(reaction.reactorIdentityId)
+              .then(identity => Object.assign({}, identity?.profile, {id: `${reaction.reactorIdentityId}`}))
+              .catch(() => null)));
+        }).then(reactors => {
```
boubaker: 🟡 **Medium — the chip filter resolves reactor profiles one request per reactor (N+1)**

```js
return Promise.all(reactions.map(reaction =>
  this.$identityService.getIdentityById(reaction.reactorIdentityId)
```

Clicking a chip fires up to `limit` (≤100, growing with "Load more") parallel identity requests. The sibling unfiltered path in this same component — `retrieveAllReactors` — gets every profile in a *single* `getActivityLikers` call, so the correct shape is a few lines above.

This is precisely the N+1 the Tech Spec flagged as this endpoint's N2 risk. The DB side was addressed (creator-targeted queries, chunked by 500); the client side was not.

**Fix:** return reactor profiles in the `ReactionList` envelope (one server-side identity pass), or fetch them through a batch identity endpoint.

component/core/src/main/java/io/meeds/social/reaction/service/ReactionServiceImpl.java
```diff
@@ -0,0 +1,361 @@
...
+  @Autowired(required = false)
+  private List<ReactionOptionPlugin> reactionOptionPlugins = new ArrayList<>();
+
+  /**
+   * Allows registering options from another Spring context (cross-addon), in
+   * addition to same-context collection injection. A later registration
+   * overrides an already registered option with the same id (edit).
+   */
+  public void addPlugin(ReactionOptionPlugin reactionOptionPlugin) {
+    reactionOptionPlugins.add(reactionOptionPlugin);
```
boubaker, line 97: 🟡 **Medium — `reactionOptionPlugins` is a plain `ArrayList` mutated after publication**

```java
@Autowired(required = false)
private List<ReactionOptionPlugin> reactionOptionPlugins = new ArrayList<>();

public void addPlugin(ReactionOptionPlugin reactionOptionPlugin) {
  reactionOptionPlugins.add(reactionOptionPlugin);
}
```

`addPlugin` is the documented cross-addon registration entry, so it is called from another WAR's `@PostConstruct` — i.e. while the platform is already serving requests. `getReactionOptionsById()` streams this same list on every `GET /options` and every `setReaction`; a concurrent `addPlugin` gives a `ConcurrentModificationException` → 500.

It also silently depends on Spring's collection injection handing back a *mutable* list. The day it doesn't, the contributing addon fails to boot on `UnsupportedOperationException` — inside its own `@PostConstruct`, which is an awkward place to debug.

**Fix:** `CopyOnWriteArrayList`, seeded from the injected collection (setter injection or a `@PostConstruct` copy) rather than mutated in place.

webapp/src/main/webapp/skin/less/portlet/ActivityStream/Style.less
```diff
@@ -51,6 +51,51 @@
...
+  .reaction-chooser .reaction-selected-pop,
+  .reaction-morph-enter-active {
+    animation: none;
+  }
+  .reaction-chooser .reaction-others-collapse,
+  .reaction-morph-leave-to {
+    transform: none;
+  }
+}
+
```
boubaker: 🟡 **Medium — the reaction framework's CSS is bound to one portlet, so it doesn't travel with the framework**

These 45 lines (keyframes `reaction-pop`, `.reaction-chooser`, the `reaction-morph-*` transition classes, the `prefers-reduced-motion` guard) live in the ActivityStream portlet skin, which `gatein-resources.xml:74-88` binds via `<portlet-skin>` to `SpaceActivityStreamPortlet` / `UserActivityStreamPortlet`. But the JS module is `activityReactions`, injected through `extensionRegistry`, and the chooser popover is detached to the application root by `v-menu`.

The Tech Spec's Extensibility section promises the framework is reusable for notes/tasks/events "with no core modification — only a binding in the consuming addon". On a notes page no ActivityStream skin is loaded, so the chooser would render with no pop animation, no collapse transition, and **no `prefers-reduced-motion` guard** — the accessibility commitment of §2 silently stops applying.

Against `frontend-vue.md` ("Centralized skin — no per-portlet custom CSS; build on platform-ui `helpers.less` / `vuetify-all.less`; Vuetify 2 for the rest"), three reuse points, verified in `platform-ui@56af4d5f`:

- `reaction-morph-*` re-implements Vuetify 2's shipped `scale-transition` (`vuetify/lib/vuetify.less` defines `.scale-transition-enter/-enter-active/-leave/-leave-active/-leave-to`). `ReactionChooser.vue` already uses `transition="fade-transition"` on the v-menu, so the mechanism is known — keeping only the overshoot keyframe on top of `scale-transition` removes ~14 of these 45 lines.
- `overflow: visible !important; contain: none` overrides `.v-menu__content` (`vuetify.less:19386` — `overflow-y: auto; overflow-x: hidden; contain: content`). The override is legitimate (the `scale(1.5)` overshoot must escape the menu box), but "detached popover whose animation overshoots its box" is a platform-wide concern deserving a named helper, not an ActivityStream detail.
- `helpers.less:1341` already carries `@keyframes fadePulseScale`. **Not a duplicate** — that one is a looping opacity pulse, `reaction-pop` is a one-shot overshoot — but it is the precedent that generic motion keyframes belong in `helpers.less`.

Worth stating fairly: every colour/utility class the components use *is* a platform-ui helper (`light-grey-background`, `grey-lighten1-background`, `icon-default-color`, `baseline-vertical-align`, `overflow-x-auto` all resolve in `core/helpers.less`), so §2's "no new colour classes, platform-ui untouched" holds **on the colour axis**. The gap is that §5 rejected touching platform-ui to stay single-repo, and that argument was made entirely about *colour* helpers — which emojis genuinely eliminated. **Motion** helpers are a case the spec never examined.

**Fix (Architect's call):** either (1) promote `reaction-pop`, the `.v-menu__content` overflow-escape and the reduced-motion block into `platform-ui/core/helpers.less` — norm-correct, but a two-repo delivery with platform-ui released first; or (2) keep them in `social` in their own skin file that follows the `activityReactions` module rather than the ActivityStream portlet, and drop `reaction-morph-*` for Vuetify's `scale-transition`. I'd take (2) for this delivery plus a follow-up to promote the motion helpers when the second consumer lands — it honours the single-repo promise without nailing the framework's presentation to one portlet.

component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
```diff
@@ -1391,6 +1399,29 @@ public static Map<String, List<MetadataItemEntity>> retrieveMetadataItems(ExoSoc
...
+    reactionItems.stream()
+                 .filter(item -> item.getCreatorId() == authentiatedUserId)
+                 .findFirst()
+                 .ifPresent(cappedItems::add);
+    reactionItems.stream()
+                 .filter(item -> item.getCreatorId() != authentiatedUserId)
+                 .limit(REACTIONS_PUBLISHED_LIMIT - (long) cappedItems.size())
+                 .forEach(cappedItems::add);
+    return cappedItems;
+  }
```
boubaker: 🟡 **Medium — `capReactionItems` is untested, and the like button depends on it**

`ActivityLikeAction.userReactionId` derives the button's emoji from `activity.metadatas.reactions` and relies entirely on the invariant this method implements: *the viewer's own item always survives the cap*. If that breaks, every user on a >100-reactor post sees 👍 instead of their own ❤️ — and nothing would catch it.

`ReactionServiceImplTest` covers 30 scenarios and `ReactionStorageTest` 13, but nothing in `component/service` covers this method — the feature's only `EntityBuilder` seam, and the one place the payload contract is enforced.

**Fix:** a unit test building >100 `MetadataItemEntity` with the viewer's own item **last** in the list, asserting the result is exactly 100 items and still contains it.

component/core/src/main/java/io/meeds/social/reaction/service/ReactionServiceImpl.java
```diff
@@ -0,0 +1,372 @@
...
+    }
+    return hasEmojiBase;
+  }
+
+  private boolean isEmojiModifier(int codePoint) {
+    return codePoint == 0x200D
+           || codePoint == 0xFE0E
+           || codePoint == 0xFE0F
+           || codePoint == 0x20E3
+           || (codePoint >= 0xE0020 && codePoint <= 0xE007F);
```
boubaker: 🟡 **Medium — widening the modifier range to tag characters lets one user mint unbounded distinct reaction ids that all render identically**

```java
|| (codePoint >= 0xE0020 && codePoint <= 0xE007F);
```

This correctly unblocks the three GB subdivision flags (finding #3 — thank you). But tag characters are **invisible**, and they are now accepted as free modifiers anywhere in the sequence, in any number, as long as one real emoji base is present. With `CUSTOM_REACTION_MAX_LENGTH = 16` UTF-16 units, a 2-unit base leaves **7 free tag slots over a 96-character alphabet**:

```
🔥 + U+E0020 U+E0021 U+E0022 U+E0023 U+E0024 U+E0025 U+E0026   → 16 units, accepted
   encoded name: emoji_1f525_e0020_e0021_e0022_e0023_e0024_e0025_e0026  (53 chars, persists fine)
```

That's ~96^7 ≈ 7.5 × 10^13 accepted ids that all **display as plain 🔥**. Each distinct one creates a fresh `SOC_METADATAS` row on first use and its own chip in the drawer, so an authenticated user hitting `PUT /reactions/activity/{id}` in a loop grows the metadata table without bound and pollutes the drawer with visually identical chips. Before this commit tag chars were rejected, and the other modifiers (ZWJ, VS15/16, U+20E3) only ever yielded a handful of combinations — so the Tech Spec's "storage growth bounded by the emoji vocabulary" invariant held. It no longer does.

**Fix:** accept tag characters only in the shape they actually exist for — a `U+1F3F4` base followed by a tag sequence terminated by `U+E007F` — or, simpler and tighter, drop the range again and allowlist the three flag sequences explicitly. Either keeps the bounded-vocabulary guarantee that the ledger's storage claim rests on.

webapp/src/main/webapp/vue-apps/activity-stream/components/activity/footer/actions/ActivityLikeAction.vue
```diff
@@ -82,38 +140,81 @@ export default {
...
           this.activity.hasLiked = 'true';
-          this.computeLikes(data);
-          this.$root.$emit('activity-liked', this.activityId);
+          this.updateLocalReactionItem(reactionId);
+          return this.refreshLikers();
         })
+        .catch(() => this.$root.$emit('alert-message', this.$t('UIActivity.reaction.error'), 'error'))
```
boubaker, line 180: 🟢 **Nit — `.catch` also fires when the reaction succeeded but the liker refresh failed**

```js
.catch(() => this.$root.$emit('alert-message', this.$t('UIActivity.reaction.error'), 'error'))
```

The chain is `.then(… return this.refreshLikers())` → `.catch(…)`, so a failure inside `refreshLikers()` (a `getActivityLikers` hiccup) surfaces "Unable to save your reaction, please try again" even though the `PUT` succeeded and the reaction is persisted. The user then retries a write that already landed.

Same shape in `unlikeActivity` (line 191) and in both `ActivityCommentLikeAction.vue` handlers.

**Fix:** attach the `.catch` to the `setReaction` call itself and chain the local-state update after it, or let `refreshLikers()` swallow its own error (it is a best-effort refresh, not part of the write).

component/service/src/main/java/io/meeds/social/reaction/rest/ReactionRest.java
```diff
@@ -0,0 +1,206 @@
...
+  private ReactionService   reactionService;
+
+  @Autowired
+  private IdentityManager   identityManager;
+
+  private List<ProfileEntity> buildReactorProfiles(List<Reaction> reactions) {
+    return reactions.stream()
+                    .map(Reaction::getReactorIdentityId)
+                    .distinct()
+                    .map(identityId -> identityManager.getIdentity(String.valueOf(identityId)))
```
boubaker: 🟢 **Nit — new call to an API marked `forRemoval`**

```java
.map(identityId -> identityManager.getIdentity(String.valueOf(identityId)))
```

`IdentityManager.getIdentity(String)` is `@Deprecated(forRemoval = true, since = "7.0")` — "use `getIdentity(Long)` since the id is no more a UUID inherited from JCR". `Reaction::getReactorIdentityId` already returns a `long`, so `String.valueOf(...)` is converting *away* from the overload that survives the removal.

To be fair: `getIdentity(long)` currently just delegates to the deprecated one, so this is about which overload new code names, not behaviour.

**Fix:** `.map(identityManager::getIdentity)` — the `long` overload binds directly.

webapp/src/main/webapp/skin/less/portlet/ActivityReactions/Style.less
```diff
@@ -0,0 +1,56 @@
...
+  }
+  .reaction-chooser .reaction-others-collapse {
+    transform: none;
+  }
+  .v-btn .scale-transition-enter-active,
+  .v-btn .scale-transition-leave-active {
+    transition: opacity 0.2s ease !important;
+  }
+  .v-btn .scale-transition-enter,
+  .v-btn .scale-transition-leave-to {
```
boubaker: 🟢 **Nit — the reduced-motion override reaches every `.v-btn` on every page**

```less
.v-btn .scale-transition-enter-active,
.v-btn .scale-transition-leave-active {
  transition: opacity 0.2s ease !important;
}
```

`scale-transition` is a shared Vuetify transition, and this file is now declared as a `<portal-skin>` — so it loads on **every** portal page and these `!important` rules apply to any addon's `v-scale-transition` nested in a `v-btn`, not just the reaction buttons.

Under `prefers-reduced-motion` that is arguably the behaviour you'd want anyway, so this is not a defect — but it is a cross-addon reach from a social skin, and it will be invisible to whoever debugs it later.

**Fix:** scope the selectors to the reaction context (a wrapper class on the like button, or `.reaction-chooser-wrapper`) so the blast radius matches the feature.

boubaker (commented):

Re-reviewed at head `9eacc03` (`fix: Harden reactions after integration review round 1 - EXO-89289`), scoped to `71b3c01..9eacc03` plus verification of every Round #1 finding **in the source**, not from the commit message.

### Status of Round #1 findings

| # | Finding | Status |
|---|---|---|
| 1 | 🟠 `#️⃣` chip filter returned an empty list (unencoded `#`) | ✅ **Fixed** — `encodeURIComponent` on `reactionId`, `objectType` and `objectId` across all four `ReactionService.js` calls |
| 2 | 🟠 Reactors past the 100 cap shown as plain 👍 | ✅ **Fixed** — `loadMore` now appends with `offset = likers.length`, `fetchReactionPages` chunks by 100, and `hasMoreReactors` compares against `likers.length` so it terminates. Both endpoints now receive the same window |
| 3 | 🟠 No `.catch` on the write path + 11 bank emoji rejected | ✅ **Fixed** — `.catch` on all four handlers with a new `UIActivity.reaction.error` key; whitelist widened (`0x1FBFF`, tag chars, ℗ ℠ ⊜). **Re-ran the bank scan: 0 of 2066 entries now rejected** (was 11). See the new Medium finding on how the tag range was widened |
| 4 | 🟡 Chip filter did one identity request per reactor | ✅ **Fixed** — `ReactionList.reactorProfiles` built server-side and correlated by id, not index. Traced `ProfileEntity.id`: `buildEntityProfile` uses `profile.getId()`, and `RDBMSIdentityStorageImpl:496` sets it from the `IdentityEntity` id — so it *does* equal `reactorIdentityId` and the lookup is sound. Residual: still one `getIdentity` per reactor, but in-process and behind `CachedIdentityStorage` |
| 5 | 🟡 `reactionOptionPlugins` mutated after publication | ✅ **Fixed** — `final CopyOnWriteArrayList` + `@Autowired(required = false)` setter with a null guard. Both the CME and the immutable-injected-list fragility are gone |
| 6 | 🟡 Reaction CSS bound to the ActivityStream portlet | ✅ **Fixed** — own `portlet/ActivityReactions/Style.less`, declared as a `<portal-skin>` (so it is present wherever the framework is injected) and registered in `webapp/pom.xml`; `reaction-morph-*` replaced by Vuetify's `scale-transition`, reduced-motion guard retained and extended to cover it. Residual: now global rather than scoped — see the new nit |
| 7 | 🟡 `capReactionItems` untested | ✅ **Fixed** — `EntityBuilderReactionsTest` with three cases including the viewer's item appended **last** past the limit, plus the `assertSame` fast path; registered in `component/service`'s `NoContainerTestSuite`, so it runs |

**All seven substantive findings are fixed and verified at head.** The eight 🟢 nits from Round #1 were not taken (`Applausing`, `popoverWidth` missing the "+" button, the `hashCode()` comment, `java.util.Arrays.stream`, the double type resolution in `MetadataServiceImpl`, the unused `emojis` prop, the undefined `.reaction-emoji` class, the `authentiatedUserId` typo) — a fair call for an integration PR, and they stay non-blocking. `.reaction-emoji` is the one worth a second look: it is referenced in four places and still defined nowhere, so the inline `:style` on the like button is doing that job alone.

### New findings

Per the org model, fix commits re-enter classification — the hardening commit touched the trust-boundary validation itself, so it got a full pass rather than a diff skim. One Medium and three nits, inline.

The Medium one matters: unblocking the subdivision flags was the right call, but the way the range was opened also opened an unbounded-vocabulary hole that the spec's storage guarantee depends on being closed.

### Also verified in this round

- The new negative test `setReaction(…, "󠁧󠁿", …)` → `IllegalArgumentException` proves `hasEmojiBase` still guards, so a bare tag sequence with no emoji base is rejected. Exactly the right test for the widening.
- `ReactionRest` autowiring `IdentityManager` to decorate a DTO is norm-compliant (`backend-spring.md` §5 permits reaching another domain's Service for decoration).
- `capReactionItems` going `private` → package-private is a reasonable testability trade; the test sits in the same package.
- `fetchReactionPages` returns `counts` from `pages[0]` only — correct, since `countReactionsByOption` is not paged and every page carries the full map.
- `getIdentity(String)` delegates to `getIdentity(id, true)`, so the profile *is* loaded and the `identity.getProfile() != null` filter won't silently empty the list.

### Classification: N1 (unchanged)

The hardening commit modifies `isEmojiModifier` / `isEmojiBase` — the trust boundary itself — so it re-enters classification and stays N1; it does not inherit Round #1's review. **Still requires validation by an Architect/Senior Developer before merging into `feature/mips`; not to be auto-merged on AI review alone.** Author ≠ approver: @AzmiTouil authored both commits.

webapp/src/main/webapp/vue-apps/common/js/ReactionService.js
boubaker: ✅ **Fixed** in `9eacc03`, verified. `encodeURIComponent` now wraps `reactionId`, `objectType` and `objectId` in all four calls — `#️⃣` serialises as `%23%EF%B8%8F%E2%83%A3` and reaches the server intact.

webapp/src/main/webapp/vue-apps/activity-reactions/components/ActivityReactionsList.vue
boubaker: ✅ **Fixed** in `9eacc03`, verified. `loadMore` appends with `offset = likers.length` instead of growing `limit`, `fetchReactionPages` chunks requests by 100, and `hasMoreReactors` now compares against `likers.length` so it terminates. Both endpoints receive the same window, so the `reactionIdsByReactor` desync is gone.

webapp/src/main/webapp/vue-apps/activity-stream/components/activity/footer/actions/ActivityLikeAction.vue
boubaker: ✅ **Fixed** in `9eacc03`, verified on both halves. `.catch` added to all four handlers with a new `UIActivity.reaction.error` key, and the whitelist was widened (`0x1FBFF`, tag characters, ℗ ℠ ⊜) — I re-ran the bank scan and **0 of 2066 entries are rejected now**, down from 11. One caveat on how the tag range was opened: see the new Medium finding on `ReactionServiceImpl.java:304`. Minor follow-up on this `.catch` itself in Round #2 as well.

webapp/src/main/webapp/vue-apps/activity-reactions/components/ActivityReactionsList.vue
boubaker: ✅ **Fixed** in `9eacc03`, verified. Profiles are now built server-side into `ReactionList.reactorProfiles` and correlated **by id**, not by index, so the de-dup/filter in `buildReactorProfiles` cannot desync the lists. I traced the id equality rather than assuming it: `buildEntityProfile` does `new ProfileEntity(profile.getId())`, and `RDBMSIdentityStorageImpl:496` sets `profile.setId(String.valueOf(entity.getId()))` from the `IdentityEntity` — so it genuinely equals `reactorIdentityId`. Residual (not blocking): still one `getIdentity` per reactor, but in-process and behind `CachedIdentityStorage`.

component/core/src/main/java/io/meeds/social/reaction/service/ReactionServiceImpl.java
boubaker: ✅ **Fixed** in `9eacc03`, verified. `final CopyOnWriteArrayList` plus an `@Autowired(required = false)` setter with a null guard — this closes both the `ConcurrentModificationException` on the cross-addon `addPlugin` path and the dependency on Spring handing back a mutable list.

webapp/src/main/webapp/skin/less/portlet/ActivityStream/Style.less
boubaker: ✅ **Fixed** in `9eacc03`, verified — via a third option rather than the two I sketched, and a reasonable one. The block moved to its own `portlet/ActivityReactions/Style.less`, declared as a `<portal-skin>` (so the styles are present wherever the framework is injected, not just under ActivityStream) and registered in `webapp/pom.xml`. `reaction-morph-*` is gone in favour of Vuetify's `scale-transition`, and the `prefers-reduced-motion` guard was kept and extended to cover it. Residual: the rules are now global rather than scoped — one nit on that in Round #2. Promoting the motion helpers to platform-ui is still the right long-term home when a second consumer lands.

component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
boubaker: ✅ **Fixed** in `9eacc03`, verified. `EntityBuilderReactionsTest` covers three cases — including 150 items with the viewer's own item appended **last**, asserting size 100 and the viewer still present, which is exactly the invariant the like button depends on — plus the `assertSame` under-limit fast path. Registered in `component/service`'s `NoContainerTestSuite`, so it actually runs.

component/core/src/main/java/io/meeds/social/reaction/service/ReactionServiceImpl.java
boubaker: ✅ **Fixed** in `b64f279`, verified — via the tighter of the two options: the tag range is removed from `isEmojiModifier` again, and any sequence containing a tag character must now be exactly one of the three allowlisted GB subdivision flags (`TAGGED_EMOJI_SEQUENCES`). I decoded all three constants surrogate-pair by surrogate-pair: they are the correct England/Scotland/Wales sequences (…-E007F terminated, 14 UTF-16 units). Re-ran the exploit and the bank scan against the final logic: 🔥+7 tag chars → **rejected**, 🏴+garbage tags → **rejected**, all 2066 bank entries → **accepted**. The "storage bounded by the emoji vocabulary" invariant holds again, and the two new negative tests pin it.

webapp/src/main/webapp/vue-apps/activity-stream/components/activity/footer/actions/ActivityLikeAction.vue
boubaker: ✅ **Fixed** in `b64f279`, verified. `refreshLikers()` now swallows its own error (`.catch(() => null)`), so a liker-refresh hiccup after a successful write no longer surfaces "Unable to save your reaction". The comment component needed no equivalent — its `.then` chain is local state + synchronous emits, no network step.

component/service/src/main/java/io/meeds/social/reaction/rest/ReactionRest.java
boubaker: ✅ **Fixed** in `b64f279`, verified — `.map(identityManager::getIdentity)` binds the `long` overload directly.

webapp/src/main/webapp/skin/less/portlet/ActivityReactions/Style.less
boubaker: ✅ **Fixed** in `b64f279`, verified — the reduced-motion overrides are scoped to `.reaction-chooser-wrapper` (the class on the chooser's root element, which wraps the like button via the slot), so the blast radius now matches the feature.

boubaker:

Re-reviewed at head `b64f279` (`fix: Harden reactions after integration review - EXO-89289`), scoped to `9eacc03..b64f279` plus verification of every Round #2 finding in the source.

### Status of Round #2 findings

| # | Finding | Status |
|---|---|---|
| 1 | 🟡 Tag characters as free modifiers → unbounded distinct ids rendering identically | ✅ **Fixed** — tag range removed from `isEmojiModifier`; any sequence containing a tag character must now be exactly one of the three allowlisted GB subdivision flags (`TAGGED_EMOJI_SEQUENCES`). Verified the three constants surrogate-pair by surrogate-pair (correct England/Scotland/Wales sequences, `E007F`-terminated). Re-ran the exploit and the bank scan against the final logic: 🔥+7 tag chars **rejected**, 🏴+garbage tags **rejected**, **all 2066 bank entries accepted**. The spec's "storage bounded by the emoji vocabulary" invariant holds again, and two new negative tests pin it |
| 2 | 🟢 `.catch` fired when the write succeeded but the liker refresh failed | ✅ **Fixed** — `refreshLikers()` swallows its own error; the comment component needed no change (no async post-write step) |
| 3 | 🟢 New call to `getIdentity(String)` (`forRemoval`) | ✅ **Fixed** — `identityManager::getIdentity` binds the `long` overload |
| 4 | 🟢 Reduced-motion override reached every `.v-btn` platform-wide | ✅ **Fixed** — scoped to `.reaction-chooser-wrapper` |

### Also verified in this round

- The `ObjectAlreadyExistsException` import switch (`org.exoplatform.social.common` → `org.exoplatform.commons`) is a sound deprecation cleanup: the `social.common` variant is `@Deprecated(forRemoval = true)` and **extends** the `commons` one, so declaring/catching the parent still catches exactly what `MetadataService.createMetadataItem` throws.
- The allowlist check runs before the keycap/base loop and after the length gate — no ordering hole; a bare tag sequence with no base stays rejected.
- The `ActivityReactionsList.vue` spread-operator changes are cosmetic; behaviour identical to the verified Round #2 paging fix.

One suggestion, not a finding: the three `TAGGED_EMOJI_SEQUENCES` literals are unreadable as escaped surrogates — a one-line comment naming them (🏴󠁧󠁢󠁥󠁮󠁧󠁿 England, 🏴󠁧󠁢󠁳󠁣󠁴󠁿 Scotland, 🏴󠁧󠁢󠁷󠁬󠁳󠁿 Wales) would spare the next maintainer the decoding.

### Close-out

**All findings from Rounds #1 and #2 are resolved — nothing outstanding from the AI review side.** Round #1's eight nits remain not taken (documented trade-off, non-blocking).

### Classification: N1

Unchanged, and this final fix commit again touched the trust-boundary validation (`isCustomEmojiReaction`), so it re-entered review rather than inheriting earlier rounds. **The PR still requires validation by an Architect/Senior Developer before merging into `feature/mips` — it must not be auto-merged on AI review alone.** Author ≠ approver: @AzmiTouil authored all three commits and cannot be the approver. The PR carries no N1 label yet — worth setting before the merge decision.

boubaker (approved):
LGTM @AzmiTouil , please prepare a PR to update knowledge domains on Meeds-io/eng-standards

PR #361, opened 2020-10-22
Task-35726: Hide change password from settings for SSO users

azayati:
Hide security setting from settings for SSO users

PR #6008, opened 2026-08-22
feat: ask for many objects' category links in one query EXO-89589

bmestrallet:
`CategoryLinkService` could only be asked about one object at a time, so anything listing rows and showing what each is filed under ran **one query per row** — and there was no way to avoid it from outside social.

## Where it bites

The mail drawer. `EmailBoxStorage.fromEntity` calls `getLinkedIds` per message, so a mailbox with 500 cached messages runs ~500 serial SELECTs to draw a list; 5000 at the cache ceiling an administrator may configure. It isn't specific to mail — any domain listing categorised rows meets the same wall.

## Why nothing existing would do

| Existing method | Why not |
|---|---|
| `getLinkedIds(CategoryObject)` | the per-object call — this *is* the N+1 |
| `getLinkedIds(String objectType)` | which **categories** a type uses, not which categories each object is in |
| `getLinkedObjects(categoryId, types, offset, limit)` | goes by category, and isn't user-scoped |
| `CategoryService.getCategoryEntries(...)` | user-scoped, but a UI card builder: over-fetches 3× and filters per item — heavier than the N+1 it would replace |

## What this adds

`getLinkedIds(objectType, objectIds)` → `Map<String, List<Long>>`, down through a new named query beside the existing one.

A map rather than a flat list, so the grouping isn't pushed onto every caller. An object with no category is **absent** from the map rather than present with an empty list, so a caller can tell "none" from "not asked for".

## Purely additive

**196 lines, no deletions.** Every current caller keeps its method, its SQL and its results — the new named query sits beside the old one, the DAO method beside the old one, and each layer above gets an overload.

No schema change and no migration: the same `objectType` + `objectId` columns the existing query already uses, with `IN` instead of `=`.

The `IN` list is chunked at 1000 — Oracle's limit (ORA-01795) — following `IdentityDAOImpl`, which meets the same bound in that package.

## Considered and rejected

Adding `@Cacheable` to `CategoryStorage.getLinkedIds`, which its three sibling methods already carry. It's one annotation, but it changes the behaviour of **every existing caller** rather than adding a new path, needs eviction to be right on every link and unlink, and helps least where it's needed most: auto-categorization links constantly, so entries would be evicted often, and a cold mailbox still pays every query.

## Verification

`component/api` and `component/core` compile. `CategoryLinkServiceTest` 6/6, metadata suites 38/38 green.

The new test asserts the bulk answer **equals the per-object one, object by object**. That equivalence is the whole contract: the call exists so a caller can stop asking per row, and is worth nothing if the two can drift apart.

Not measured against a large dataset — the win is structural (N queries become one), and the caller that will demonstrate it is the follow-up below.

## Follow-up, not part of this

email-connector's listing switches to the bulk call once this is released: one query for the page instead of one per row, taking the drawer's open from ~505 statements to about 6. Context in [EXO-89588](https://community.exoplatform.com/portal/dw/tasks/taskDetail/89588) and PRs [#360](https://github.com/exoplatform/email-connector/pull/360) / [#361](https://github.com/exoplatform/email-connector/pull/361).

**N2** — shared component, additive API, no schema/ACL/trust-boundary surface. Worth a reviewer who knows the metadata layer, since that is where the new query lives.

### DIFF
--- a/component/api/src/main/java/io/meeds/social/category/service/CategoryLinkService.java
+++ b/component/api/src/main/java/io/meeds/social/category/service/CategoryLinkService.java
@@ -19,6 +19,7 @@
 package io.meeds.social.category.service;
 
 import java.util.List;
+import java.util.Map;
 
 import org.exoplatform.commons.ObjectAlreadyExistsException;
 import org.exoplatform.commons.exception.ObjectNotFoundException;
@@ -38,6 +39,22 @@ public interface CategoryLinkService {
    * @return {@link List} of linked {@link Category} identifiers
    */
   List<Long> getLinkedIds(CategoryObject object);
+
+  /**
+   * The categories each of several objects is filed under, in one query.
+   * <p>
+   * {@link #getLinkedIds(CategoryObject)} is the right call for a reader looking at one
+   * thing, and the wrong one for a list: a page of rows showing what each is filed under
+   * would ask it once per row. This answers the whole page, so the caller decorates its
+   * rows by lookup instead of by a query each.
+   *
+   * @param objectType {@link CategoryObject} type, example: Space, Activity ...
+   * @param objectIds the objects' identifiers
+   * @return {@link Map} of linked {@link Category} identifiers keyed by object id; an
+   *         object with no category is absent from the map rather than present with an
+   *         empty list
+   */
+  Map<String, List<Long>> getLinkedIds(String objectType, List<String> objectIds);
 
   /**
    * @param objectType {@link CategoryObject} type, example: Space, Activity ...
--- a/component/api/src/main/java/org/exoplatform/social/metadata/MetadataService.java
+++ b/component/api/src/main/java/org/exoplatform/social/metadata/MetadataService.java
@@ -380,6 +380,20 @@ public interface MetadataService {
   List<MetadataItem> getMetadataItemsByMetadataTypeAndObject(String metadataType, MetadataObject object);
 
   /**
+   * Retrieves the {@link MetadataItem} of several objects of one type at once, so a
+   * caller listing rows can ask a single question instead of one per row.
+   *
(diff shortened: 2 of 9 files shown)


bmestrallet:
## Measured

Against the suite's own database, 500 linked objects, timing the per-object read against the bulk one:

```
per-object (500 queries) :  38.8 ms
bulk       (1 query)     :   4.0 ms
factor                   :   9.6x
links found              :  500 both ways  ✓
```

The equality at the end is worth as much as the timing: both paths returned the same 500 links, which is the property the unit test asserts on two objects, holding at page scale.

### 9.6× is the floor, not the gain

That ran on the in-process HSQLDB the suite uses — the JVM and the database in one process, so **no network round trip**. Per query it is 0.078 ms, which is why 500 of them cost only 39 ms.

A real deployment adds a round trip to every one of the 500, and to exactly one of the bulk call:

| deployment | per-object | bulk | factor |
|---|---|---|---|
| in-process (measured) | 39 ms | 4 ms | 9.6× |
| MySQL, local socket (~0.15 ms RTT) | ~115 ms | ~4 ms | ~28× |
| MySQL, same-DC network (~0.4 ms RTT) | ~240 ms | ~5 ms | ~50× |

The last two rows are arithmetic — 500 × RTT added to the measured floor — not measurement. I could not measure them here: the local platform runs HSQLDB and no MySQL was available.

That is the shape of an N+1 generally, and why the in-process number understates it: the cost that matters scales with **round trips**, not with bytes or rows, and it is precisely the cost an in-process database does not have.

### Context

These figures are per page. The caller that prompted this — the mail drawer — currently also re-requests its listing every 2s for up to 30 polls after each open, so the same figures apply thirty times over the following minute until that is fixed separately.

Benchmark was a temporary test, not committed: it linked 500 objects, warmed both paths, then timed them.

PR #6022, opened 2026-08-27
fix: Fix the navigation menu accessibility - EXO-88911

MayTekayaa:
Prior to this change, the top bar menu could not be walked with the keyboard beyond the level the focus happened to sit on — Vuetify's own highlight and the code's focus were two competing cursors, so an arrow pressed inside a submenu moved the level above it — and a screen reader announced neither the parent node nor the submenu being entered.

This change leaves a single cursor, the focus, and reads the levels from the component tree instead of the DOM, so the arrows always walk the list on screen at any depth (Escape/ArrowLeft step back, RTL flipped), and every menu is announced with the name of the node that opens it.

### DIFF
--- a/webapp/src/main/webapp/skin/less/portlet/TopBarMenu/Style.less
+++ b/webapp/src/main/webapp/skin/less/portlet/TopBarMenu/Style.less
@@ -67,3 +67,24 @@
   top: -10%;
   position: absolute;
 }
+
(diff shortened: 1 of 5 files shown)


ahamdi (commented):
@MayTekayaa  could you resolve the conflicts ?

ahamdi (dismissed):

ahamdi (approved):

PR #6028, opened 2026-08-28
fix: Allow submenus control with the keyboard in the left menu - EXO-76247

mbensalem:
Before this change, submenus in the left menu could not be operated with the keyboard.To fix this, keyboard navigation/focus handling was added to the sidebar submenu components.After this change, submenus can be opened and navigated entirely with the keyboard


webapp/src/main/webapp/skin/less/portlet/Sidebar/Style.less
```diff
@@ -51,6 +51,34 @@
...
+  // at all times; only its visibility is hover/focus-driven, so keyboard
+  // users can actually tab to it instead of it never existing until hovered.
+  .menu-toggle-arrow {
+    opacity: 0;
+  }
+  .v-list-item:hover .menu-toggle-arrow,
+  .v-list-item:focus-within .menu-toggle-arrow,
+  .menu-toggle-arrow.menu-toggle-arrow-visible {
+    opacity: 1;
+  }
```
ahamdi, line 88: 🟠 **High — the always-rendered arrow permanently steals 44px from every menu row, and becomes an invisible tap target on touch**

Related: [`SidebarListItem.vue:318-320`](https://github.com/Meeds-io/social/blob/bd8bd03bce378ae40c344e0ef257844012812bdd/webapp/src/main/webapp/vue-apps/sidebar/components/list/SidebarListItem.vue#L318-L320) (`hasArrow`) · [`SpaceNavigationItem.vue:76-96`](https://github.com/Meeds-io/social/blob/bd8bd03bce378ae40c344e0ef257844012812bdd/webapp/src/main/webapp/vue-apps/sidebar/components/space/SpaceNavigationItem.vue#L76-L96)

```less
.menu-toggle-arrow { opacity: 0; }   // still in flow, still hit-testable
```

`opacity: 0` removes neither the layout box nor pointer events. The arrow is a `v-btn icon` = **36px** (`.v-btn--icon.v-size--default`, verified in `platform-ui`'s `vuetify.less:18034`) plus `ms-2`/`me-2` = **44px reserved on every SITE / SPACE / SPACES / SPACE_TEMPLATE / SPACE_CATEGORY row**, in a 310px drawer, in the non-hovered state — so `text-truncate` titles that used to fit are now permanently clipped.

Second consequence: on touch there is no `:hover`, so the arrow stays invisible **and** stays a 36px tap target with `@touchend.stop.prevent="emitAction(...)"` — tapping the right edge of a space row now opens the submenu instead of navigating. Before this PR the element was `v-if`'d out entirely on touch (`hover` never becomes true: the wrapper is a plain `div`, not `v-hover`, when `displaySequentially` is false).

**Fix:** keep it rendered (that is what makes it tabbable) but out of flow and inert while hidden — `position: absolute; pointer-events: none` in the `opacity: 0` state, reverting to `position: static; pointer-events: auto` under the existing `:hover` / `:focus-within` / `.menu-toggle-arrow-visible` rules. Worth checking in both LTR and RTL, and confirming with the PO that the non-hovered row width is unchanged.

webapp/src/main/webapp/vue-apps/site-details/components/SiteNavigationTree.vue
```diff
@@ -63,6 +63,12 @@ export default {
     selectedNodeUri: eXo.env.portal.selectedNodeUri,
     currentSite: eXo.env.portal.siteKeyName,
   }),
+  mounted() {
+    this.enhanceToggleButtonsAccessibility();
+  },
+  updated() {
+    this.enhanceToggleButtonsAccessibility();
+  },
```
ahamdi, line 72: 🟡 **Medium — `aria-expanded` on the treeview toggles is set once and never updated**

```js
updated() {
  this.enhanceToggleButtonsAccessibility();   // never fires on expand/collapse
},
```

`v-treeview` owns its open state internally, and `:open.sync="openLevel"` targets a **computed with no setter** ([`openLevel`, L73-95](https://github.com/Meeds-io/social/blob/bd8bd03bce378ae40c344e0ef257844012812bdd/webapp/src/main/webapp/vue-apps/site-details/components/SiteNavigationTree.vue#L73-L95)) — the `update:open` assignment is a no-op. So expanding a node re-renders `VTreeviewNode`, not this component; the parent's render watcher never re-runs and `updated()` never fires. `aria-expanded` therefore reports the state at mount forever, which is worse than omitting it — a screen reader is told "collapsed" on an expanded node.

**Fix:** drive the sync from the state change instead of the parent's lifecycle — re-run the attribute pass from the toggle's own `click`/`keydown` handler on the next frame, or attach a `MutationObserver` on each toggle's `class` (`v-treeview-node__toggle--open` is the source of truth).

webapp/src/main/webapp/vue-apps/site-details/components/SiteNavigationTree.vue
```diff
@@ -109,6 +115,26 @@ export default {
...
+      this.$nextTick(() => {
+        this.$el.querySelectorAll?.('.v-treeview-node__toggle')?.forEach(toggle => {
+          const open = toggle.classList.contains('v-treeview-node__toggle--open');
+          toggle.setAttribute('aria-expanded', open ? 'true' : 'false');
+          if (toggle.dataset.a11yEnhanced) {
+            return;
+          }
+          toggle.dataset.a11yEnhanced = 'true';
+          toggle.setAttribute('tabindex', '0');
+          toggle.setAttribute('role', 'button');
```
ahamdi, line 131: 🟡 **Medium — the treeview toggles get `role="button"` but no accessible name**

```js
toggle.setAttribute('tabindex', '0');
toggle.setAttribute('role', 'button');   // no aria-label; content is an icon font
```

`.v-treeview-node__toggle` is an `<i class="v-icon">` whose glyph comes from a CSS class — no text node. Declaring it a button without a label makes it a tab stop announced as an unnamed "button", which fails WCAG 4.1.2 and adds a stop that a keyboard user cannot identify.

**Fix:** set an `aria-label` from the node's label (e.g. `$t('menu.expandNode', {0: name})` / `menu.collapseNode`) — the node text is available from the toggle's sibling `.v-treeview-node__label` — and add `aria-controls` pointing at the child group if one can be identified.

webapp/src/main/webapp/skin/less/portlet/Sidebar/Style.less
```diff
@@ -51,6 +51,34 @@
...
+  // Keyboard users navigating the first/second/third level content need a
+  // clearly visible focus indicator; the browser default outline is too thin
+  // against this panel's background. Inset so it isn't clipped by the
+  // scrollable panels' overflow.
+  a:focus-visible,
+  button:focus-visible,
+  [tabindex]:focus-visible {
+    outline: 2px solid var(--primary, #3f51b5);
+    outline-offset: -2px;
+  }
```
ahamdi, line 104: 🟡 **Medium — the new focus ring uses a CSS variable that does not exist, so it never matches the theme**

```less
outline: 2px solid var(--primary, #3f51b5);
```

`--primary` is defined nowhere in `platform-ui`'s skin (grepped: the platform's custom property is `--allPagesPrimaryColor`, wrapped as `@primaryColor` in `variables.less:30`, default `#3f8487`). So the fallback always wins and every focus ring is Vuetify's indigo `#3f51b5`, ignoring the administrator's branding entirely — and `frontend-vue.md` is explicit that colours are variabilised because theming is set through the UI.

**Fix:** `outline: 2px solid @primaryColor;` — this file already does `@import "../../variables.less"` at line 21, so nothing else is needed.

webapp/src/main/webapp/vue-apps/sidebar/components/drawer/SidebarSecondLevel.vue
```diff
@@ -102,6 +106,14 @@ export default {
...
       return this.$root.expand;
     },
+    dialogLabel() {
+      if (this.secondLevel === 'site') {
+        return this.site?.name || this.$t('menu.spaces.yourSpaces');
+      } else if (this.secondLevel === 'spaceMenu') {
+        return this.openedSpace?.displayName || this.$t('menu.spaces.yourSpaces');
+      }
+      return this.$t('menu.spaces.yourSpaces');
+    },
```
ahamdi, line 123: 🟡 **Medium — the second-level panel is announced by the site's technical name**

```js
if (this.secondLevel === 'site') {
  return this.site?.name || ...;   // "global" / "intranet", not the label
}
```

`site.name` is the site key — the same value passed as `:site-name` to the navigation tree — while the human label is `site.displayName`, which is what `SiteDetails.vue:46-47` renders in the panel's own title. The `aria-label` therefore contradicts the visible heading.

**Fix:** `return this.site?.displayName || this.site?.name || …`. While there: the `'spaces'` case falls through to the generic `menu.spaces.yourSpaces` even when `$root.openedSpaceTemplateName` / `openedSpaceCategoryName` hold the actual heading.

webapp/src/main/webapp/vue-apps/sidebar/components/drawer/SidebarSecondLevel.vue
```diff
@@ -123,5 +135,28 @@ export default {
...
+      const first = focusable[0];
+      const last = focusable[focusable.length - 1];
+      if (event.shiftKey && document.activeElement === first) {
+        event.preventDefault();
+        last.focus();
+      } else if (!event.shiftKey && document.activeElement === last) {
+        event.preventDefault();
+        first.focus();
+      }
+    },
```
ahamdi: 🟡 **Medium — the focus trap is unconditional, including when the panel is not modal**

Same code in [`SidebarThirdLevel.vue:114-131`](https://github.com/Meeds-io/social/blob/bd8bd03bce378ae40c344e0ef257844012812bdd/webapp/src/main/webapp/vue-apps/sidebar/components/drawer/SidebarThirdLevel.vue#L114-L131).

`showInnerOverlay` is `$root.sticky && (secondLevelDrawer || thirdLevelDrawer)` ([`Sidebar.vue:139-141`](https://github.com/Meeds-io/social/blob/bd8bd03bce378ae40c344e0ef257844012812bdd/webapp/src/main/webapp/vue-apps/sidebar/components/Sidebar.vue#L139-L141)) — so in **ICON** mode there is no overlay, the panel is a non-modal part of the page, and Tab is nonetheless cycled inside it. A user who tabs in from the page can only leave via Esc, which is never announced (`role="region"`, not `dialog`). That is WCAG 2.1.2 territory.

Related gap: the first-level panel has **no** Esc handler and no focus restore at all (grepped — no `keydown`/`esc` in `SidebarFirstLevel.vue` or `SidebarButton.vue`), so a keyboard user who Escapes out of the second level is left inside an open first level with no way to close it.

**Fix:** gate the trap on modality — trap only while `$root.sticky` (overlay shown), and let Tab flow out otherwise; and give the first-level drawer the same `@keydown.esc.native` + focus-restore treatment. If the trap stays unconditional, `role="dialog"` + `aria-modal="true"` would at least make the modality announced (the computed is already named `dialogLabel`).

ahamdi (commented):
## AI review — Round #1

Reviewed at `bd8bd03` — 8 files, frontend only (`.vue` / `.less` / `.js`). One 🟠, five 🟡 and a handful of nits, all posted inline except the nits below.

### 🟢 Nit — a few small ones

- **Duplicated logic.** `handlePanelKeydown` is byte-identical in the two panel components, and `focusOpenedPanel` byte-identical in `SidebarListItem` / `SpaceNavigationItem`. Four copies of a focus helper is what a mixin (or `vue-apps/common`) is for — `frontend-vue.md`'s reuse rule.
- **`querySelector` with a selector list returns document order, not first-selector-match** ([`SidebarListItem.vue:437`](https://github.com/Meeds-io/social/blob/bd8bd03bce378ae40c344e0ef257844012812bdd/webapp/src/main/webapp/vue-apps/sidebar/components/list/SidebarListItem.vue#L437)). Listing `ThirdLevelParent` first does not prefer it; it only works because the desktop arm of `Sidebar.vue` happens to render third before second. Query the two panels separately if the preference is intended.
- **Tab can be swallowed.** `onArrowKeydown` calls `preventDefault()` before `focusOpenedPanel` knows a target exists; on the early-return path focus is stuck on the arrow. Prevent only after a target is found.
- **`aria-owns` is redundant** with `aria-controls` pointing at the same node, and is the more brittle of the two in practice — consider dropping it.
- **Dangling `aria-controls`** — `HamburgerMenuSecondLevelPanel` does not exist in the DOM while `allowDisplayLevels` is false.
- **`:arial-label` typo** survives the rewrite at [`SpaceNavigationItem.vue:58`](https://github.com/Meeds-io/social/blob/bd8bd03bce378ae40c344e0ef257844012812bdd/webapp/src/main/webapp/vue-apps/sidebar/components/space/SpaceNavigationItem.vue#L58) — pre-existing, but this is the a11y PR that touches that exact line.
- **Listeners are never removed** in `enhanceToggleButtonsAccessibility` (bounded by node lifetime, so low impact), and `$root.lastSecondLevelFocusElement` is also written on the mouse paths, where `document.activeElement` is `<body>` — a later Esc then blurs instead of restoring.

### Verified conform

- **Attribute fall-through and `@keydown.native` on `ripple-hover-button`** — its root *is* the `v-btn` ([`RippleHoverButton.vue:22-33`](https://github.com/Meeds-io/social/blob/bd8bd03bce378ae40c344e0ef257844012812bdd/webapp/src/main/webapp/vue-apps/common/components/RippleHoverButton.vue#L22-L33)), so `aria-expanded` / `aria-controls` / `title` and the new keydown listener all land on the real `<button>`.
- **The `preventDefault()` rationale in the comment is accurate** — suppressing the native Enter/Space→click does bypass the 500 ms `emitted` debounce in `emitActionOnce` ([`RippleHoverButton.vue:82-90`](https://github.com/Meeds-io/social/blob/bd8bd03bce378ae40c344e0ef257844012812bdd/webapp/src/main/webapp/vue-apps/common/components/RippleHoverButton.vue#L82-L90)), and `stopPropagation()` does prevent the parent `v-list-item`'s Vuetify keydown→click from double-toggling on the `isSpaces` branch (that item carries `@click="handleSpacesClick"`).
- **No duplicate DOM ids** — the two `<sidebar-second-level>` / `<sidebar-third-level>` blocks in `Sidebar.vue` are the `v-if` / `v-else` arms of `$root.displaySequentially`, never both mounted.
- **`v-navigation-drawer--open` is a real Vuetify 2 class** (confirmed in the shipped `webapp/src/main/webapp/js/lib/vuetify.min.js`), so `focusOpenedPanel`'s selector resolves.
- **The focus trap can't misfire on a closed panel** — the panel body is `<v-hover v-if="drawer">`, so a closed drawer has no focusable children and `handlePanelKeydown` returns early.
- **CSS scope is right** — `.HamburgerNavigationMenu` is the `v-app` root and does contain both drawers, so the new `.menu-toggle-arrow` / `:focus-visible` rules reach `SpaceNavigationItem` inside the panels too.
- **Esc restore chain** — `thirdLevel` is passed `true` down `SpacesHamburgerNavigation → SpacesNavigationContent → SpaceNavigationItem`, so `lastThirdLevelFocusElement` / `lastSecondLevelFocusElement` are stored on the right slot and Esc returns focus to the arrow that opened the panel.

### What the PR does well

The hard part of this fix — the `ripple-hover-button` is a *mouse* component whose `click` path is debounced and whose `@ripple-hover` contract is hover-shaped — is diagnosed correctly and solved at the right place, with the reasoning written into the code as comments rather than left implicit. Removing `v-hover`/`showItemActions` from `SpaceNavigationItem` in favour of CSS state is the right direction: hover state that gates *existence* is exactly what makes a control unreachable by keyboard. And keeping the arrow's activation, the panel's focus containment, and the Esc-restore in three separate, individually reviewable pieces made this diff easy to follow.

### Classification

**N2 — mixed.** No REST / DAO / schema / ACL surface is touched, but the sidebar is a shared component rendered on *every* page of the platform, so the blast radius is cross-domain (`ai-review-and-merge.md` §1/§2). Per the interim control in §5: this PR must be acked by an Architect/Developer who knows it is N2 — not merged on AI review alone. There is no automated frontend test suite in `webapp/`, so the 🟠 and the stale-`aria-expanded` finding in particular need a manual pass (desktop hover, touch, and a screen-reader check on the treeview).

mbensalem:
@ahamdi regarding the unconditional focus-trap in ICON mode: left as-is on purpose for this PR. Gating the trap on modality (trap only while the panel is actually modal) is tracked as a separate follow-up task, to be handled once this PR is validated.

ahamdi (approved):

PR #6029, opened 2026-08-31
feat: ask for many objects' category links in one query EXO-89589

Jihed525:
Backport of `feature/ai-contribution` to `develop` for EXO-89589 (source PR #6008), tested and validated.

### Included tasks
| Ticket | Source PR | What it brings |
|---|---|---|
| EXO-89589 | #6008 | a page of objects asks for its category links once, not once per row |

### Scope
Only EXO-89589. The three commits before it on the branch (EXO-88511, EXO-88493) belong to
another board and are deliberately left behind.

### Conflict resolution — needs a real look
`MetadataItemDAO.java` conflicted: develop has since gained
`getMetadataItemsByMetadataTypeAndObjectAndCreators` and
`countMetadataItemsByMetadataTypeAndObjectGroupedByMetadataName` in the same region. Resolved
**keep-both** — the two develop methods and the new bulk read are independent additions, none
was dropped or reordered. That hunk re-enters classification and wants a full review.

### Verification
`mvn install`: **BUILD SUCCESS** across the reactor.

### Downstream
exoplatform/email-connector EXO-89588 calls `getLinkedIds(String, List<String>)` from this PR
and does not compile without it. EXO-89588 is on Wait behind this one; once this merges, it can go.

### Classification
**N1** by the deterministic trigger — `MetadataItemEntity.java` is a JPA entity file. In
substance the change is a `@NamedQuery` addition plus a chunked DAO read: no column, no table,
no migration. Flagged up rather than argued down; the level is a human's to lower, not mine.

Knowledge: none — mechanical backport, no new mechanism.

### DIFF
--- a/component/core/src/main/java/io/meeds/social/category/service/CategoryLinkServiceImpl.java
+++ b/component/core/src/main/java/io/meeds/social/category/service/CategoryLinkServiceImpl.java
@@ -19,6 +19,7 @@
 package io.meeds.social.category.service;
 
 import java.util.List;
+import java.util.Map;
 
 import org.apache.commons.lang3.ObjectUtils;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -60,6 +61,11 @@ public class CategoryLinkServiceImpl implements CategoryLinkService {
   @Override
   public List<Long> getLinkedIds(CategoryObject object) {
     return categoryStorage.getLinkedIds(getObject(object));
+  }
+
+  @Override
+  public Map<String, List<Long>> getLinkedIds(String objectType, List<String> objectIds) {
+    return categoryStorage.getLinkedIds(objectType, objectIds);
   }
 
   @Override
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
@@ -72,9 +72,12 @@ public class MetadataItemDAO extends GenericDAOJPAImpl<MetadataItemEntity, Long>
 
   private static final String OBJECT_IDS           = "objectIds";
 
-  // Oracle's hard limit on IN literals (ORA-01795), as IdentityDAOImpl uses in this
-  // package. ReactionStorage chunks the same table at 500; that is a throughput
-  // choice for a hot path, this one is the dialect bound itself.
+  // A self-imposed bound, not a dialect limit: the platform supports MySQL and
+  // PostgreSQL, and neither refuses an IN list at any particular count. It is here
+  // because an unbounded list still has to be serialised into one statement and
+  // digested by the planner, and a page of rows can be arbitrarily long. 1000 matches
+  // IdentityDAOImpl in this package - which declares the same constant with no stated
+  // reason - so the two agree; ReactionStorage chunks the same table at 500.
   private static final int    MAX_ITEMS_PER_IN_CLAUSE = 1000;
 
   private static final String SPACE_ID             = "spaceId";
@@ -179,10 +182,12 @@ public class MetadataItemDAO extends GenericDAOJPAImpl<MetadataItemEntity, Long>
    * The same read for a page of objects at once, so a caller listing rows can ask once
    * instead of once per row.
    * <p>
-   * Issued in chunks: Oracle refuses an IN list past 1000 literals (ORA-01795), which a
-   * page of rows can exceed — a mailbox caches a thousand messages by default and may be
-   * configured for five thousand. The chunk size follows {@code IdentityDAOImpl}, which
-   * meets the same limit in this package.
+   * Issued in chunks, because a page of rows can be arbitrarily long — a mailbox caches a
+   * thousand messages by default and may be configured for five thousand — and the whole
+   * list would otherwise go into a single statement. The bound is self-imposed rather than
+   * a dialect limit: on MySQL and PostgreSQL, the databases the platform supports, an IN
+   * list has no fixed maximum count. The size follows {@code IdentityDAOImpl} in this
+   * package.
    *
    * @param metadataType the metadata type id
    * @param objectType the object type, e.g. the one a category plugin registers
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
@@ -72,8 +72,6 @@ public class MetadataItemDAO extends GenericDAOJPAImpl<MetadataItemEntity, Long>
 
   private static final String OBJECT_IDS           = "objectIds";
 
-  // Oracle refuses an IN list past 1000 literals (ORA-01795); same bound, and same
-  // reason, as IdentityDAOImpl in this package.
   // Oracle's hard limit on IN literals (ORA-01795), as IdentityDAOImpl uses in this
   // package. ReactionStorage chunks the same table at 500; that is a throughput
   // choice for a hot path, this one is the dialect bound itself.
@@ -203,17 +201,23 @@ public class MetadataItemDAO extends GenericDAOJPAImpl<MetadataItemEntity, Long>
                                                                              MetadataItemEntity.class);
     query.setParameter(METADATA_TYPE, metadataType);
     query.setParameter(OBJECT_TYPE, objectType);
+    // SQL IN collapses duplicate literals, so a repeated id costs nothing inside one
+    // chunk - but the chunks partition this list positionally, and a duplicate that
+    // straddles a boundary is asked for twice and its rows added twice, which hands the
+    // caller that object's categories doubled. Distinct once, up front; it also makes
+    // the sort guard below exact rather than approximate.
+    List<String> distinctIds = objectIds.stream().distinct().toList();
     List<MetadataItemEntity> items = new ArrayList<>();
-    for (int start = 0; start < objectIds.size(); start += MAX_ITEMS_PER_IN_CLAUSE) {
-      int end = Math.min(start + MAX_ITEMS_PER_IN_CLAUSE, objectIds.size());
-      query.setParameter(OBJECT_IDS, objectIds.subList(start, end));
+    for (int start = 0; start < distinctIds.size(); start += MAX_ITEMS_PER_IN_CLAUSE) {
+      int end = Math.min(start + MAX_ITEMS_PER_IN_CLAUSE, distinctIds.size());
+      query.setParameter(OBJECT_IDS, distinctIds.subList(start, end));
       items.addAll(query.getResultList());
     }
     // Each chunk is ordered, their concatenation is not: object ids ascend within a chunk
     // and restart at the next one. All the rows of one object sit in a single chunk, so a
     // stable sort on the object id restores the documented grouping while leaving the
     // within-object ordering the query established untouched.
-    if (objectIds.size() > MAX_ITEMS_PER_IN_CLAUSE) {
+    if (distinctIds.size() > MAX_ITEMS_PER_IN_CLAUSE) {
       items.sort(Comparator.comparing(MetadataItemEntity::getObjectId));
     }
     return items;
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
@@ -18,6 +18,7 @@
  */
 package org.exoplatform.social.core.jpa.storage.dao.jpa;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
@@ -67,6 +68,12 @@ public class MetadataItemDAO extends GenericDAOJPAImpl<MetadataItemEntity, Long>
   private static final String PARENT_OBJECT_ID     = "parentObjectId";
 
   private static final String OBJECT_ID            = "objectId";
+
+  private static final String OBJECT_IDS           = "objectIds";
+
+  // Oracle refuses an IN list past 1000 literals (ORA-01795); same bound, and same
+  // reason, as IdentityDAOImpl in this package.
+  private static final int    MAX_ITEMS_PER_IN_CLAUSE = 1000;
 
   private static final String SPACE_ID             = "spaceId";
 
@@ -164,6 +171,40 @@ public class MetadataItemDAO extends GenericDAOJPAImpl<MetadataItemEntity, Long>
                 .stream()
                 .collect(Collectors.toMap(tuple -> tuple.get("name", String.class),
                                           tuple -> tuple.get("itemsCount", Long.class)));
+  }
+
+  /**
+   * The same read for a page of objects at once, so a caller listing rows can ask once
+   * instead of once per row.
+   * <p>
+   * Issued in chunks: Oracle refuses an IN list past 1000 literals (ORA-01795), which a
+   * page of rows can exceed — a mailbox caches a thousand messages by default and may be
+   * configured for five thousand. The chunk size follows {@code IdentityDAOImpl}, which
+   * meets the same limit in this package.
+   *
+   * @param metadataType the metadata type id
+   * @param objectType the object type, e.g. the one a category plugin registers
+   * @param objectIds the object ids to read, in any order
+   * @return the items of those objects, empty when no id is given
+   */
+  public List<MetadataItemEntity> getMetadataItemsByMetadataTypeAndObjectIds(long metadataType,
+                                                                             String objectType,
+                                                                             List<String> objectIds) {
+    if (CollectionUtils.isEmpty(objectIds)) {
+      return Collections.emptyList();
+    }
+    TypedQuery<MetadataItemEntity> query =
+                                         getEntityManager().createNamedQuery("SocMetadataItemEntity.getMetadataItemsByMetadataTypeAndObjectIds",
+                                                                             MetadataItemEntity.class);
+    query.setParameter(METADATA_TYPE, metadataType);
+    query.setParameter(OBJECT_TYPE, objectType);
+    List<MetadataItemEntity> items = new ArrayList<>();
+    for (int start = 0; start < objectIds.size(); start += MAX_ITEMS_PER_IN_CLAUSE) {
+      int end = Math.min(start + MAX_ITEMS_PER_IN_CLAUSE, objectIds.size());
+      query.setParameter(OBJECT_IDS, objectIds.subList(start, end));
+      items.addAll(query.getResultList());
+    }
+    return items;
   }
 
   public List<MetadataItemEntity> getMetadataItemsByMetadataTypeAndObject(long metadataType,
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
@@ -21,6 +21,7 @@ package org.exoplatform.social.core.jpa.storage.dao.jpa;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -73,6 +74,9 @@ public class MetadataItemDAO extends GenericDAOJPAImpl<MetadataItemEntity, Long>
 
   // Oracle refuses an IN list past 1000 literals (ORA-01795); same bound, and same
   // reason, as IdentityDAOImpl in this package.
+  // Oracle's hard limit on IN literals (ORA-01795), as IdentityDAOImpl uses in this
+  // package. ReactionStorage chunks the same table at 500; that is a throughput
+  // choice for a hot path, this one is the dialect bound itself.
   private static final int    MAX_ITEMS_PER_IN_CLAUSE = 1000;
 
   private static final String SPACE_ID             = "spaceId";
@@ -184,7 +188,8 @@ public class MetadataItemDAO extends GenericDAOJPAImpl<MetadataItemEntity, Long>
    *
    * @param metadataType the metadata type id
    * @param objectType the object type, e.g. the one a category plugin registers
-   * @param objectIds the object ids to read, in any order
+   * @param objectIds the object ids to read; the order they are given in does not
+   *          matter, the result is ordered by object either way
    * @return the items of those objects, empty when no id is given
    */
   public List<MetadataItemEntity> getMetadataItemsByMetadataTypeAndObjectIds(long metadataType,
@@ -203,6 +208,13 @@ public class MetadataItemDAO extends GenericDAOJPAImpl<MetadataItemEntity, Long>
       int end = Math.min(start + MAX_ITEMS_PER_IN_CLAUSE, objectIds.size());
       query.setParameter(OBJECT_IDS, objectIds.subList(start, end));
       items.addAll(query.getResultList());
+    }
+    // Each chunk is ordered, their concatenation is not: object ids ascend within a chunk
+    // and restart at the next one. All the rows of one object sit in a single chunk, so a
+    // stable sort on the object id restores the documented grouping while leaving the
+    // within-object ordering the query established untouched.
+    if (objectIds.size() > MAX_ITEMS_PER_IN_CLAUSE) {
+      items.sort(Comparator.comparing(MetadataItemEntity::getObjectId));
     }
     return items;
   }
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/entity/MetadataItemEntity.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/entity/MetadataItemEntity.java
@@ -70,16 +70,20 @@ import jakarta.persistence.Table;
         + " ORDER BY mi.id ASC"
 )
 // The same read for a page of objects at once. A caller listing rows and showing what
-// each is filed under would otherwise ask the query above once per row, which is a query
-// per message on a mail folder and per document on a drive. Ordered by object first so a
-// caller can group the answer as it walks it.
+// each is filed under would otherwise ask getSortedMetadataItemsByMetadataTypeAndObject
+// once per row, which is a query per message on a mail folder and per document on a drive.
+// Grouped by object first, then ordered within an object exactly as that per-object query
+// orders it - most recently touched first - so one object's slice of this result is the
+// same list, in the same order, that asking for it alone would return.
 @NamedQuery(
     name = "SocMetadataItemEntity.getMetadataItemsByMetadataTypeAndObjectIds",
     query = "SELECT mi FROM SocMetadataItemEntity mi WHERE "
         + " mi.metadata.type = :metadataType AND"
         + " mi.objectType = :objectType AND"
-        + " mi.objectId IN :objectIds"
-        + " ORDER BY mi.objectId ASC, mi.id ASC"
+        + " mi.objectId IN (:objectIds)"
+        + " ORDER BY mi.objectId ASC,"
+        + " COALESCE(mi.updatedDate, mi.createdDate) DESC,"
+        + " mi.createdDate DESC, mi.id DESC"
 )
 @NamedQuery(
     name = "SocMetadataItemEntity.getSortedMetadataItemsByMetadataTypeAndObject",
--- a/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/entity/MetadataItemEntity.java
+++ b/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/entity/MetadataItemEntity.java
@@ -69,6 +69,18 @@ import jakarta.persistence.Table;
         + " mi.objectId = :objectId"
         + " ORDER BY mi.id ASC"
 )
+// The same read for a page of objects at once. A caller listing rows and showing what
+// each is filed under would otherwise ask the query above once per row, which is a query
+// per message on a mail folder and per document on a drive. Ordered by object first so a
+// caller can group the answer as it walks it.
+@NamedQuery(
+    name = "SocMetadataItemEntity.getMetadataItemsByMetadataTypeAndObjectIds",
+    query = "SELECT mi FROM SocMetadataItemEntity mi WHERE "
+        + " mi.metadata.type = :metadataType AND"
+        + " mi.objectType = :objectType AND"
+        + " mi.objectId IN :objectIds"
+        + " ORDER BY mi.objectId ASC, mi.id ASC"
+)
 @NamedQuery(
     name = "SocMetadataItemEntity.getSortedMetadataItemsByMetadataTypeAndObject",
     query = "SELECT mi FROM SocMetadataItemEntity mi WHERE "
--- a/component/core/src/test/java/io/meeds/social/category/service/CategoryLinkServiceTest.java
+++ b/component/core/src/test/java/io/meeds/social/category/service/CategoryLinkServiceTest.java
@@ -56,8 +56,9 @@ public class CategoryLinkServiceTest extends AbstractCategoryConfigurationTest {
    * listing rows can stop asking once per row, and it is worth nothing if the two can
    * drift apart. It holds for ids in stored form - the bulk call does not put its ids
    * through the category plugin, as the caveat on CategoryLinkService#getLinkedIds
-   * spells out - which is the form this test uses. An object with no category is absent from the map rather than present
-   * with an empty list, so a caller can tell "none" from "not asked for".
+   * spells out - which is the form this test uses. An object with no category is
+   * absent from the map rather than present with an empty list, so a caller can tell
+   * "none" from "not asked for".
    */
   @Test
   @SneakyThrows
--- a/component/core/src/test/java/io/meeds/social/category/service/CategoryLinkServiceTest.java
+++ b/component/core/src/test/java/io/meeds/social/category/service/CategoryLinkServiceTest.java
@@ -71,6 +71,16 @@ public class CategoryLinkServiceTest extends AbstractCategoryConfigurationTest {
                                                      linkedSpace.getSpaceId());
     categoryLinkService.link(rootCategory.getId(), linkedObject);
 
+    // A second category on the same object is what makes the equivalence assertion below
+    // bite: with a single link the two calls return one-element lists and any ordering
+    // difference between them is invisible.
+    Category secondCategory = new Category();
+    secondCategory.setOwnerId(getAdminGroupIdentityId());
+    secondCategory.setParentId(rootCategory.getId());
+    secondCategory.setIcon("test-icon");
+    secondCategory = categoryService.createCategory(secondCategory, ROOT_USER);
+    categoryLinkService.link(secondCategory.getId(), linkedObject);
+
     Space unlinkedSpace = new Space();
     unlinkedSpace.setRegistration(Space.OPEN);
     unlinkedSpace.setVisibility(Space.PUBLIC);
@@ -82,9 +92,12 @@ public class CategoryLinkServiceTest extends AbstractCategoryConfigurationTest {
     Map<String, List<Long>> linkedIds = categoryLinkService.getLinkedIds(SpaceCategoryPlugin.OBJECT_TYPE,
                                                                          List.of(linkedObject.getId(), unlinkedObject.getId()));
 
-    // what the per-object call says, said for the whole page in one query
+    // what the per-object call says, said for the whole page in one query - same
+    // categories, same order, which only means something now the list has two entries
+    assertEquals(2, linkedIds.get(linkedObject.getId()).size());
     assertEquals(categoryLinkService.getLinkedIds(linkedObject), linkedIds.get(linkedObject.getId()));
     assertTrue(linkedIds.get(linkedObject.getId()).contains(rootCategory.getId()));
+    assertTrue(linkedIds.get(linkedObject.getId()).contains(secondCategory.getId()));
     // an object with no category is absent, not empty
     assertTrue(CollectionUtils.isEmpty(categoryLinkService.getLinkedIds(unlinkedObject)));
     assertNull(linkedIds.get(unlinkedObject.getId()));
--- a/component/core/src/test/java/io/meeds/social/category/service/CategoryLinkServiceTest.java
+++ b/component/core/src/test/java/io/meeds/social/category/service/CategoryLinkServiceTest.java
@@ -54,7 +54,9 @@ public class CategoryLinkServiceTest extends AbstractCategoryConfigurationTest {
    * <p>
    * That equivalence is the whole contract: the bulk call exists only so a caller
    * listing rows can stop asking once per row, and it is worth nothing if the two can
-   * drift apart. An object with no category is absent from the map rather than present
+   * drift apart. It holds for ids in stored form - the bulk call does not put its ids
+   * through the category plugin, as the caveat on CategoryLinkService#getLinkedIds
+   * spells out - which is the form this test uses. An object with no category is absent from the map rather than present
    * with an empty list, so a caller can tell "none" from "not asked for".
    */
   @Test
--- a/component/core/src/test/java/io/meeds/social/category/service/CategoryLinkServiceTest.java
+++ b/component/core/src/test/java/io/meeds/social/category/service/CategoryLinkServiceTest.java
@@ -19,6 +19,7 @@
 package io.meeds.social.category.service;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
@@ -26,6 +27,7 @@ import static org.junit.Assert.assertTrue;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.junit.Test;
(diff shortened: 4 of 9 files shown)


Jihed525:
Rebased onto current develop (it had moved 2 commits: #6030, #6032) and force-pushed `7063c007`. The keep-both resolution in `MetadataItemDAO` still holds — develop's `getMetadataItemsByMetadataTypeAndObjectAndCreators` and `countMetadataItemsByMetadataTypeAndObjectGroupedByMetadataName` sit alongside the new bulk read, none dropped. Rebuilt green.

component/core/src/main/java/io/meeds/social/category/service/CategoryLinkServiceImpl.java
```diff
@@ -62,6 +63,11 @@ public List<Long> getLinkedIds(CategoryObject object) {
     return categoryStorage.getLinkedIds(getObject(object));
   }
 
+  @Override
+  public Map<String, List<Long>> getLinkedIds(String objectType, List<String> objectIds) {
+    return categoryStorage.getLinkedIds(objectType, objectIds);
+  }
```
azayati, line 69: 🟠 **High — the bulk path skips the plugin object normalization the per-object path applies**

```java
public List<Long> getLinkedIds(CategoryObject object) {
  return categoryStorage.getLinkedIds(getObject(object));   // normalized through the plugin
}

@Override
public Map<String, List<Long>> getLinkedIds(String objectType, List<String> objectIds) {
  return categoryStorage.getLinkedIds(objectType, objectIds);  // raw ids, no getObject()
}
```

Every other read/write in this service (`isLinked`, `link`, `unlink`) routes its `CategoryObject` through `getObject()` → `categoryPluginService.getObject()` before touching storage, and two shipped plugins actually transform it: [`SpaceCategoryPlugin.getObject`](https://github.com/Meeds-io/social/blob/7063c007fa3b077d1735d90de192b71f74777d6c/component/core/src/main/java/io/meeds/social/space/plugin/SpaceCategoryPlugin.java#L98-L107) resolves a pretty name or a social identity id down to `String.valueOf(space.getSpaceId())`, and [`ActivityCategoryPlugin.getObject`](https://github.com/Meeds-io/social/blob/7063c007fa3b077d1735d90de192b71f74777d6c/component/core/src/main/java/io/meeds/social/activity/plugin/ActivityCategoryPlugin.java#L88-L91) redirects to `activity.getMetadataObject()`, which changes the **type** as well as the id.

So `getLinkedIds("space", List.of("mySpacePrettyName"))` returns an empty map where `getLinkedIds(new CategoryObject("space", "mySpacePrettyName", …))` returns the categories — and for activities the bulk call queries `objectType = "activity"` for links stored on the redirected object.

There is a second half to this: the map is keyed by `MetadataItem::getObjectId`, i.e. the **stored** id. Even if normalization were added inside, a caller that passed identity ids would get back a map keyed by technical ids and could not look up the row it asked about. The key must be the id the caller handed in.

**Fix:** normalize in `CategoryLinkServiceImpl` — map each incoming id through `categoryPluginService.getObject(new CategoryObject(objectType, id, null))`, group the storage answer by the *normalized* id, then re-key the returned map back to the caller's original ids. Where a plugin also rewrites the type, the bulk read has to be issued per resulting type; if that is deliberately out of scope, say so in the interface javadoc ("ids must already be in stored form; not equivalent to the per-object read for plugins that redirect").

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/entity/MetadataItemEntity.java
```diff
@@ -69,6 +69,18 @@
...
+// per message on a mail folder and per document on a drive. Ordered by object first so a
+// caller can group the answer as it walks it.
+@NamedQuery(
+    name = "SocMetadataItemEntity.getMetadataItemsByMetadataTypeAndObjectIds",
+    query = "SELECT mi FROM SocMetadataItemEntity mi WHERE "
+        + " mi.metadata.type = :metadataType AND"
+        + " mi.objectType = :objectType AND"
+        + " mi.objectId IN :objectIds"
+        + " ORDER BY mi.objectId ASC, mi.id ASC"
+)
```
azayati, line 87: 🟠 **High — the bulk read returns an object's categories in the opposite order to the per-object read**

```java
// new
+ " ORDER BY mi.objectId ASC, mi.id ASC"

// what the per-object category read actually runs — getSortedMetadataItemsByMetadataTypeAndObject, L88-L90
+ " ORDER BY COALESCE(mi.updatedDate, mi.createdDate) DESC,"
+ " mi.createdDate DESC, mi.id DESC"
```

The new query was modelled on `getMetadataItemsByMetadataTypeAndObject` (`ORDER BY mi.id ASC`), but that is **not** the query `CategoryStorage.getLinkedIds(CategoryObject)` ends up on: the chain [`CategoryStorage:157`](https://github.com/Meeds-io/social/blob/7063c007fa3b077d1735d90de192b71f74777d6c/component/core/src/main/java/io/meeds/social/category/storage/CategoryStorage.java#L157-L164) → `MetadataServiceImpl` → [`MetadataStorage:400-406`](https://github.com/Meeds-io/social/blob/7063c007fa3b077d1735d90de192b71f74777d6c/component/core/src/main/java/org/exoplatform/social/core/metadata/storage/MetadataStorage.java#L400-L411) → `MetadataItemDAO.getMetadataItemsByMetadataTypeAndObject(…, 0, 0)` runs the **Sorted** named query, most-recently-touched first.

So for any object filed under two or more categories, `bulk.get(id)` is the reverse of `getLinkedIds(object)`. The test asserts exactly that equivalence — [`CategoryLinkServiceTest.java:86`](https://github.com/Meeds-io/social/blob/7063c007fa3b077d1735d90de192b71f74777d6c/component/core/src/test/java/io/meeds/social/category/service/CategoryLinkServiceTest.java#L86) — but links only one category, so a list of size 1 makes the ordering untestable and the assertion passes vacuously.

**Fix:** order the new query `mi.objectId ASC, COALESCE(mi.updatedDate, mi.createdDate) DESC, mi.createdDate DESC, mi.id DESC` so the per-object slice matches, and extend the test to link a second category to `linkedSpace` — that one added link is what turns the equivalence assertion from decorative into real. The comment above should also point at `getSortedMetadataItemsByMetadataTypeAndObject`, not its unsorted neighbour ("the query above" is not the one the category path uses).

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
```diff
@@ -166,6 +173,40 @@ public Map<String, Long> countMetadataItemsByMetadataTypeAndObjectGroupedByMetad
...
+                                         getEntityManager().createNamedQuery("SocMetadataItemEntity.getMetadataItemsByMetadataTypeAndObjectIds",
+                                                                             MetadataItemEntity.class);
+    query.setParameter(METADATA_TYPE, metadataType);
+    query.setParameter(OBJECT_TYPE, objectType);
+    List<MetadataItemEntity> items = new ArrayList<>();
+    for (int start = 0; start < objectIds.size(); start += MAX_ITEMS_PER_IN_CLAUSE) {
+      int end = Math.min(start + MAX_ITEMS_PER_IN_CLAUSE, objectIds.size());
+      query.setParameter(OBJECT_IDS, objectIds.subList(start, end));
+      items.addAll(query.getResultList());
+    }
```
azayati: 🟡 **Medium — the documented "ordered by object" contract does not hold past 1000 ids**

```java
for (int start = 0; start < objectIds.size(); start += MAX_ITEMS_PER_IN_CLAUSE) {
  int end = Math.min(start + MAX_ITEMS_PER_IN_CLAUSE, objectIds.size());
  query.setParameter(OBJECT_IDS, objectIds.subList(start, end));
  items.addAll(query.getResultList());   // ordered per chunk, concatenated
}
```

Each chunk is ordered, the concatenation is not — objectIds ascend within a chunk and restart at the next. The named-query comment explicitly invites the usage that breaks ("Ordered by object first so a caller can group the answer as it walks it"): a caller doing a single-pass group over a >1000-id result closes a group and reopens it later, silently dropping or duplicating rows. `CategoryStorage` uses `Collectors.groupingBy`, so it is safe today — the exposure is the documented contract, and the mailbox scenario the javadoc itself cites ("a mailbox caches a thousand messages by default and may be configured for five thousand") is precisely the case that exceeds one chunk.

**Fix:** either drop the ordering promise from the named-query comment and this method's javadoc (say the result is unordered across chunks and callers must group by lookup), or sort `items` by `objectId` once after the loop. Also worth a test: the chunking loop has no coverage at all — no case crosses 1000 ids, so a boundary slip in `start`/`end` would ship green. `component/core` runs these against a real engine, so a ~1001-id case is cheap.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
```diff
@@ -68,6 +69,12 @@ public class MetadataItemDAO extends GenericDAOJPAImpl<MetadataItemEntity, Long>
 
   private static final String OBJECT_ID            = "objectId";
 
+  private static final String OBJECT_IDS           = "objectIds";
+
+  // Oracle refuses an IN list past 1000 literals (ORA-01795); same bound, and same
+  // reason, as IdentityDAOImpl in this package.
+  private static final int    MAX_ITEMS_PER_IN_CLAUSE = 1000;
```
azayati: 🟢 **Nit — a few small ones**

- The ORA-01795 bound is 1000 here, while the other chunked read of the same table uses 500 ([`ReactionStorage.java:42`](https://github.com/Meeds-io/social/blob/7063c007fa3b077d1735d90de192b71f74777d6c/component/core/src/main/java/io/meeds/social/reaction/storage/ReactionStorage.java#L42)). Both are defensible; a one-line note on why this one sits at the hard limit would stop the next reader reconciling them.
- [`MetadataItemDAO.java:187`](https://github.com/Meeds-io/social/blob/7063c007fa3b077d1735d90de192b71f74777d6c/component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java#L187) — `@param objectIds the object ids to read, in any order` reads oddly next to a query whose ordering is part of its contract; it means "input order is irrelevant", worth saying it that way.
- The new named query uses `IN :objectIds` where every sibling in `MetadataItemEntity` writes `IN (:objectTypes)`. Both parse in Hibernate; matching the local idiom costs nothing.

azayati (commented):
## AI review — Round #1

Backport of EXO-89589 (source #6008): a bulk `(objectType, objectIds) → category ids` read to kill an N+1 on list pages. 9 files, +196/-0. Everything below was verified against the source at head `7063c007`, read from a local checkout at that exact SHA.

**4 findings, inline:** 2 🟠 High (the bulk read is not equivalent to the per-object read it is meant to replace — plugin normalization skipped, and ordering reversed), 1 🟡 Medium (the documented ordering does not survive chunking past 1000 ids, and the chunking loop is untested), 1 🟢 Nit bundle.

The two High findings share one root: the bulk call is documented and tested as saying what the per-object call says, and in two respects it does not. Neither has a caller inside this repo today, so the exposure is the published API contract and the downstream `email-connector` consumer (EXO-89588), not a live regression here.

### Verified conform

- **Conflict resolution is clean.** `git diff origin/develop...HEAD` on `MetadataItemDAO.java` shows **zero deleted lines**; `getMetadataItemsByMetadataTypeAndObjectAndCreators` (:141) and `countMetadataItemsByMetadataTypeAndObjectGroupedByMetadataName` (:158) are both present and untouched, with the new bulk read added after them at :190. The keep-both claim in the PR body holds — nothing dropped, nothing reordered.
- **N1 classification is right and the substance is additive.** `MetadataItemEntity.java` gains a `@NamedQuery` only: no column, no table, no Liquibase changeset. Nothing to migrate.
- **Layering is respected end to end** — API interface → `ServiceImpl` passthrough → `Storage` → DAO, no layer skipped, `getMetadataTypeWithCheck` applied in `MetadataStorage` exactly as the siblings do.
- **Chunking sits in the DAO**, which is where the SQL dialect limit belongs, and follows `IdentityDAOImpl`'s `MAX_ITEMS_PER_IN_CLAUSE = 1000` in the same package (verified, :61) rather than inventing a bound.
- **No ACL regression.** The new read has no ACL, matching its per-object sibling; per the domain doc, `MetadataService` owns no ACL and callers do. No cross-audience widening: the query filters on the same `(metadata.type, objectType)` predicates as the single-object one.
- **No cache bypass.** `MetadataStorage`'s two `FutureExoCache` instances cover `Metadata` and `MetadataKey`, not metadata *items*; the per-object read is uncached too, so the bulk read introduces no coherence gap. `CategoryStorage.getLinkedIds` is not `@Cacheable`, so no missing `@CacheEvict`.
- Empty-input guards are present at all three layers and return immutable empties.

### Done well

The N+1 is attacked at the right altitude — one new query at the bottom, threaded up without any layer inventing semantics — and the javadoc explains *why* the bulk call exists and when the per-object one is still correct, which is the kind of comment that stops the next developer from calling the wrong one. The ORA-01795 chunking was anticipated rather than discovered in production, with the in-repo precedent cited. And the test asserts the bulk result against the per-object result rather than against hardcoded expectations — the right shape of assertion; it just needs a second link to bite.

### Knowledge gap for close-out

`domains/social.md` §10 documents the metadata query API as "name-keyed" and lists the reactions eXip's two additive methods; it now needs the third — a `(type, objectType, objectIds)` bulk read with DAO-level IN chunking — plus a note that `CategoryLinkService`'s per-object reads normalize through `CategoryPluginService.getObject` while the bulk one does not. Worth a `/domain-doc` refresh once the findings above land.

### Classification

**N1** — confirmed, not merely inherited from the PR body: the diff touches a JPA entity and adds a hand-written JPQL query on the metadata store, a socle surface. Must be validated by a human Architect/Developer; **not to be auto-merged on this AI review alone**.

azayati:
@Jihed525 the AI pre-review of Round #1 is posted above ([review](https://github.com/Meeds-io/social/pull/6029#pullrequestreview-5065457456)) — 4 findings as inline comments, verified against head `7063c007`.

Short version, in the order I'd take them:

1. 🟠 `CategoryLinkServiceImpl` — the bulk `getLinkedIds(String, List<String>)` doesn't route its ids through `categoryPluginService.getObject()`, which every other read/write in that service does. `SpaceCategoryPlugin` and `ActivityCategoryPlugin` both transform the object (a pretty name / identity id → space technical id; an activity → its redirected object, type included), so the bulk call answers empty where the per-object call answers correctly. Related: the map is keyed by the *stored* objectId, not the id the caller passed in.
2. 🟠 `MetadataItemEntity` — the new query orders `mi.id ASC`, but the query the category path actually runs is `getSortedMetadataItemsByMetadataTypeAndObject` (`COALESCE(updatedDate, createdDate) DESC, …, id DESC`), not the unsorted neighbour it was modelled on. Reversed order for any object with 2+ categories; the equivalence test only links one, so it passes vacuously.
3. 🟡 `MetadataItemDAO` — the "ordered by object so a caller can group as it walks it" promise doesn't survive the chunk concatenation past 1000 ids, which is exactly the mailbox case the javadoc cites. And the chunking loop itself has no test.
4. 🟢 Three small ones on the constant, a javadoc phrase and the `IN` idiom.

On the parts you flagged in the description: the `MetadataItemDAO` keep-both resolution checks out — `git diff origin/develop...HEAD` on that file has **zero deleted lines**, and both develop methods are present and untouched with the bulk read added after them. N1 confirmed, for the JPA entity + hand-written JPQL rather than for anything migratory.

Nothing here blocks the backport mechanically; findings 1 and 2 are about the new API's contract, which `email-connector` EXO-89588 is about to consume — worth settling before that lands. Human validation still required, this shouldn't merge on the AI review alone.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/entity/MetadataItemEntity.java
Jihed525: Confirmed and fixed in `219ce5b6`.

I verified the chain rather than take it on trust, and it is exactly as you describe: `MetadataStorage:402` passes **five** arguments, which binds the `(metadataType, objectType, objectId, offset, limit)` overload at `MetadataItemDAO:210` — and that one runs `getSortedMetadataItemsByMetadataTypeAndObject`. The three-argument overload at `:129` running the unsorted query is the one I modelled on, and nothing on the category path reaches it.

The query now reads:

```java
+ " ORDER BY mi.objectId ASC,"
+ " COALESCE(mi.updatedDate, mi.createdDate) DESC,"
+ " mi.createdDate DESC, mi.id DESC"
```

and the comment above it now names `getSortedMetadataItemsByMetadataTypeAndObject` instead of "the query above".

On the vacuous test — you were right, and it was worse than vacuous: the assertion could never have failed. I linked a second category to `linkedSpace`, then **mutation-verified** it by reverting the `ORDER BY` and re-running: `expected:<[22, 1]> but was:<[1, 22]>`. Restored, green. That is the check that turns the assertion from decorative into real.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
Jihed525: Confirmed and fixed in `219ce5b6`.

Kept the ordering promise and made it true, rather than dropping it — the grouping is the reason the bulk read exists, and a contract that holds only under 1000 ids is a trap for exactly the mailbox case the javadoc cites.

```java
// Each chunk is ordered, their concatenation is not: object ids ascend within a chunk
// and restart at the next one. All the rows of one object sit in a single chunk, so a
// stable sort on the object id restores the documented grouping while leaving the
// within-object ordering the query established untouched.
if (objectIds.size() > MAX_ITEMS_PER_IN_CLAUSE) {
  items.sort(Comparator.comparing(MetadataItemEntity::getObjectId));
}
```

`List.sort` is stable, and every row of one object falls in a single chunk, so this restores the global grouping without touching the per-object order fixed for finding 2.

On the missing coverage: I have **not** added a >1000-id case. It is the right test and I did not want to slip a container test with a thousand fixture rows into a backport unannounced — say the word and I will add it here, otherwise it belongs with the normalization work in the follow-up.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
Jihed525: All three taken, in `219ce5b6`.

- The constant now carries why it sits at the hard limit: ORA-01795 as `IdentityDAOImpl` uses it in this package, with a note that `ReactionStorage`'s 500 is a throughput choice for a hot path rather than a different reading of the dialect bound. Good catch that the two would otherwise look like a disagreement.
- `@param objectIds` now reads "the order they are given in does not matter, the result is ordered by object either way", which is what it meant.
- Moved to `IN (:objectIds)` to match every sibling in `MetadataItemEntity`.

component/core/src/main/java/io/meeds/social/category/service/CategoryLinkServiceImpl.java
Jihed525: Confirmed — and **not fixed as you specified**. I want to be explicit about that rather than let it read as done.

Everything in the finding checks out. `getLinkedIds(CategoryObject)` at `:63` normalizes; the bulk overload at `:68` does not; `isLinked`, `link` and `unlink` all do. `SpaceCategoryPlugin.getObject` resolves to `String.valueOf(space.getSpaceId())` and `ActivityCategoryPlugin.getObject` returns `new CategoryObject(activity.getMetadataObject())`, rewriting the type as well. `DefaultCategoryPlugin` is a passthrough, so only `space` and `activity` transform. Your second half is right too: the map is keyed by the stored id.

It also explains why the test passed: it uses `linkedSpace.getId()`, a form already equal to the stored id, so normalization is the identity there. The bug is real but latent — it needs a non-canonical id form to show.

**What I did instead: narrowed the contract on the interface** rather than implement normalization, because the correct fix is not small. A plugin can rewrite the *type*, so a correct bulk read must group normalized objects by resulting type, issue one read per type, and re-key the answer back to the caller ids. That is new behaviour, invented on the develop side, in a **backport whose whole review scope is diff-equivalence with what the PO validated** — the shape the ADR calls a backport rewrite masquerading as conflict resolution. `CategoryLinkService` now says ids must be in stored form, that `CategoryPluginService.getObject` is not applied, what the two plugins do, and that keys are the ids as given.

**Two reasons I think that is the right split, and one reason you may disagree:**

- No in-repo caller today, and `email-connector` EXO-89588 reads mail objects, which fall to the passthrough plugin — so nothing that consumes this is affected by the gap.
- The normalization belongs on `feature/ai-contribution` first, or the FB keeps the defective version and re-introduces it on the next backport out of that lane.

Against that: a published API narrower than its sibling is a papercut for whoever meets it next, and you flagged it as the first thing to take.

**Your call.** If you want it in this PR, say so and I will implement the per-type fan-out and re-key here, with a test covering a pretty-name id and an activity redirect. Otherwise it wants a ticket against the FB, and I would rather you raised it than me.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
```diff
@@ -166,6 +177,48 @@ public Map<String, Long> countMetadataItemsByMetadataTypeAndObjectGroupedByMetad
...
+                                         getEntityManager().createNamedQuery("SocMetadataItemEntity.getMetadataItemsByMetadataTypeAndObjectIds",
+                                                                             MetadataItemEntity.class);
+    query.setParameter(METADATA_TYPE, metadataType);
+    query.setParameter(OBJECT_TYPE, objectType);
+    List<MetadataItemEntity> items = new ArrayList<>();
+    for (int start = 0; start < objectIds.size(); start += MAX_ITEMS_PER_IN_CLAUSE) {
+      int end = Math.min(start + MAX_ITEMS_PER_IN_CLAUSE, objectIds.size());
+      query.setParameter(OBJECT_IDS, objectIds.subList(start, end));
+      items.addAll(query.getResultList());
+    }
```
azayati: 🟡 **Medium — a duplicate object id spanning two chunks returns its rows twice**

```java
for (int start = 0; start < objectIds.size(); start += MAX_ITEMS_PER_IN_CLAUSE) {
  int end = Math.min(start + MAX_ITEMS_PER_IN_CLAUSE, objectIds.size());
  query.setParameter(OBJECT_IDS, objectIds.subList(start, end));  // positional partition, no dedup
  items.addAll(query.getResultList());
}
```

`objectIds` is never de-duplicated. Under one chunk this is invisible — SQL `IN` collapses duplicate literals, so a repeated id yields its rows once. Past 1000 ids, a duplicate that straddles a chunk boundary is queried twice and its rows added twice, and `CategoryStorage`'s `Collectors.groupingBy` then hands that object a category list with every entry doubled — a page showing the same category twice on one row.

Raised now rather than in Round 1 because this code re-entered review with the fix commit; the org rule is that a fix commit gets no free pass.

**Fix:** de-duplicate once before chunking — `List<String> distinctIds = objectIds.stream().distinct().toList();` — and chunk that. It also makes the `> MAX_ITEMS_PER_IN_CLAUSE` sort guard exact rather than approximate. Worth folding into the >1000-id test when it lands.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
```diff
@@ -68,6 +70,15 @@ public class MetadataItemDAO extends GenericDAOJPAImpl<MetadataItemEntity, Long>
...
   private static final String OBJECT_ID            = "objectId";
 
+  private static final String OBJECT_IDS           = "objectIds";
+
+  // Oracle refuses an IN list past 1000 literals (ORA-01795); same bound, and same
+  // reason, as IdentityDAOImpl in this package.
+  // Oracle's hard limit on IN literals (ORA-01795), as IdentityDAOImpl uses in this
+  // package. ReactionStorage chunks the same table at 500; that is a throughput
+  // choice for a hot path, this one is the dialect bound itself.
+  private static final int    MAX_ITEMS_PER_IN_CLAUSE = 1000;
```
azayati: 🟢 **Nit — two small ones from the fix itself**

**1. The new comment was appended rather than replacing the old one**, so the constant now carries five lines saying the ORA-01795 thing twice:

```java
// Oracle refuses an IN list past 1000 literals (ORA-01795); same bound, and same
// reason, as IdentityDAOImpl in this package.
// Oracle's hard limit on IN literals (ORA-01795), as IdentityDAOImpl uses in this
// package. ReactionStorage chunks the same table at 500; that is a throughput
// choice for a hot path, this one is the dialect bound itself.
private static final int    MAX_ITEMS_PER_IN_CLAUSE = 1000;
```

**Fix:** drop the first two lines — the new version says everything they said and more.

**2. The narrowed contract landed on `CategoryLinkService` only.** `CategoryStorage.getLinkedIds`'s javadoc still frames the per-object call as simply "the wrong one for a list", and the test method's javadoc still says "That equivalence is the whole contract" — both now true only for stored-form ids. **Fix:** one clause in each pointing at the interface caveat, so the three places do not disagree.

azayati (commented):

New head `219ce5b6`. Scope re-confirmed from GitHub: still **9 files, +232/-0** — my local `origin/develop` was stale and made a two-dot diff look like 68 files; it is not, the PR is unchanged in scope.

| # | Round 1 finding | Status |
|---|---|---|
| F1 | 🟠 Bulk path skips plugin normalization | ➖ **Accepted — contract narrowed, implementation deferred**, decision explicitly escalated |
| F2 | 🟠 Ordering reversed vs the per-object read | ✅ **Fixed**, verified at head |
| F3 | 🟡 Ordering promise false past 1000 ids | ✅ **Fixed**, verified · test half ❌ **still open** (deferred, asked) |
| F4 | 🟢 Nits ×3 | ✅ **Fixed** (all three) |

**F2** — verified in the source, not from the commit message: the query now orders `mi.objectId ASC, COALESCE(mi.updatedDate, mi.createdDate) DESC, mi.createdDate DESC, mi.id DESC`, matching `getSortedMetadataItemsByMetadataTypeAndObject` exactly, and the comment names that query instead of "the query above". You re-derived the five-argument overload binding independently and got it right. The test now links a second category, asserts `size == 2` before the equality, and you mutation-verified it (`expected:<[22, 1]> but was:<[1, 22]>`) — that is what makes the assertion real rather than decorative.

**F3** — the stable-sort reasoning holds, and I checked both halves: `List.sort` is stable, and because the chunks partition `objectIds` positionally each object's rows do land in one chunk, so sorting on `objectId` restores grouping without disturbing the within-object order F2 fixed. Guarding it behind `size() > MAX_ITEMS_PER_IN_CLAUSE` is correct.

**F1 — my answer, since you asked for it: accept the split.** Your reasoning holds on the facts. A correct fix needs per-type fan-out and re-keying, which is new behaviour invented on the develop side inside a backport whose review scope is diff-equivalence with what the PO validated; there is no affected caller today (mail objects fall to the passthrough `DefaultCategoryPlugin` — verified); and left only on develop it gets re-introduced by the next backport out of that lane. It wants a ticket against `feature/ai-contribution`, not a rewrite here. You were right to name it rather than let it read as done — and right that this is the reviewer's call, so it stays a live decision for the human approver rather than something this review closes.

**Still open, by agreement — the >1000-id test.** My answer: **yes, but not in this PR.** A thousand-row fixture in a backport is disproportionate. It belongs with the normalization ticket, and it should cover the duplicate-id case in the new finding below in the same test.

### Verified conform (Round 2)

- The fix commit touches exactly four files — no drive-by edits.
- The keep-both `MetadataItemDAO` conflict resolution from Round 1 still stands at the new head; both develop methods present and untouched.
- `IN (:objectIds)` now matches every sibling in `MetadataItemEntity`.
- The reworded `@param objectIds` javadoc is accurate about what it means.

### Classification

**N1**, unchanged — JPA entity plus a hand-written JPQL query on the metadata store. Human Architect/Developer validation required; **F1 is a live decision for that approver**, not something settled here.

component/core/src/main/java/io/meeds/social/category/service/CategoryLinkServiceImpl.java
azayati: ➖ **Accepted — contract narrowed, implementation deferred.** You asked for the call, so: I agree with the split.

Your reasoning holds on the facts, and I re-verified them at `219ce5b6`. A correct fix needs per-type fan-out and re-keying because a plugin can rewrite the type — that is new behaviour invented on the develop side, inside a backport whose review scope is diff-equivalence with what the PO validated. There is no affected caller today: mail objects fall to the passthrough `DefaultCategoryPlugin`. And implemented only here it would be re-introduced defective by the next backport out of that lane.

The narrowed javadoc is the right artifact for this PR and it is accurate — it names both plugins, what each rewrites, and that the keys are the ids as given.

Two things I would still ask for, neither in this PR: a ticket against `feature/ai-contribution` for the normalization, and that its test cover a pretty-name id and an activity redirect, as you proposed.

Flagging it as a live decision for the human approver rather than closing it — it is theirs to ratify, not mine. You were right to state it as not-fixed instead of letting it read as done.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/entity/MetadataItemEntity.java
azayati: ✅ **Fixed — verified in the source at `219ce5b6`**, not from the commit message.

The query now reads `ORDER BY mi.objectId ASC, COALESCE(mi.updatedDate, mi.createdDate) DESC, mi.createdDate DESC, mi.id DESC`, which matches `getSortedMetadataItemsByMetadataTypeAndObject` exactly, so one object's slice of the bulk result is the same list in the same order the per-object call returns. The comment above now names that query instead of "the query above".

You re-derived the five-argument overload binding independently and got it right — that is the part I would have expected to be taken on trust.

The test fix is the more valuable half: with the second link, `assertEquals(2, ...)` before the equality, and the mutation check (`expected:<[22, 1]> but was:<[1, 22]>`), the assertion can now actually fail. Nothing further from me here.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
azayati: ✅ **Fixed** for the ordering — verified, and I checked the reasoning rather than the claim.

`List.sort` is stable, and because the chunks partition `objectIds` positionally, every row of one object does land in a single chunk — so the sort on `objectId` restores the documented grouping without disturbing the within-object order finding 2 established. Guarding it behind `objectIds.size() > MAX_ITEMS_PER_IN_CLAUSE` is right: under one chunk the database has already done it. Keeping the promise and making it true was the better of the two options I offered.

❌ **Still open: the >1000-id test.** You asked, so — **yes, but not in this PR.** A thousand-row fixture in a backport is disproportionate to what it protects here. It belongs with the normalization ticket.

One thing to fold into it when it lands: I have raised a separate finding on this same loop — `objectIds` is never de-duplicated, so past 1000 ids a duplicate id straddling a chunk boundary is read twice and `groupingBy` doubles that object's category list. The chunking test is the natural home for that case too.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
azayati: ✅ **Fixed — all three, verified.** `IN (:objectIds)` now matches every sibling in `MetadataItemEntity`, the `@param` wording says what it meant, and the constant explains why 1000 here and 500 in `ReactionStorage` — the throughput-choice-versus-dialect-bound distinction is exactly the right framing and better than what I asked for.

One leftover, raised as a nit on the constant: the new comment was **appended** rather than replacing the old two lines, so it now says the ORA-01795 thing twice in five lines. Deleting the first two finishes the change.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
Jihed525: Confirmed and fixed in `b864ec59`.

Real, and I had to walk the interaction to see it: the `IN` collapse is exactly why this hides. Under one chunk a repeated id costs nothing, so no small test would ever show it; it only bites once the positional partition splits a duplicate pair across a boundary, and then `groupingBy` doubles that object's list — a row showing the same category twice.

```java
// SQL IN collapses duplicate literals, so a repeated id costs nothing inside one
// chunk - but the chunks partition this list positionally, and a duplicate that
// straddles a boundary is asked for twice and its rows added twice, which hands the
// caller that object's categories doubled. Distinct once, up front; it also makes
// the sort guard below exact rather than approximate.
List<String> distinctIds = objectIds.stream().distinct().toList();
```

The loop and the `> MAX_ITEMS_PER_IN_CLAUSE` guard both run off `distinctIds` now, so the guard is exact rather than approximate, as you noted.

Agreed it belongs in the >1000-id test when that lands — a duplicate straddling the boundary is the case worth writing, not just the boundary itself. And your framing is right that this earned a fresh look because the fix commit re-entered review; it is a defect the round-1 fix walked past rather than one it introduced.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
Jihed525: Both fixed in `b864ec59`. The first one was mine and worth owning.

**1.** I appended the rationale instead of replacing what was already there, so the constant carried the ORA-01795 point twice in five lines. My edit inserted above the existing comment without reading what it already said. Dropped the older two lines; the newer text says everything they did and the `ReactionStorage` distinction besides.

**2.** You are right that a caveat in one of three places is worse than none — it reads as though the other two were considered and found exempt. `CategoryStorage.getLinkedIds` now carries:

> It is not a drop-in replacement in one respect: the ids are used as given, where the single-object call puts its object through the category plugin first.

and the test javadoc's "that equivalence is the whole contract" now finishes "...it holds for ids in stored form — which is the form this test uses", which also stops the test from reading as proof of something broader than it checks.

Build green, `CategoryLinkServiceTest` 6/6.

component/core/src/test/java/io/meeds/social/category/service/CategoryLinkServiceTest.java
```diff
@@ -45,6 +47,64 @@
...
+
+  /**
+   * The bulk read answers a page of objects at once, and answers it exactly as the
+   * single-object read would have, object by object.
+   * <p>
+   * That equivalence is the whole contract: the bulk call exists only so a caller
+   * listing rows can stop asking once per row, and it is worth nothing if the two can
+   * drift apart. It holds for ids in stored form - the bulk call does not put its ids
+   * through the category plugin, as the caveat on CategoryLinkService#getLinkedIds
+   * spells out - which is the form this test uses. An object with no category is absent from the map rather than present
```
azayati: 🟢 **Nit — the javadoc edit left one line at 121 characters**

```
   * spells out - which is the form this test uses. An object with no category is absent from the map rather than present
```

The inserted sentence was spliced in without re-wrapping the paragraph, so this line is 121 chars where its neighbours in the same block are 74-86.

I found no checkstyle or formatter config in the repo to cite a hard limit against, so this is a consistency point rather than a rule violation — but 121 is one over the eXo formatter's usual 120, so it may trip a build formatter if one runs.

**Fix:** re-wrap the paragraph. Purely cosmetic — fold it into any later push rather than spinning a commit for it.

azayati (commented):

New head `b864ec59`. Scope still **9 files, +242/-0**. The fix commit touches three files, all of them ones the previous round pointed at — no drive-by edits.

| # | Round 2 finding | Status |
|---|---|---|
| R2-1 | 🟡 Duplicate id spanning two chunks returns rows twice | ✅ **Fixed**, verified |
| R2-2a | 🟢 Duplicated ORA-01795 comment | ✅ **Fixed** |
| R2-2b | 🟢 Caveat stated on the interface only | ✅ **Fixed** in both other places |
| R1-F1 | 🟠 Bulk path skips plugin normalization | ➖ **Accepted** — deferred by agreement, decision sits with the human approver |
| R1-F3b | ❌ >1000-id test | ➖ **Deferred by agreement** to the normalization ticket |

**R2-1** — verified, and the fix is better than what I asked for. `distinct()` is applied once up front and **both** the loop bounds and the sort guard now run off `distinctIds`, so the guard is exact rather than approximate — the second-order point in the finding was picked up rather than only the first. `Stream.distinct()` preserves encounter order, so the chunk partition stays deterministic. The comment explains why the defect hides below one chunk (`IN` collapses duplicate literals), which is the part a future reader would otherwise have to rediscover.

**R2-2a** — the older two lines are gone; the constant carries one five-line rationale instead of two overlapping ones.

**R2-2b** — both places fixed, and the test javadoc is the more careful of the two: "...it holds for ids in stored form — which is the form this test uses" stops the test reading as proof of something broader than it checks. A test whose doc overclaims is worse than one with no doc, and that instinct is the right one.

### Verified conform (Round 3)

- The `@link` in `CategoryStorage`'s new caveat resolves — `component/core` depends on `component/api`, and the erasure-style `getLinkedIds(String, java.util.List)` reference matches the interface method.
- The caveat is now consistent across all three places (interface, storage, test); none reads as exempt.
- Round 1 and 2 fixes all still stand at the new head; the keep-both `MetadataItemDAO` conflict resolution is intact.

### Two items carried forward — both by agreement, neither an AI-review finding

1. **F1 (plugin normalization).** The reasoning is grounded in the corpus, and I checked rather than took it: **ADR-0005** scopes a backport review to *"diff-equivalence with the validated mips state"* and names *"backport rewrites masquerading as conflict resolution"* as a specific risk that scope exists to catch. Documenting and ticketing rather than rewriting is what that ADR prescribes. Wants a ticket against `feature/ai-contribution`.
2. **The >1000-id test** belongs with that ticket, and should cover a duplicate straddling the chunk boundary, not just the boundary itself.

**From the AI review side, nothing is outstanding.** All findings from all three rounds are fixed or accepted with a documented rationale. What remains is one cosmetic nit (inline) and two decisions that are the human approver's.

### Norm capture — two candidates, checked against the corpus first

Searched before proposing; neither is in the KB today (closest hits were unrelated per-domain pitfalls):

1. **A bulk read added beside a per-object read owes exact equivalence for one object** — same items, same order — and its equivalence test is vacuous with fewer than two items per object. This PR hit both halves: ordering diverged silently, and the test could not have failed. Proposed for `backend-spring.md` §7.
2. **`IN`-clause chunking must de-duplicate before partitioning positionally.** `IN` collapses duplicates inside one chunk, so the defect is invisible below the bound and doubles rows above it. Belongs beside the existing ORA-01795 guidance, with `IdentityDAOImpl` / `ReactionStorage` / this method as the three in-tree precedents — including that they disagree on the bound for legitimate reasons (dialect limit vs. hot-path throughput).

**Not proposed:** the backport-scope norm — ADR-0005 already holds it, verbatim.

### `/domain-doc` refresh

`domains/social.md` §10 needs the third additive `MetadataService` query — `(type, objectType, objectIds)` bulk read, DAO-level `IN` chunking at 1000, de-duplicated, ordered by object then most-recently-touched — plus a note that `CategoryLinkService`'s per-object reads normalize through `CategoryPluginService.getObject` while the bulk one deliberately does not.

### Classification

**N1**, unchanged — JPA entity plus hand-written JPQL on the metadata store. Human Architect/Developer validation required; **not to be auto-merged on this AI review alone.** The F1 decision is that approver's to ratify.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
azayati: ✅ **Fixed — verified at `b864ec59`.** And the fix is better than what I asked for.

`distinct()` is applied once up front, and **both** the loop bounds and the sort guard now run off `distinctIds` — so the guard is exact rather than approximate. That was the second-order point in the finding and it would have been easy to take only the first half. `Stream.distinct()` preserves encounter order, so the chunk partition stays deterministic.

The comment is the part that will age well: it records why the defect hides below one chunk (`IN` collapses duplicate literals), which is exactly what would otherwise have to be rediscovered by whoever next touches this loop.

Agreed on your framing too — this was a defect the round-1 fix walked past, not one it introduced. Nothing further here.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
azayati: ✅ **Fixed — both, verified at `b864ec59`.**

The older two lines are gone; the constant now carries one rationale instead of two overlapping ones.

On the caveat: the test javadoc is the more careful of the two edits. "...it holds for ids in stored form — which is the form this test uses" stops the test from reading as proof of something broader than it actually checks, which is a failure mode I did not name and you closed anyway. A test whose doc overclaims is worse than one with no doc.

One cosmetic leftover, raised as a separate nit: the inserted sentence was not re-wrapped, so `CategoryLinkServiceTest.java:59` sits at 121 chars against 74-86 for its neighbours. Not worth a commit of its own.

component/core/src/test/java/io/meeds/social/category/service/CategoryLinkServiceTest.java
Jihed525: Fixed in `ce005bdb` — and thank you for catching it, because I pushed that fix without replying here, which left it looking unanswered.

The paragraph is re-wrapped; the block now runs 81/85/33 where it had one line at 121:

```
   * spells out - which is the form this test uses. An object with no category is
   * absent from the map rather than present with an empty list, so a caller can tell
   * "none" from "not asked for".
```

You are right that there is no checkstyle or formatter config in the repo to cite — I looked as well and found none, so this is consistency with the surrounding block rather than a rule. Folded into its own small commit in the end rather than a later push, since there was no later push to fold it into.

Jihed525:
## Round #3 answered — `ce005bdb`

The one open item from Round #3 was the 121-char javadoc line; re-wrapped and replied inline. I had pushed that fix without commenting, so it read as outstanding when it was not — my omission, not your miss.

**State of the three rounds:**

| Round | Finding | Status |
|---|---|---|
| R1-F1 | 🟠 Bulk path skips plugin normalization | ➖ Deferred by agreement — **live decision for the human approver** |
| R1-F2 | 🟠 Ordering reversed vs the per-object read | ✅ Fixed `219ce5b6`, mutation-verified |
| R1-F3 | 🟡 Ordering promise false past 1000 ids | ✅ Fixed `219ce5b6` · test deferred by agreement |
| R1-F4 | 🟢 Nits ×3 | ✅ Fixed `219ce5b6` |
| R2-1 | 🟡 Duplicate id spanning two chunks | ✅ Fixed `b864ec59` |
| R2-2a/b | 🟢 Duplicated comment · caveat in one place only | ✅ Fixed `b864ec59` |
| R3 | 🟢 121-char javadoc line | ✅ Fixed `ce005bdb` |

**Two items carried forward, neither in this PR, both by agreement:**

1. A ticket against `feature/ai-contribution` for the F1 plugin normalization — per-type fan-out and re-keying, with a test covering a pretty-name id and an activity redirect. **I have not raised it**: creating Tribe tickets is the developer's call, so it is flagged for @Jihed525 rather than filed by me. Left only on develop it would be re-introduced defective by the next backport out of that lane.
2. The >1000-id chunking test belongs with that ticket, and should cover a duplicate straddling the boundary rather than only the boundary.

**Your two proposed norms both look right to me**, and neither is in the corpus today — a bulk read added beside a per-object read owing exact equivalence (with the ≥2-items-per-object rule that makes its test non-vacuous), and `IN`-clause chunking having to de-duplicate before a positional partition. This PR is evidence for both: the ordering diverged silently and the test could not have failed, and the duplicate defect was invisible below the chunk bound. Worth an `eng-standards` PR for `backend-spring.md` §7 — again @Jihed525's to open.

**Classification unchanged: N1.** JPA entity plus hand-written JPQL on the metadata store. This should not merge on the AI review alone, and the F1 deferral is the approver's to ratify, not something these rounds settled.

Re-requesting review.

component/core/src/main/java/org/exoplatform/social/core/jpa/storage/dao/jpa/MetadataItemDAO.java
Jihed525: **Correcting something we both got wrong** — pushed as `547e5222`.

eXo supports **MySQL and PostgreSQL**. Oracle is not a supported database, and the codebase agrees: there is no Oracle JDBC driver anywhere in the platform (no `ojdbc`, no `OracleDialect`, no `oracle.jdbc`). The `dbms="oracle"` attributes still scattered through old Liquibase changelogs are residue, not support.

So the ORA-01795 rationale explained a real mechanism with a constraint that **cannot apply**. You verified that comment and called the framing "better than what I asked for" — it was better *prose*, and wrong.

The chunking stays; only the reason changes:

```java
// A self-imposed bound, not a dialect limit: the platform supports MySQL and
// PostgreSQL, and neither refuses an IN list at any particular count. It is here
// because an unbounded list still has to be serialised into one statement and
// digested by the planner, and a page of rows can be arbitrarily long. 1000 matches
// IdentityDAOImpl in this package - which declares the same constant with no stated
// reason - so the two agree; ReactionStorage chunks the same table at 500.
```

Two things I want on the record because they are my errors, not yours:

1. **I invented the precedent.** `IdentityDAOImpl` declares `MAX_ITEMS_PER_IN_CLAUSE = 1000` with **no comment at all**. My "same bound, same reason" attributed my reasoning to code that states none. The comment now says that plainly.
2. **The ORA-01795 wording arrived with the source PR (#6008)** and I *extended* it in the round-2 nit fix instead of questioning it. A wrong rationale that reads well is harder to dislodge than no rationale, and I made it read better.

**This changes one of your proposed norms.** The `IN`-chunking entry should not be filed "beside the existing ORA-01795 guidance" — on MySQL/PostgreSQL there is no fixed maximum, so the norm is about *de-duplicating before a positional partition* and about *not justifying a bound with an unsupported dialect*, which is the sharper lesson. The de-dup defect itself is unaffected: it is a property of positional chunking, whatever the bound.

Same correction is owed on `feature/ai-contribution`, since that is where the wording originated.

No behaviour change. Build green, `CategoryLinkServiceTest` 6/6. Re-requesting review.

azayati (approved):

PR #6023, opened 2026-08-27
feat: Post Reporting EXO-89471

akhanfir:
Report a space-feed activity, comment or article for moderation: per-reporter metadata storage, derived canReport/hasReported flags, report drawer and menu entries (feed, comments, news article view), edit-reset across all reporters, space managers notification (onsite and email with target link), websocket sync across tabs.

PR #6039, opened 2026-08-31
feat: Post Reporting EXO-89471

akhanfir:
Report a space-feed activity, comment or article for moderation: per-reporter metadata storage, derived canReport/hasReported flags, report drawer and menu entries (feed, comments, news article view), edit-reset across all reporters on actual content edits only, space managers notification (onsite and email with target link), websocket sync across tabs.

Notes:
- The report notification is deliberately **Instantly-only** (no digest): moderation is time-sensitive (24h app-store expectation), a digest arriving days later serves no moderation purpose.
- Every regression pin in the delivery is mutation-verified (fix reverted → test fails → fix restored).

Replaces #6023 (retargeted from feature/mips to feature/devx).


component/core/src/main/java/io/meeds/social/report/listener/ReportActivityListener.java
```diff
@@ -0,0 +1,61 @@
...
+  private ActivityReportService activityReportService;
+
+  @PostConstruct
+  public void init() {
+    activityManager.addActivityEventListener(this);
+  }
+
+  @Override
+  public void updateActivity(ActivityLifeCycleEvent event) {
+    activityReportService.markReportsStale(event.getActivity());
```
boubaker: 🟠 **High — every `updateActivity` lifecycle broadcast flips reports stale, not only author content edits**

```java
@Override
public void updateActivity(ActivityLifeCycleEvent event) {
  activityReportService.markReportsStale(event.getActivity()); // fires on category link/unlink and unhide too
}
```

The `updateActivity` lifecycle event is broadcast by more than the author-edit path:

- [`CategoryLinkModifiedListener.java:100`](https://github.com/Meeds-io/social/blob/3d3c69e7bcf26f033b323b182c0c3508e8eb7eca/component/core/src/main/java/io/meeds/social/category/listener/CategoryLinkModifiedListener.java#L100) calls `activityManager.updateActivity(activity)` (broadcast=true) whenever a category is linked/unlinked on the activity. A space manager who adds a category to a reported post — a plausible triage gesture right after receiving the report notification — silently resets every reporter's "Reported" state, and each of them can re-report, producing duplicate notifications.
- [`ActivityRest.java:812`](https://github.com/Meeds-io/social/blob/3d3c69e7bcf26f033b323b182c0c3508e8eb7eca/component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRest.java#L812) (the unhide endpoint) fires the same broadcast with no content change.
- [`ActivitySharedActionsListener.java:56`](https://github.com/Meeds-io/social/blob/3d3c69e7bcf26f033b323b182c0c3508e8eb7eca/component/core/src/main/java/io/meeds/social/activity/listener/ActivitySharedActionsListener.java#L56) cascades `updateActivity(sharedactivity)` onto every shared copy when the original is edited — flipping the copies' reports as a side effect (arguably correct since their content mirrors the original, but it should be a decided behavior).

US03: the Reported state holds "until the post's author edits the post". Notably, content#902 protects articles from exactly this (only `CONTENT_AND_TITLE` resets, `CATEGORIES` does not — with a test pinning each direction); plain posts get the opposite behavior.

**Fix:** gate the flip on an actual content change rather than on the raw lifecycle event — e.g. an explicit service call from the edit path (the same explicit-delegation shape content#902 already uses), or compare title/body against `existingActivity` (already loaded in `ActivityManagerImpl.updateActivity`) before firing a dedicated signal. Please verify which of the remaining `updateActivity(…, true)` callers should legitimately reset (likes and pins already bypass with `broadcast=false`; `RelationshipPublisher`/`ProfileUpdatesPublisher` are personal-stream only, hence unreachable).

component/core/src/main/java/org/exoplatform/social/core/listeners/AbstractMetadataItemListener.java
```diff
@@ -81,10 +83,24 @@ protected void handleMetadataModification(MetadataItem metadataItem) { // NOSONA
...
+              // guard and stale sweep query the activity anchor
+              // (ActivityReportService), so moving them to the redirected
+              // content object would silently disable both. A redirected
+              // activity is cached under its redirected metadata object, so
+              // the selector-based eviction of
+              // handleMetadataModification(objectType, objectId) never matches
+              // it: evict it by key here so the derived canReport/hasReported
+              // flags are recomputed on the next read
+              cachedActivityStorage.clearActivityCached(objectId);
+              reindexActivity(objectId);
```
boubaker, line 97: 🟡 **Medium — reporter identity ids end up serialized in the activity ES document (plain posts/comments)**

Not this branch itself (a redirected activity's ES document is built from the redirected content object, so its reports are *not* indexed) — but for a **plain** activity, the pre-existing generic path below (`handleMetadataModification(objectType, objectId)` → `reindexActivity`) rebuilds the document through [`ActivityIndexingServiceConnector.addDocumentMetadata`](https://github.com/Meeds-io/social/blob/3d3c69e7bcf26f033b323b182c0c3508e8eb7eca/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ActivityIndexingServiceConnector.java#L220-L222), which serializes **all** metadata items of the object — the new `reports` items included, whose metadata name and `creatorId` are the reporter's identity id:

```java
private void addDocumentMetadata(DocumentWithMetadata document, MetadataObject metadataObject) {
  List<MetadataItem> metadataItems = metadataService.getMetadataItemsByObject(metadataObject);
  document.setMetadataItems(metadataItems); // includes "reports" items → reporter identity ids in the index
}
```

No user-facing path returns ES metadata today, so nothing leaks *now*. But the spec's posture is stronger than "no current path exposes it": report data should not exist anywhere a future read path could pick it up ("no audience rule, present or future, can expose one") — a future metadata-driven search facet would carry reporter identity. It is also asymmetric: articles' reports are not indexed, plain posts' are.

**Fix:** exclude `ActivityReportService.METADATA_TYPE_NAME` in `addDocumentMetadata` — the ES mirror of the `EntityBuilder.retrieveMetadataItems` exclusion — so the index carries no report data for any activity shape. (To verify on your side: whether favorites/tags rely on that same field in a way that constrains the filtering spot.)

component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
```diff
@@ -1781,6 +1792,29 @@ private static DataEntity getActivityStream(ExoSocialActivity activity, String r
     return as;
   }
 
+  private static boolean canReportActivity(ExoSocialActivity activity, Identity authentiatedUser) {
```
boubaker, line 1795: 🟢 **Nit — the read-path eligibility resolves the stream owner through a different path than the write-side guard**

The create guard, the notification recipients and the analytics dimension all resolve through `ActivityManager.getActivityStreamOwnerIdentity` (the single helper the spec names — the latter two via the event payload's `streamOwnerIdentityId`). This flag is a fourth resolution using the local `getStreamOwnerIdentity`, whose lookup is **user-identity-first** (`getOrCreateUserIdentity(streamOwner)` then space): a user whose username equals the space's pretty name would make `canReport` false while the write guard accepts the report. The drift is in the safe direction (entry hidden though reporting allowed) and the helper is pre-existing — but since the spec centralized specifically to avoid flag/guard drift, worth either routing both through the same resolution or a one-line comment stating why the divergence is acceptable here.

boubaker (commented):
## AI review — Round #1

Reviewed as one delivery with Meeds-io/content#902 and Meeds-io/analytics#435, against the eXip note (Post Reporting, activity 475326, spec at its round-3 close-out) and the board (project 8369). 2 findings are anchored inline (one 🟠, one 🟡), plus one anchored 🟢; the delivery-level finding and the nits are below.

---

🟡 **Medium — the spec needs a resync on three points before the mips integration PR**

1. **Content-backed activities (news) are now in scope** — the delivery ships dedicated machinery for them (`ReportActivityProcessor`, the move-exemption in `AbstractMetadataItemListener`, the `article` target type, content#902's menu entry), and board US01 says "a post or a news activity". The spec says the opposite in *Functional divergences confirmed* ("additionally excludes content-backed activities such as notes/news shared as a post", PO-confirmed 2026-08-21) and in *Ledgers > Out of scope*. The board wins for acceptance, but the PO reversal must be traced and the spec's Security/Data reasoning re-read under the new scope (it holds — reports anchor on the activity object, never the redirected one — but the spec must say so).
2. **The *Repositories touched & release order* table omits `content`** — content#902 is part of the delivery (menu entry + the explicit stale-flip on article content edits).
3. **"Neither repository needs to compile against the other" is falsified** — `AnalyticsActivityReportListener` is typed `Listener<ActivityReport, Long>` and imports `io.meeds.social.report.model.ActivityReport`, so analytics now requires a social version carrying that class (see the note on analytics#435). The code choice is reasonable (analytics already builds on social's BOM); the spec property is not.

**Fix:** run `/resync-spec` on the note before the `feature/mips` integration PR; record the news-in-scope PO decision.

---

🟢 **Nit — a few small ones**

- `canReport`/`hasReported` are serialized as Strings (`"true"`) while the sibling `canPin` on the same entity is a boolean — every consumer must remember to string-compare (`activity.canReport === 'true'`). One entity, two typings.
- US03 asks for a hover title ("You already reported this post") on the disabled entry; [`ActivityHeadMenu.vue:41`](https://github.com/Meeds-io/social/blob/3d3c69e7bcf26f033b323b182c0c3508e8eb7eca/webapp/src/main/webapp/vue-apps/activity-stream/components/activity/header/ActivityHeadMenu.vue#L41) implements it as `aria-label` only, which produces no visual tooltip on hover (a `title` attr or `v-tooltip` would).
- `PostReportPlugin.makeNotification` dereferences `parentActivity.getId()` without a null guard — an orphan comment (parent deleted between the report and the async notification) NPEs the notification thread.
- The mail template builder's `makeDigest` returns `false`: a manager whose Feed notifications are set to daily/weekly digest will never see report notifications at all. If deliberate (reports are Instantly-only), worth one line in the PR body.
- The drawer's 409 branch closes without setting `hasReported` locally — the menu entry stays enabled until the websocket round-trip refreshes it (minor race UX).

---

**Verified conform** (checked against the spec and the code at 3d3c69e, so reviewers don't need to re-litigate):

- **Guards**: 404/403(view)/403(author)/400(space-feed)/400(reason)/409(duplicate, last) in the spec's exact order, all server-side, reporter identity from `ConversationState` — pinned by `ActivityReportServiceImplTest.testReportActivityGuardsInContractOrder` and `ActivityReportRestTest.testExceptionToStatusContract`.
- **Privacy**: report items are excluded outright from the published payload for every caller, the reporter included (`EntityBuilder.retrieveMetadataItems`), and the spec-mandated regression test exists through the real entity-building path (`EntityBuilderReportsTest`). The websocket message carries no report data; hidden activities are skipped.
- **Data model**: one metadata item per reporter (`MetadataTypePlugin` with `shareable=false`, `allowMultipleItemsPerObject=false`); a stale item is reactivated in place, preserving the one-active-item-per-(reporter, object) invariant across edit cycles; the stale sweep flips every active report across all reporters and keeps reason/timestamp; the stale flip raises no notification and no event. All container- or unit-tested.
- **Redirected (news) activities**: reports anchor on the activity object; the move-exemption in `AbstractMetadataItemListener` is proven by a real-listener-chain container test **with a contrast metadata type** showing the exemption stays scoped to reports (`ActivityReportServiceTest`), including the by-key cache-eviction assertions.
- **Reaction eXIP coordination** (the spec's time-sensitive ledger item): both special-cases coexist in `EntityBuilder` (reports exclusion at L1369, reactions at L1399) — nothing was overwritten.
- **Notification**: recipients are the owning space's managers, reporter filtered out of their own notification (tested), one notification per report, Feed category (`activity_stream` group), onsite + email with the board's exact wording asserted in `PostReportMailBuilderTest`.
- **No schema change, no Liquibase, no new DAO** — generic metadata persistence end to end, as specced.
- **Frontend**: reuses `exo-drawer`, the existing `activity`/`comment` action extension points, and a singleton drawer guard; no catalogued shared component was reinvented.

CI (chat-level observation, not blocking): Sonar quality gate passed, 79.7% coverage on new code.

This is a carefully built PR: the container test through the real `metadataItem.created` listener chain with a contrast type is exactly the "execute the grammar" discipline the org asks for, the in-place comments in `AbstractMetadataItemListener` and `ActivityReportServiceImpl` explain *why* the anchor exemption exists rather than what the code does, and the spec's hardest requirements (payload exclusion, per-reporter anchoring, edit-reset across reporters) all landed with their mandated tests.

---

**Classification: N1** — the diff modifies the shared read-path publication method (`EntityBuilder.retrieveMetadataItems`) and the Kernel metadata listener chain (`AbstractMetadataItemListener`), both socle surfaces; this matches the spec's own N1 proposal. The PR targets a sandbox FB (`feature/devx`), so the classification applies at the `feature/mips` integration PR: its approver must be an Architect/Developer who knows the eXip is N1 — not an approval on AI review alone, and no auto-merge.

component/core/src/main/java/io/meeds/social/report/listener/ReportActivityListener.java
akhanfir: Fixed in c96477c21343 (amended into the single delivery commit), taking your "dedicated signal" option: `ActivityManagerImpl.updateActivity` now computes a `contentChanged` flag from a title/body comparison against the already-loaded `existingActivity` (null = kept, per the storage semantics), carries it on `ActivityLifeCycleEvent` (additive field, defaults to true so unknown broadcast paths keep the historical behavior), and `ReportActivityListener` flips reports stale only when it is true.

Verified against your three call sites: category link/unlink and unhide pass the reloaded activity with identical title/body → `contentChanged=false`, no reset; the shared-copies cascade copies the original's new content onto the copies → they reset exactly when their mirrored content actually changed, which we take as the decided behavior. Pinned at two levels, both mutation-verified: `ReportActivityListenerTest.testContentNeutralUpdatesNeverFlipReportsStale` (the gate) and `ActivityReportServiceTest.testUpdateActivityLifecycleCarriesContentChangeFlag` (the real `updateActivity` path, exercising the exact reload-then-update shape `CategoryLinkModifiedListener` uses).

component/core/src/main/java/org/exoplatform/social/core/listeners/AbstractMetadataItemListener.java
akhanfir: Fixed in c96477c21343: `addDocumentMetadata` now filters out `reports` items — the ES mirror of the `EntityBuilder.retrieveMetadataItems` exclusion — so the index carries no report data for any activity shape. Checked your caveat: favorites/tags keep flowing into the document untouched, asserted in the new `ActivityIndexingServiceConnectorTest.testReportMetadataItemsNeverIndexed` (reports-absent + favorites-still-present through the real `create()` path, mutation-verified).

component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
akhanfir: Addressed in c96477c21343 with the comment option: `canReportActivity` now states why the local user-first helper is acceptable here — the only divergence (a username equal to a space pretty name) hides the entry while the write guard stays the authority, i.e. the safe direction. Routing the flag through `getActivityStreamOwnerIdentity` would have introduced a second resolution style next to this payload's other stream properties, so documenting the boundary seemed the lesser drift.

akhanfir:
## Review fixes — all Round #1 findings addressed at c96477c21343 (amended into the single delivery commit)

| Finding | Status |
|---|---|
| 🟠 Lifecycle reset on content-neutral updates | ✅ Fixed — `contentChanged` computed in `ActivityManagerImpl`, carried on `ActivityLifeCycleEvent` (additive, default true), gated in `ReportActivityListener`; pinned at both levels, mutation-verified (details in the thread) |
| 🟡 Reporter ids in the ES document | ✅ Fixed — `reports` filtered in `addDocumentMetadata`, pinned through the real `create()` path (thread) |
| 🟡 Spec resync (news scope, repositories table, analytics compile-dep wording) | ⏳ Pending — needs the PO's news-in-scope arbitration first; will be run as `/resync-spec` on note 50533 before the `feature/mips` integration PR |
| 🟢 String vs boolean flag typing | ✅ Fixed — `canReport`/`hasReported` now serialized as booleans like `canPin`; all consumers (social + content#902) switched to truthy checks |
| 🟢 US03 hover title | ✅ Fixed — `:title` added on the disabled entry (activity + comment menus), `aria-label` kept |
| 🟢 NPE on orphan comment in `PostReportPlugin` | ✅ Fixed — null-parent guard returns null (skips the notification), the LikePlugin-precedented shape |
| 🟢 Digest returns false | ➖ Deliberate — reports are Instantly-only: moderation is time-sensitive (the 24h app-store expectation in the eXip rationale), a digest arriving days later serves no moderation purpose. Now stated in the PR description |
| 🟢 Drawer 409 leaves the menu enabled | ✅ Fixed — the 409 branch dispatches `activity-reported` so menus flip locally without waiting for the websocket round-trip |

All report suites green after the fixes (`ActivityReportServiceTest` ×4, `ReportActivityListenerTest` ×3, `ActivityIndexingServiceConnectorTest` ×5, `PostReportMailBuilderTest` ×3, `EntityBuilderReportsTest`, `ActivityReportRestTest`, `ReportWebSocketBroadcastListenerTest`); the three new pins were each reverted-red then restored-green.

component/core/src/main/java/io/meeds/social/report/listener/ReportActivityListener.java
boubaker: ✅ Fixed in c96477c, verified: `contentChanged` is computed in `ActivityManagerImpl.updateActivity` against the stored activity (a null title/body counts as "kept as is"), and the listener gates on `event.isContentChanged()`. The unit test covers both directions and the container test captures the real broadcast for the exact reload-then-update shape `CategoryLinkModifiedListener` and the unhide endpoint use. The comment-edit path (`saveComment`, now line 568) keeps the default `contentChanged=true`, which is correct — a comment update there is always a content edit. Only the two gated sites plus that one broadcast the update lifecycle at head.

component/core/src/main/java/org/exoplatform/social/core/listeners/AbstractMetadataItemListener.java
boubaker: ✅ Fixed in c96477c, verified: `addDocumentMetadata` filters out the reports type, and `ActivityIndexingServiceConnectorTest.testReportMetadataItemsNeverIndexed` pins the exclusion with a favorites item as contrast (other types keep being indexed).

component/service/src/main/java/org/exoplatform/social/rest/api/EntityBuilder.java
boubaker: ✅ Addressed in c96477c as the documented-divergence option: the in-place comment states the read-path/user-first resolution, the single divergence case, and that the write guard stays the authority. Nothing further needed.

boubaker (commented):
## AI review — Round #2 (final)

Re-review scoped to `3d3c69e..c96477c` plus verification of every Round #1 finding, at head `c96477c21343b72518486881910ea80ca5888bc6`.

| Round #1 finding | Status |
|---|---|
| 🟠 Reports flipped stale on non-content updates (category link/unlink, unhide) | ✅ **Fixed** — `contentChanged` computed in `ActivityManagerImpl.updateActivity` against the stored activity (null title/body = "kept as is"), listener gates on `event.isContentChanged()`, default `true` preserves historical behavior for every other listener. Verified: unit test in both directions, plus a container test capturing the real broadcast for the exact reload-then-update shape the category listener and unhide use. The comment-edit broadcast (`saveComment`) correctly keeps the default. |
| 🟡 Reporter identity ids serialized into the activity ES document | ✅ **Fixed** — `addDocumentMetadata` filters the reports type (ES mirror of the `EntityBuilder` exclusion); pinned by `testReportMetadataItemsNeverIndexed` with a favorites contrast item. |
| 🟡 Spec resync (news in scope vs *Functional divergences confirmed*; `content` missing from the repos table; "no compile coupling with analytics" falsified) | ❌ **Still open — not a code change.** To close via a `/resync-spec` of the eXip note (and the news-in-scope PO trace) before the `feature/mips` integration PR. |
| 🟢 Read-path stream-owner resolution divergence | ➖ **Accepted as documented** — in-place comment states the divergence, its single trigger case, and that the write guard stays the authority. |
| 🟢 Nits (string-typed flags, missing hover tooltip, `parentActivity` NPE, 409 local sync) | ✅ **Fixed** — flags are native booleans end-to-end (backend, stream menus, extensions, content), `:title` added on both menus, orphan-comment null guard returns null cleanly, the 409 branch now dispatches `activity-reported`. The digest-`false` question (reports invisible to digest-mode managers) stays a deliberate-behavior note, fine as is. |

No new findings on the incremental diff. The fix quality is worth noting: the `contentChanged` flag defaults to `true` so no existing listener changes behavior, the container test exercises the real lifecycle broadcast rather than mocking the event, and the boolean-serialization change was propagated to every consumer including content#902 in the same round.

**All code findings from Round #1 are resolved — nothing outstanding from the AI review side on this PR.** The one open item is the spec resync above, which lives on the eXip note, not in this diff. Reminder for the `feature/mips` integration PR: it carries the delivery's `Knowledge:` line (eng-standards PR or `none — <reason>`).

**Classification: N1** (unchanged — shared read-path publication method, Kernel metadata listener chain, and now the activity lifecycle event contract). The `feature/mips` integration PR's approver must be an Architect/Developer who knows the eXip is N1 — no auto-merge on AI review alone.

akhanfir:
@boubaker — the spec resync your Round #2 left open is done: [note 50533](https://community.exoplatform.com/portal/s/273/notes/50533) updated on 2026-08-31 (round-trip hash-verified, Rationale and §1 Functional byte-identical).

What changed, tracked in the note's new **Resync — substantive changes** journal:
- **Content-backed (news) activities brought into scope** on PO confirmation (2026-08-31) — scoped to the **hosting activity**: the standalone news detail view (article reached outside the stream) stays without a Report entry, now recorded as an explicit *Out of scope* decision. §Existing Features, §Functional divergences, §Access (two eligibility rules, 400 = personal-stream only) and §Data and persistence (the anchor + move-exemption + by-key eviction mechanics, plus a new rejected alternative for the redirected-object anchor) updated accordingly.
- **`content` added to *Repositories touched*** and the release-order paragraph corrected: social first — analytics' typed `ActivityReport` payload and content's report-Service injection both need the social artifact; §Integrations no longer claims "no compile-time dependency in either direction".
- **Edit-reset narrowed to actual content edits** (the additive `contentChanged` lifecycle flag + content's explicit delegation) recorded in §Upgradability and §Services & processing; the **ES-index exclusion** mirrored in §Security.

With that, every item from your two rounds is closed. Remaining for the `feature/mips` integration PR: its `Knowledge:` line (the eng-standards domain-doc refresh is the next step on our side).

PR #6041, opened 2026-08-31
eXIP Deliver: eXIP 7.3.0.15: Account Deactivation Request - EXO-89834

mkrout:


component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRest.java
```diff
@@ -1379,8 +1396,12 @@ public Response deleteUserById(
     if (identity == null) {
       throw new WebApplicationException(Response.Status.BAD_REQUEST);
     }
+    // a deleted account is firstly deactivated: an account deleted by an
+    // admin while still active must never keep surfacing as enabled
+    identityManager.processEnabledIdentity(id, false);
```
mkrout, line 1416: 🟡 **Medium — `deleteUserById` enforces the LDAP-sync and self-deletion rules only in the UI**

EXO-89816 ("Disable option to remove a user when synced using LDAP") is implemented purely in `UsersManagementItemMenu.vue` (disabled item + tooltip), and the "can't delete your own account" rule purely as a client-side dialog. A direct REST call by an admin still deletes an LDAP-synchronized user — the next directory sync recreates the IDM user while the social identity stays hard-deleted with a `deleted-`-prefixed email, an inconsistent state with no UI to repair — or deletes the admin's own logged-in account. The self-deactivation path does re-check `isInternalStore()` server-side (`AccountDeactivationService.isDeactivationAllowed`); the admin path should hold itself to the same standard (org norm: checks live server-side, the UI only mirrors them).

**Fix:** in `deleteUserById`, reject external-store users (`!user.isInternalStore()`) and `id.equals(currentUser)` with 403/400 before deleting.

component/web/src/main/java/io/meeds/social/handler/ProfileAccessHandler.java
```diff
@@ -0,0 +1,88 @@
...
+
+  @Override
+  public boolean execute(ControllerContext controllerContext) throws Exception {
+    String path = controllerContext.getParameter(REQUEST_PATH);
+    String profileOwner = StringUtils.isBlank(path) ? null : path.split("/")[0];
+    String username = controllerContext.getRequest().getRemoteUser();
+    if (StringUtils.isBlank(profileOwner) || StringUtils.equals(profileOwner, username)) {
+      return false;
+    }
+    Identity identity = identityManager.getOrCreateUserIdentity(profileOwner);
```
mkrout, line 80: 🟡 **Medium — this handler hands anonymous visitors a username-enumeration oracle**

Before this change, an unauthenticated request to `/portal/dw/profile/<anything>` always got the same login redirect. Now the response differs: existing-and-enabled username → login flow; missing/disabled/deleted → `page-not-found` redirect. An unauthenticated client can enumerate valid usernames (and detect deactivations), and each probe costs an identity lookup. The anonymous branch is deliberate (`testRedirectsAnonymousViewerToPublicPageNotFound`), but pre-authentication the check protects nothing — the anonymous visitor cannot see any profile anyway.

**Fix:** when `getRemoteUser()` is null, `return false` and let the login flow answer uniformly.

component/core/src/main/java/io/meeds/social/security/service/AccountDeactivationService.java
```diff
@@ -0,0 +1,223 @@
...
+  /**
+   * The deactivation is offered only for accounts whose lifecycle belongs to
+   * the platform: externally synchronized accounts (LDAP) are excluded, and
+   * among the others only those created through the platform UI qualify. An
+   * account with no creation source predates the stamping of that attribute:
+   * it keeps the option rather than losing it on an unknown provenance.
+   */
+  @SneakyThrows
+  public boolean isDeactivationAllowed(String username) {
+    if (!securitySettingService.getRegistrationSetting().isAccountDeactivationEnabled()) {
```
mkrout: 🟡 **Medium — the security gate reads a node-local memoized `RegistrationSetting`: an admin's OFF toggle isn't honored on other cluster nodes**

`SecuritySettingService.getRegistrationSetting()` memoizes into a plain field, invalidated only in the JVM that executed the save; the modification events are broadcast on the node-local kernel `ListenerService`. The underlying `SettingService` reads *are* cluster-safe (`CacheSettingServiceImpl`'s ExoCache), but the memoized object bypasses them. This is a pre-existing pattern for registration type/groups (display-level) — the difference is that the spec explicitly promises "the admin option is re-checked server-side at confirmation time", and on a cluster node B keeps saying yes until restart. Note the inconsistency inside this very delivery: `EntityBuilder.anonymizeDeletedUser` reads the flags through `settingService` directly (cluster-correct), while this gate uses the memoized copy.

**Fix:** have `isDeactivationAllowed` / `isDeletionAllowed` call `isAccountDeactivationEnabled()` / `isAccountDeletionEnabled()` directly (they hit the cluster-invalidated setting cache), keeping the memoized object for display paths only.

component/notification/src/main/java/io/meeds/social/security/plugin/EmailOtpPlugin.java
```diff
@@ -131,110 +111,50 @@ public boolean canUse(String userName) {
...
-  private String getUserLang(String userName) throws Exception {
-    UserProfile userProfile = organizationService.getUserProfileHandler().findUserProfileByName(userName);
-    if (userProfile != null
-        && userProfile.getAttribute(Constants.USER_LANGUAGE) != null) {
-      return userProfile.getAttribute(Constants.USER_LANGUAGE);
-    } else {
-      return ResourceBundleService.DEFAULT_CROWDIN_LANGUAGE;
+    String cacheKey = getCacheKey(userName);
+    if (getOtpSendLockCache().get(cacheKey) != null) {
+      throw new IllegalStateException(String.format("OTP email already sent to user %s less than %s seconds ago",
```
mkrout, line 122: 🟡 **Medium — the new 60-second send-throttle silently breaks "Re-send" in the pre-existing email-change flow**

This throttle applies to the shared `email` method used by the email-change and API-key flows. The email-change drawer's `sendOtpCode` has **no catch** (`UserSettingSecurityEmailDrawer.vue:183-191`): a "Re-send" within 60 s now rejects unhandled, while the `finally` still flips the UI to the "email sent" state — the user is told nothing and no mail goes out. Mitigated by the code being non-consuming (the previous code stays valid for the TTL), but the existing flow regresses silently. The new deactivation drawer handles the error; the old drawer doesn't.

**Fix:** catch the 429 in `UserSettingSecurityEmailDrawer.sendOtpCode` and show a "wait a minute" message (in both drawers); alternatively return 204 idempotently when throttled, since the previously sent code is still valid.

mkrout (commented):
## AI review — Round #1

Reviewed as part of the cross-repo delivery **eXIP 7.3.0.15 — Account Deactivation Request (EXO-89834)**, against the Tech Spec ([Tribe note 50463](https://community.exoplatform.com/portal/s/273/notes/50463)), together with Meeds-io/portal#1314, Meeds-io/meeds#4277, Meeds-io/platform-ui#1001, Meeds-io/analytics#436 and exoplatform/platform-private-distributions#607. Every finding was verified against the source at head `986243db16` (local checkout at that commit).

**4 findings are posted as inline comments** (all 🟡 Medium). One 🟠 High delivery-level finding is posted on Meeds-io/meeds#4277 but concerns the flow this PR ships: the `/portal/account-deactivated` global-site page (Meeds-io/meeds#4276 / EXO-89345) is merged to `feature/experience` only — nothing in this delivery brings it to `feature/mips`, so the logout redirect issued by [`UserSettingDeactivateAccountDrawer.vue:141-142`](https://github.com/Meeds-io/social/blob/986243db1612eb15ac45d91951fda6626d5bc5b3/webapp/src/main/webapp/vue-apps/user-setting-security/components/UserSettingDeactivateAccountDrawer.vue#L141-L142) lands on a non-existent page on this feature branch.

### Spec drift — for `/resync-spec`, deviations rather than defects

The delivery is a visible superset/evolution of the spec; the spec must be resynced. Verified deviations: **(1)** deletion requests are stored as `SettingService` rows under `Context.USER` — the spec *rejected* this alternative as "not enumerable", which is factually wrong: `SettingService.getContextsByTypeAndScopeAndSettingName` exists in the portal API and `AccountDeletionService.collectPendingRequests` uses it correctly; the spec's dedicated `SOC_ACCOUNT_DELETION_REQUEST` table + Liquibase changeset were never built (a sound simplification — no schema change at all). **(2)** Property names differ from the spec's customer-facing configurability table: `exo.accountDeletion.delay.days` / `exo.accountDeletion.job.expression` (default 05:15) vs `social.accountDeletion.delayDays` / `social.accountDeletionJob.expression` (23:30). **(3)** A dedicated OTP purpose (`accountDeactivationEmail`, purpose-bound cache keys) replaces the spec's "reuse the `email` method" — a security improvement (cross-purpose code reuse is impossible, tested in both directions). **(4)** One admin notification plugin instead of the spec's two; the admin email does not distinguish deletion requests from plain deactivations. **(5)** Whole areas absent from the spec: anonymization option + per-language deleted-user labels, profile-access routing (EXO-89280), admin-side deletion UX (EXO-89699/89816/89697), the `creationSource ∈ {blank, "ui"}` eligibility rule (users provisioned via API/CSV never get the option — a product decision that deserves an explicit spec sentence), email prefixing on hard delete, and the new OTP send-throttle.

### 🟢 Nit — a few small ones

- Hardcoded `/portal/` context in [`ProfileAccessHandler.java:78-86`](https://github.com/Meeds-io/social/blob/986243db1612eb15ac45d91951fda6626d5bc5b3/component/web/src/main/java/io/meeds/social/handler/ProfileAccessHandler.java#L78-L86) and the new `MailTemplateProvider` builder — existing idiom in the latter, but the handler could resolve the container/context name.
- `Login_fr.properties` edited by hand — bundles are Crowdin-managed; the manual French rows risk being overwritten or duplicated at the next sync.
- `component/web/pom.xml` sets `exo.test.coverage.ratio` to 0.02 — understandable for a module getting its first tests, but worth a ticket so the bar doesn't stay there.
- [`BrandedEmailSender.getSenderFullEmail()`](https://github.com/Meeds-io/social/blob/986243db1612eb15ac45d91951fda6626d5bc5b3/component/core/src/main/java/io/meeds/social/core/mail/BrandedEmailSender.java#L153-L162) builds `"Name<email>"` without quoting (inherited from `EmailOtpPlugin`); a company name containing `,` or `(` produces an invalid From header — now that it's centralized, `new InternetAddress(email, name)` is a one-line hardening.
- The deactivation drawer sends the OTP immediately on open — reopening within 60 s trips the new throttle and shows "Error while sending the code… please retry", although the previous code is still valid; a dedicated "code already sent recently" message would be less alarming.

### Verified conform (checked and found correct — no need to re-litigate)

- **OTP security model**: single authoritative server-side validation on the state-changing POST; `OtpService.validateOtp` (verified at portal head) throws on blank code, unknown method, wrong code, and after 5 tentatives; codes never logged nor persisted; purpose-bound cache keys tested in both directions.
- **REST contract**: `@Secured("users")`, username only from `request.getRemoteUser()`, `IllegalStateException→403` / `IllegalAccessException→401` / 204, anonymous rejected — covered by MockMvc tests running the real `WebSecurityConfiguration` filter chain.
- **`@ContainerTransactional` self-invocation in `AccountDeletionService` is safe**: the aspect is AspectJ **compile-time woven** (`execution()` pointcut, `aspectj-maven-plugin` + `portal.component.common` aspect library active for every social jar module), so `this.prepareDeletion()` / `this.deleteAccount()` do get their own container lifecycle — the per-account failure isolation the javadoc promises is real.
- **Job idempotency & cancellation**: marker removed in its own committed step before deletion; re-enabled/missing/external users revoke the marker (also via `postSetEnabled` on admin re-activation); ordering (email → marker → disable → events → session invalidation) verified with `InOrder` tests; a marker-write failure aborts before any state change.
- **Session invalidation** per spec (`IdentityRegistry.unregister` + `ConversationRegistry.unregisterByUserId`); the logout redirect is sanitized by the real `RedirectUrlValidator` (see portal#1314).
- **controller.xml route placement** (meeds/ppd PRs) follows the terms-and-conditions/space-access interceptor precedent exactly; `ProfileAccessHandler` registered like `SpacePermanentLinkHandler`.
- **Cross-PR contracts**: event name `social.account.deactivation.requested` identical between the constant, the XML listener registration and analytics' compile-time reference; `analytics-listeners` already depends on `social-component-core`.
- **ETag additions are cheap**: `CacheSettingServiceImpl` caches misses (`NullSettingValue`), so `EntityBuilder.getAnonymizationEtagValue()` per request costs cache hits, not queries.
- **Notification hygiene**: the "account" group is hidden from non-admin settings (answers the spec's deferred PO question); receivers = `/platform/administrators` minus the requester; resolution failures are swallowed into the context; the digest deduplicates requesters.
- The admin confirm dialog's HTML message goes through `v-sanitized-html` — no XSS regression; the compile-scope test-jar in `component/web` follows the repo's established pattern.

### Merits

The refactoring of `EmailOtpPlugin` into the reusable `BrandedEmailSender` removes ~150 lines of duplicated mail plumbing while adding a clean layout/fragment split; the test investment is exemplary for a security feature (391-line service test with ordering assertions, MockMvc security tests, container tests for the mail builder); and several spec blind spots were caught and closed during implementation (purpose-bound OTP keys, the deactivated-avatar leak in `UserRest.getUserAvatar`, the admin notification group hidden from non-admins).

---

**Classification: N1** (socle: authentication state, OTP, platform access settings; irreversible destructive operation; cross-repo delivery — the spec's provisional "N3" label uses the scale inverted and should be corrected at resync). Per `ai-review-and-merge.md`, this PR requires validation by an Architect/Developer and must not be merged on AI review alone.

component/service/src/main/java/org/exoplatform/social/rest/impl/user/UserRest.java
mkrout: ✅ **Fixed** in `a89832ee` — the rules are now enforced server-side, before any state change: self-deletion is rejected with 400, the user is resolved with `UserStatus.ANY` (a disabled account stays deletable, a missing one gets 400), and an external-store (LDAP-synchronized) user is rejected with 403 — mirroring the `item.isInternal` rule the UI applies.

component/web/src/main/java/io/meeds/social/handler/ProfileAccessHandler.java
mkrout: ✅ **Fixed** in `a89832ee` — an anonymous request now falls through (`return false`) before any identity lookup, so every username gets the same uniform login flow and the enumeration oracle is gone. The hardcoded `/portal/` was also replaced by `request.getContextPath()`, and the anonymous test was replaced by `testLetsAnonymousViewerFallThroughToLoginFlow` (asserts no redirect and no identity lookup).

component/core/src/main/java/io/meeds/social/security/service/AccountDeactivationService.java
mkrout: ✅ **Fixed** in `a89832ee` — `isDeactivationAllowed`, `isDeletionAllowed` and the `requestDeactivation` deletion gate now call `securitySettingService.isAccountDeactivationEnabled()` / `isAccountDeletionEnabled()` directly (cluster-invalidated setting cache); the memoized `RegistrationSetting` stays display-only, with the contract documented in its javadoc on Meeds-io/portal#1314.

component/notification/src/main/java/io/meeds/social/security/plugin/EmailOtpPlugin.java
mkrout: ✅ **Fixed** in `a89832ee` — `OtpService.js` now surfaces the HTTP status (same idiom as `AccountDeactivationService.js`), and both drawers catch it: a 429 shows the new `UserSettings.security.otpAlreadySent` warning (the previously sent code stays valid for its TTL), any other failure shows a send-error message — the email-change drawer previously swallowed the rejection entirely. The server-side throttle itself is unchanged (non-consuming). This also covers the reopen-within-60s nit on the deactivation drawer.

mkrout:
## AI review — Round #2 (final)

Re-review scope: `986243db16..a89832ee4e`, verified against the source at head `a89832ee4e` (local checkout at that commit).

| Round #1 finding | Status |
|---|---|
| 🟡 `deleteUserById` enforces LDAP-sync / self-deletion rules only in the UI | ✅ **Fixed** (`a89832ee`) — self-deletion → 400, missing user → 400 with `UserStatus.ANY`, external store → 403, all before any state change |
| 🟡 `ProfileAccessHandler` username-enumeration oracle for anonymous visitors | ✅ **Fixed** (`a89832ee`) — anonymous falls through before any identity lookup; test replaced accordingly |
| 🟡 Security gate on the node-locally memoized `RegistrationSetting` | ✅ **Fixed** (`a89832ee`) — all three service checks now use the direct cluster-safe readers; javadoc contract added on portal#1314 |
| 🟡 OTP send-throttle silently breaking "Re-send" in the email-change drawer | ✅ **Fixed** (`a89832ee`) — both drawers catch the 429 with the dedicated `otpAlreadySent` warning; `OtpService.js` throws the status code |
| 🟢 Nits: hardcoded `/portal` in the handler, unquoted From header, drawer reopen throttle message | ✅ **Fixed** (`a89832ee`) — `getContextPath()`, `InternetAddress(email, name, UTF-8)`, dedicated warning |
| 🟢 Nits: manual `Login_fr` rows, `component/web` coverage ratio 0.02, `MailTemplateProvider` `/portal` idiom | ➖ **Not taken** — acceptable as-is (Crowdin risk and coverage bar noted for follow-up tickets) |

**Fix-range re-review**: no new findings. The new guards are placed before any state change; `findUserByName(id, UserStatus.ANY)` keeps disabled accounts deletable; the `InternetAddress` change correctly encodes the personal name (UTF-8); the handler's context-path resolution and the updated tests are sound.

The delivery-level 🟠 (missing `/portal/account-deactivated` page on `feature/mips`) is closed on Meeds-io/meeds#4277 — the EXO-89345 page, navigation node and upgrade plugin are now part of that PR and were re-verified (`access-permissions` = `Everyone`, so the logged-out user can land on it).

**All findings from previous rounds are resolved — nothing outstanding from the AI review side.** Still owed at close-out, outside this PR: the spec resync of Tribe note 50463 (drift list in Round #1, including the inverted N-label) and a `/domain-doc` refresh of social while the knowledge is fresh.

**Classification: N1** — unchanged; this PR requires validation by an Architect/Developer and must not be merged on AI review alone.

boubaker:

Re-review scope: head moved since the Round #2 close-out, `a89832ee4e` → `6a756b73ee` (branch recreated as `mips-89834`, rebasing the delivery onto the current `feature/mips`).

**The PR content is byte-identical to the state verified in Round #2.** Verified by tree comparison at both heads over every file the PR touches at either head: the only deltas are two hunks inherited from base commits already merged to `feature/mips` (the `EmojiAccessibility` JS module, EXO-87562, and the `max-width-fit` toolbar class, EXO-88493). The 19 files no longer in the diff (module `pom.xml` version alignments, EXO-88493/EXO-88511 UI files) are now inherited from the base at identical content — nothing was lost. Spot-checked at head `6a756b73ee`: the anonymous fall-through in `ProfileAccessHandler.execute` (Round #1's enumeration-oracle fix) is present.

| Previous rounds' findings | Status |
|---|---|
| All (Rounds #1–#2) | ✅ Remain fixed at `6a756b73ee` — content identical to the verified `a89832ee4e` state |

No new findings. Still owed at close-out, outside this PR: the spec resync of Tribe note 50463 (drift list in Round #1) and the `/domain-doc` refresh of social.

**Classification: N1** — unchanged; this PR requires validation by an Architect/Developer and must not be merged on AI review alone.

component/core/src/main/java/io/meeds/social/security/service/AccountDeletionService.java
```diff
@@ -0,0 +1,220 @@
...
+   * most once even without any cluster lock. The marker is also cleaned up
+   * for requests that can never execute anymore (revoked by an admin
+   * re-activation, user already gone, externally synchronized account or
+   * unreadable marker). Deliberately not re-checked: the admin deletion
+   * option — the request was confirmed by its user under an enabled policy,
+   * turning the option off only prevents new requests.
+   *
+   * @return true when the account deletion must proceed
+   */
+  @ContainerTransactional
```
boubaker: please avoid starting the transaction from service methods and start it inside the Job. You may in `processPendingDeletionRequests` use `RequestLifecycle.restartTransaction` to commit and restart a new transaction each User Deletion to ensure to not have a big transaction to commit after.

```suggestion
```

component/core/src/main/java/io/meeds/social/security/service/AccountDeletionService.java
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * Deletes the account the way the users administration does: the social
+   * identity is soft-deleted first, then the IDM user is removed without
+   * broadcast — this ordering prevents any concurrent by-username resolution
+   * from resurrecting the deleted identity while the IDM user still exists.
+   * The user content (activities, comments, documents...) is preserved by
+   * design and keeps being displayed.
+   */
+  @ContainerTransactional
```
boubaker: Same remark, please avoid `@ContainerTransaction` in Service layer and let transactions started/controlled outside of Service Behavior. 

```suggestion
```

component/core/src/main/java/io/meeds/social/security/service/AccountDeletionService.java
```diff
@@ -0,0 +1,220 @@
...
+    LOG.info("Account of user {} deleted following the deletion request confirmed more than {} day(s) ago",
+             username,
+             delayDays);
+  }
+
+  private void removeDeletionRequest(String username) {
+    settingService.remove(Context.USER.id(username), DELETION_REQUEST_SCOPE, DELETION_REQUEST_SETTING_NAME);
+  }
+
+  private void broadcastEvent(String username, String identityId) {
```
boubaker: It can be more pertinent to broadcast with the whole already deleted `Identity` IMO to let each service read the adequate property if needed from Profile, WDYT ?

```suggestion
  private void broadcastEvent(Identity identity) {
```

component/core/src/main/java/io/meeds/social/security/service/AccountDeletionService.java
mkrout: ✅ Done in `bae7c72f` — the service no longer starts any transaction: the job's `run()` owns the container lifecycle (`@ContainerTransactional`, same shape as `ContentPublicationJob`), and `processPendingDeletionRequests` calls `RequestLifeCycle.restartTransaction()` in a `finally` after each account (the `UserImportService` idiom), so each deletion is committed on its own and the batch never accumulates one big transaction.

component/core/src/main/java/io/meeds/social/security/service/AccountDeletionService.java
mkrout: ✅ Done in `bae7c72f` — `@ContainerTransactional` removed from all three service methods (`collectPendingRequests`, `prepareDeletion`, `deleteAccount`); the transaction is controlled by the job. One semantic note reflected in the javadocs: the request marker removal now commits atomically with the deletion of the same account (per-account transaction), which strengthens the previous "marker committed first" contract — a crash mid-account no longer loses the request, and the no-cluster-lock idempotency still holds since a replayed run finds no marker once the account's transaction is committed.

component/core/src/main/java/io/meeds/social/security/service/AccountDeletionService.java
mkrout: Agreed — done in `bae7c72f`: `broadcastEvent(Identity identity)` now broadcasts the whole identity (`source` = `identity.getRemoteId()`, `data` = the `Identity`). Since `hardDeleteIdentity` doesn't mutate the passed object, `deleteAccount` sets `deleted=true` / `enable=false` on it before broadcasting (the same post-delete flagging `UserRest#deleteUserById` does), so listeners really receive a deleted identity and can read what they need from the profile. Nothing consumes `social.account.deleted` yet, so the payload contract was still free to change.

boubaker (dismissed):

boubaker (approved):

PR #6063, opened 2026-09-03
feat: eXIP7.3.0.19 Post Reporting EXO-89953

akhanfir:
Integration of the Post Reporting eXip (eXIP7.3.0.19) from feature/devx — this repo's delivery commits cherry-picked in their original order (#6039, #6049, #6051, #6052, #6053, #6055), verified patch-identical.

Knowledge: eng-standards domain-doc PR to follow — will be referenced here before merge.

Classification: N1 — approver must be an Architect/Senior Developer (author ≠ approver); no auto-merge on AI review alone.


boubaker (commented):
## AI review — Round #1 (`feature/mips` integration)

Reviewed as **one delivery** with Meeds-io/content#909, Meeds-io/analytics#440 and exoplatform/data-upgrade#306, against the eXip note (activity 475326, resynced 2026-08-31) and the board (project 8369 — every task *Tested & Validated*). This is the integration PR where the whole eXip diff meets its classification and approver, so the scope is the full diff at `83763c6`, not only the 5 fix commits that never had an AI round on `feature/devx`.

**Patch identity with `feature/devx`** — verified: over the 52 touched files, the only differences between the devx head (`e3d4289`) and this head are other eXips' hunks (account deactivation, Portlets/Analytics keys); no line containing report-related code differs.

### Previous rounds (social#6039 Round #1–#2), re-verified at this head

| Finding | Status at `83763c6` |
|---|---|
| 🟠 Reports flipped stale on content-neutral updates | ✅ still fixed — `contentChanged` computed in `ActivityManagerImpl.updateActivity` (both sides come from the same processed `getActivity` read, so a reload-then-update compares like with like), gated in `ReportActivityListener`, default `true` on every other broadcaster |
| 🟡 Reporter ids in the ES document | ✅ still fixed — `addDocumentMetadata` filters the `reports` type, pinned with a favorites contrast item |
| 🟡 Spec resync (news in scope, `content` in the repos table, analytics compile-dep) | ✅ done — note 50533 *Resync — substantive changes* entry of 2026-08-31; the code matches the resynced text on all three points |
| 🟢 Read-path stream-owner resolution divergence | ➖ accepted as documented (in-place comment) |
| 🟢 Nits (boolean flags, hover title, orphan-comment NPE, 409 local sync) | ✅ still fixed |

### The 5 fix commits (no prior AI round)

- **#6049 placement** — `rank: 90` on both `activity`/`comment-action` extensions, Delete being 100: the entry sits just before Delete, which is the one functional rejection recorded on US01 (PO, 2026-09-02) and validated the same day.
- **#6051 intro sentence** — `activityStream.report.selectReason` above the radio group.
- **#6052 tooltip** — a `v-tooltip` whose activator is a wrapper `div`, because Vuetify's `.v-list-item--disabled` blocks pointer events on the item itself; `aria-label` kept. Same treatment on the comment menu.
- **#6053 email + onsite only** — `PostReportPlugin` removed from `ActivitySpaceWebNotificationPlugin`'s `notification.plugin.ids` (the *Unread activities* channel is a plugin-id list, so this is the right lever; the net diff of this PR never adds it there).
- **#6055 lighter disabled entry** — `text-disabled-color` on icon and label; the class exists in platform-ui `helpers.less` on `feature/mips` (L162), so no cross-repo ordering issue.

One finding is anchored inline (🟢); the cross-repo findings live on their own PRs (🟡 on data-upgrade#306, 🟡 on content#909).

---

**Verified conform** (so the approver does not need to re-litigate):

- **Guards** in the spec's exact order — 404 / 403 view / 403 author / 400 personal stream / 400 reason / 409 last — all server-side, reporter identity from `ConversationState`; `ActivityReportServiceImplTest.testReportActivityGuardsInContractOrder` and `ActivityReportRestTest.testExceptionToStatusContract` pin them.
- **Privacy** — report items excluded outright from the published payload for every caller (`EntityBuilder.retrieveMetadataItems` L1374) and from the ES document; only the derived `canReport`/`hasReported` booleans travel; the websocket message carries no report data.
- **Data model** — one metadata item per reporter (`shareable=false`, `allowMultipleItemsPerObject=false`, type id 89471), stale item reactivated in place, stale sweep across all reporters keeping reason/timestamp, no notification and no event on the flip.
- **Redirected (news) activities** — anchor on the activity object, move-exemption in `AbstractMetadataItemListener` (L86) with by-key cache eviction and reindex, `ReportActivityProcessor` at priority 90 (after the generic metadata processor that replaces the map); all container-tested through the real listener chain with a contrast type.
- **Notification** — recipients are the owning space's managers minus the reporter, one notification per report, `activity_stream` group, `Instantly` default, board wording asserted in `PostReportMailBuilderTest` (post, comment and article targets); the digest-`false` choice stays the documented Instantly-only decision.
- **Frontend** — `activity-reported` document event keeps every open menu (stream, comments drawer, article view in content) in sync without waiting for the websocket round-trip; the comment menu refetches on `activity-stream-activity-updateComment`, which `ActivityStreamUpdater` does emit with `(activityId, spaceId, commentId)`.
- **No schema change, no Liquibase, no new DAO**; Spring MVC resource under `/social/rest/activities/reports/{id}`, `@Secured("users")`, Swagger documented.

**Before merge** (not code): the body's `Knowledge:` line is still a placeholder — it must name the eng-standards PR (or `none — <reason>`) before this PR crosses into `feature/mips` (`dev-lifecycle.md` §3b step 5; the required check reads that field). That domain-doc refresh should also record the two new extension-point contracts this delivery introduces: the `disabled` / `disabledLabelKey` / `disabledTitleKey` fields on `activity`/`comment-action` extensions, and the `contentChanged` flag on `ActivityLifeCycleEvent`.

CI (chat-level, not blocking): Sonar quality gate passed here; the content and analytics PR builds are red, as expected until this PR's `7.3.x-mips-SNAPSHOT` artifact is published — release order is social first, exactly as the spec now states.

The integration is clean work: the cherry-picks are patch-identical, the two rounds' fixes survived the move, and the follow-up commits are small, single-purpose and each tied to a validated board task — the `#6053` fix in particular chose the one-line configuration lever rather than code.

**Classification: N1** — the diff modifies the shared read-path publication method (`EntityBuilder.retrieveMetadataItems`), the Kernel metadata listener chain (`AbstractMetadataItemListener`) and the activity lifecycle event contract, all socle surfaces; the delivery as a whole is N1 by max-severity aggregation. This PR targets `feature/mips`, a trusted lineage: its approver must be an Architect/Senior Developer who knows the eXip is N1 (author ≠ approver) — no auto-merge on AI review alone.

webapp/src/main/webapp/vue-apps/activity-stream/components/report/ActivityReportDrawer.vue
```diff
@@ -0,0 +1,147 @@
...
+    selectedReason: null,
+    reasons: [
+      'spam',
+      'fakeAccount',
+      'harassmentOrBullying',
+      'hateSpeechOrDiscrimination',
+      'violenceOrThreats',
+      'falseInformation',
+      'intellectualPropertyViolation',
+    ],
```
boubaker, line 85: 🟢 **Nit — the reasons list is now owned in two places**

```js
reasons: [
  'spam',
  'fakeAccount',
  // ... 7 keys, duplicated from ActivityReportService.DEFAULT_REASONS
],
```

`ActivityReportService.DEFAULT_REASONS` is the authority the 400 guard enforces; this array is a second copy the client must keep in step. The spec keeps the list fixed for this version, so no functional impact today — but the deferred admin-configurable list would turn any server-side change into silent 400s from a stale client copy.

**Fix:** at minimum a comment cross-referencing `ActivityReportService.DEFAULT_REASONS`; better, serve the keys from the reports resource (a `GET /social/rest/activities/reports/reasons`) so the drawer never carries its own copy.

akhanfir: Fixed in c9461a86 (dedicated fix commit on top of the rebased picks): the reasons array now carries a comment pinning the coupling to `ActivityReportService.DEFAULT_REASONS` and the 400 contract. Serving the keys from the reports resource (`GET .../reports/reasons`) is kept as the follow-up for when the list becomes admin-configurable, per your 'better' option.

boubaker (approved):

PR #6092, opened 2026-09-09
fix: Cover the web notification REST builder and the cleanup job in social-component-notification - EXO-90072 - eXIP7.3.0.22

azayati:
## What
The `feature/experience` build of social fails at the JaCoCo gate since #6083: `Rule violated for bundle social-component-notification: instructions covered ratio is 0.67, but expected minimum is 0.70`.

Removing the legacy digest tests (`*MailBuilderTest` digest cases, `SocialNotificationUtils` digest helpers) removed some of the best-covered code of the module, so the remaining, less covered code now weighs more in the ratio: 6857/10114 = 0.678.

## How
Two unit tests (Mockito, registered in `InitContainerTestSuite` like the other unit tests of the module) on classes the suite never executed:
- `WebNotificationRestEntityBuilderTest` — sender resolution (`from`, parameter keys, numeric identity fallback, unknown user), space resolution (by id, by pretty name, through the activity stream, deleted activity), list entity with its paging. 265 instructions, 0 → 100 %.
- `SpaceWebNotificationCleanupJobTest` — the job asks the service once with one limit date, whether or not anything was marked. 81 instructions, 0 → 100 %.

Module after: 117 tests green, ratio **0.712**, `All coverage checks have been met` (`mvn verify -Pcoverage -pl component/notification`).

No production code is changed.

## Note for a separate task
While writing the job test, the computation `keepAliveDays * DAY_IN_MS` (`int × int`) was found to overflow for the default 30 days, giving a limit date ~20 days in the future — the nightly job marks every unread space item as read, not only those older than 30 days. It is not touched here (out of the scope of this technical task); the test asserts only the glue behaviour.

### DIFF
--- a/component/notification/src/test/java/io/meeds/social/notification/job/SpaceWebNotificationCleanupJobTest.java
+++ b/component/notification/src/test/java/io/meeds/social/notification/job/SpaceWebNotificationCleanupJobTest.java
@@ -1,0 +1,82 @@
+/**
+ * This file is part of the Meeds project (https://meeds.io/).
+ *
+ * Copyright (C) 2020 - 2026 Meeds Association contact@meeds.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


azayati:
Closed: the commit will be cherry-picked and squashed with the cleanup commit on feature/experience directly.

PR #6102, opened 2026-09-15
feat: Allow to list pages in search whatever their content - EXO-89723

mkrout:
Prior to this, a page without a content block never appeared in the unified search, so a user looking for "Spaces" by name found nothing.     This is caused by the page index holding one document per content block and nothing for the page itself.                                           This commit fixes the problem by indexing every reachable page that carries no content block as its own document, by ranking name matches ahead of content matches in the search query, and by moving the index to page_content_v2 so existing installations reindex on upgrade.


component/core/src/main/java/io/meeds/social/cms/listener/PageSavedIndexingListener.java
```diff
@@ -0,0 +1,106 @@
...
+      // connector never indexes a draft, so there is nothing to queue for it
+      return;
+    }
+    // The broadcast instance is the caller's own, whose storage id isn't
+    // necessarily set when the page was just created: the stored page is
+    String storageId = resolveStorageId(pageKey);
+    if (StringUtils.isNotBlank(storageId)) {
+      indexingService.reindex(PageContentIndexingConnector.TYPE, storageId);
+    }
+  }
```
ahamdi, line 108: 🟡 **Medium — `PageSavedIndexingListener` doubles indexing-queue traffic on the hottest page-editing path**

This listener's own javadoc already notes the portal-level event is "broadcast by every path that stores a page, **the layout editor included**" — and that's exactly what happens: `io.meeds.layout.service.PageLayoutService` (the Layout addon's own service, `layout` repo) persists through the MOP `LayoutService`, whose `save(Page)` / `save(PageContext, Page)` / `save(PageContext)` overloads *all* broadcast `PAGE_UPDATED`/`PAGE_CREATED` (`LayoutServiceImpl.java:234-252` in the `portal` repo). `PageLayoutService.createPage` (L252-253), `updatePageLayout` (L364-365) and `restorePageLayout` (L399-400) call one of those overloads *and then* explicitly re-broadcast the addon's own `"layout.page.updated"` — the event `PageContentBlockIndexingListener` listens to. So for these three operations — `updatePageLayout` being the layout editor's own save, the single hottest write path in this whole feature — **both** listeners fire on the same save.

When the edited page carries live content blocks (the common case), the two listeners don't just redundantly agree: `PageContentBlockIndexingListener` directly reindexes the exact current block ids, while `PageSavedIndexingListener` reindexes the bare `storageId`, which `PageContentIndexingConnector.create()` resolves to `createPageDocument()`:

```java
List<CMSSetting> liveBlocks = findLiveContentBlockSettings(page);
if (!liveBlocks.isEmpty()) {
  // ...Queuing them here makes both paths converge on the right documents
  // one indexing cycle later.
  liveBlocks.forEach(setting -> pluginService.reindexContentBlock(setting.getType(), setting.getName()));
  ...
  return null;
}
```

(`PageContentIndexingConnector.java:377-390`) — which re-queues every one of those same live blocks a second time. Verified against `commons`' `QueueIndexingService.reindex()`: it unconditionally inserts a new `ES_INDEXING_QUEUE` row per call, no de-duplication against an already-pending identical `(connectorName, id)` entry. So every layout-editor save of a content-bearing page (and every `createPage`/`restorePageLayout`) queues each live block for reindexing twice. This is idempotent — no wrong data results — but it's needless load on the indexing queue and Elasticsearch, on the single most frequent write path this feature touches, not the one-time bootstrap/external-save scenario the surrounding comments focus on. (`cloneSection` and `updatePageLink` are genuinely exclusive to `PageSavedIndexingListener` — they call `layoutService.save(...)` but never re-broadcast `"layout.page.updated"` — so the listener isn't purely redundant, just uncomfortably wide.) No test in this PR (or the existing suite) pins the two-listeners-fire-together case, so this doubling isn't visible from the test suite alone.

**Fix:** either scope `PageSavedIndexingListener` down so it doesn't act when the save also fired (or will fire) `"layout.page.updated"`/`"layout.page.permissions.updated"` for the same page — which likely means closing the two genuine gaps (`cloneSection`, `updatePageLink`) with an explicit broadcast in `PageLayoutService` instead, so this Kernel listener can be scoped to true non-Layout saves only — or, if the duplication is judged cheap enough to accept, say so explicitly in the class javadoc (right now it explains *why* the listener must also cover the layout-editor path, but not that doing so doubles every live block's queue entry) and add a test asserting the two listeners' combined effect on one edit is still correct (even if wasteful), so a future change to either listener doesn't silently break that assumption.

ahamdi (commented):
## AI review — Round #1

Verified against the PR head commit `944cfdfb66c4bcbc8c4bedc3fa22007687328aee`, cross-checked against local checkouts of `social`, `layout`, `portal`, and `commons`, with the main finding below independently confirmed by a second, adversarial verification pass.

**Verified conform**
- Permission filtering (`buildPermissionFilter()`) is unchanged and applies identically to content-block and bare-page documents in both `search()` and `getById()` — no ACL regression.
- Every new/changed Kernel listener registration in `indexing-configuration.xml` uses an event name that is an exact string match to the constant it's meant to react to (`LayoutService.PAGE_CREATED`/`PAGE_UPDATED`, `UserPortalConfigService.SITE_TEMPLATE_INSTANTIATED`) — checked against the `portal` source directly.
- `SiteTemplateInstantiatedIndexingListener` reacts to an event verified to fire only after `UserPortalConfigService.createSiteFromTemplate` has saved the portal config, the pages, **and** the navigation (`UserPortalConfigService.java:277-288` in `portal`) — matches the javadoc's claim precisely.
- Draft-page/draft-site exclusion (`isDraftPage`, `isIndexableSite`) is applied consistently across every entry point that can produce or reconcile a bare page document.
- `getAllIds`'s block-ids-then-page-ids concatenation correctly resumes the page portion's offset across a batch boundary; the underlying `index_previous`/`index_current`/`reindexOnUpgrade` migration mechanism is the existing `ElasticIndexingServiceConnector` contract, used as designed.
- The three-tier `should`/`constant_score` query keeps the shared `filter` clause (permissions/space/favorites) applying to every tier, and the `matchedName` bypass in `buildResult` only affects which language's excerpt gate a result must pass, not who can see it.

The two-document-kind design (content block vs. bare page) is well thought through, the pagination-boundary and plugin-registration-order edge cases got real attention, and the test suite is unusually thorough for this kind of change (draft/site-type exclusion, batch-boundary pagination, tiered-query shape all have dedicated tests).

One finding below.

**Classification: N2** (shared unified-search component, list-shaped indexing/search paths) — no ACL, schema/Liquibase, or Kernel↔Spring trust-boundary trigger fires here. Needs validation by an Architect/Developer; not to be auto-merged on this AI review alone.

component/core/src/main/java/io/meeds/social/cms/listener/PageSavedIndexingListener.java
mkrout: Addressed in 2cbe724814, along the second path you outlined, kept inside `social`:

- `PageContentIndexingConnector.createPageDocument` now queues only the live blocks the index does **not** hold yet: one `PageContentSearchConnector.findIndexedDocumentIds(storageId)` lookup (the query the reconciliation listeners already run) filters out the blocks `PageContentBlockIndexingListener` is refreshing for the same edit. The upgrade self-heal keeps working — the reindex starts with a `DELETE_ALL`, so nothing is indexed and every live block gets queued — while a Layout-driven save no longer queues each block a second time. What remains of the overlap is the bare page operation itself, whose build stops at that lookup.
- `PageSavedIndexingListener`'s javadoc now states the overlap explicitly, its bounded cost, and why `PAGE_UPDATED` stays in scope (`cloneSection` and `updatePageLink` broadcast no `layout.page.*` event).
- Pinned by `shouldNotRequeueLiveContentBlocksTheIndexAlreadyHolds` (an already-indexed block is not re-queued while a missing one is), mutation-verified: dropping the filter fails it. 92 tests green on the six classes.

Closing the two Layout gaps with explicit broadcasts would be the cleaner end state, but it lives in the `layout` repo, so it is deliberately left out of this PR.

Verified on a local server: clean startup with the added constructor dependency (the connector is Kernel-instantiated), and the same tiered results as before.

ahamdi (approved):

PR #6106, opened 2026-09-17
fix: Backport ProConnect OpenID security fixes EXO-89168 EXO-89197 EXO-89198 EXO-89200 EXO-89202 EXO-89205 EXO-89222

rdenarie:
## Summary
Backport of the ProConnect/OpenID security-fix work from `feature/security-fix` to `develop`: 8 commits.

- EXO-89168: Make OpenID given_name/family_name claim names configurable
- EXO-89197: Validate OpenID nonce claim to prevent id_token replay
- EXO-89198: Increase entropy of OpenID state parameter
- EXO-89200: Validate state parameter on OpenID logout callback
- EXO-89202: URL-encode OpenID request parameters
- EXO-89205: Verify UserInfo response signature when JWT-signed
- EXO-89222: Support ES256 and HS256 in OpenID signing key resolution
- (new commit) Pin the OpenID login-callback state validation and the token-signature-rejection path with unit tests (EXO-89198 area)

This is **not** a purely mechanical backport anymore. Five picks (`707d2604`, `dec1dea8`, `ae7ad36a`, `3f8bbaa7`, `b398d2a5`) are byte-identical to their `feature/security-fix` counterpart. Two picks were amended during this PR's own review and now carry content the FB does not have:
- EXO-89200's pick (`5d6fe2bb`) adds a comment documenting that the logout-state check is intentionally detection-only, not a gate (round 1, F2).
- EXO-89222's pick (`74ff07f4`) was extended repeatedly during review: pinning accepted signing algorithms to what the provider's well-known document advertises (F3), making JWKS key parsing resilient to a single unparseable entry (F4), no longer caching that algorithm pin on a transient fetch failure (F11), making the two cached fields `volatile` (F12), catching the new signature-rejection exception alongside the other JWT validation failures (F13), fixing the id_token custom-claims filter to use the same configurable claim names as `OAuthUtils` (F16, a pre-existing gap exposed by EXO-89168), and adding a package-private test seam (`setWellKnownConfigurationForTest`) so F13 can be pinned without a real well-known-document fetch (F18).
- One commit is new and not a pick at all: `OpenIdProcessorImplTest` (`450c1dd4`) pinning the login-callback `state` mismatch/absence rejection (F5, partial) and the F13 signature-rejection session cleanup (F18).

None of this amended/added content exists on `feature/security-fix`, and none of it has run on any acceptance server — this PR, not the FB, is where it has been reviewed. `feature/security-fix` remains unaffected and continues to track `develop` via its nightly rebase.

No `pom.xml` touched, so no cross-branch dependency-version conflict.

Note: EXO-89168's commit (`707d2604`) also carries two unrelated frontend fixes folded into it on `feature/security-fix`: `LoginProviders.vue` (`oAuthEnabled`/`oAuthProviders` moved from `computed` to `methods`, since the shared `params` object is never Vue-observed) and `LoginForm.vue` (`values: null` declared in `data()`, so the object `created()` assigns to it is reactive).

## Classification
N1 — touches OpenID/OAuth authentication (nonce replay protection, state/CSRF validation, signature verification, signing-key resolution), and now also adds a new gate on the token-verification path (F3/F4) that this PR itself introduced and that has not been validated on any ACC.

## Knowledge
None — `component/oauth-auth` is out of the corpus's documented scope (`domains/social.md` §11).

## Test plan
- [x] `mvn -pl component/oauth-auth,webapp -am install` and `mvn -pl component/oauth-auth verify -Pcoverage`: BUILD SUCCESS, all coverage checks met
- [x] Full test suite run across dependencies: only pre-existing failure is `SpaceCategorySidebarPluginTest` (module `component/core`, untouched by this backport) — unrelated to the OpenID changes
- [x] Every fix added during review except F16 is pinned by a unit test, mutation-verified against the pre-fix code (F3, F4, F11, F13). F16 is not: the claim-name constants it fixes are `static final`, resolved once from a system property at class-load, and already equal the pre-fix hardcoded literals under the default config used in this test suite — a test running under that config could not distinguish the fixed code from the bug it fixes.
- [x] Functional validation on ACC for the six original picks (see review round 1); the content added during this PR's review (F3/F4 extensions, F11, F12, F13, F16) has **not** run on any ACC


azayati (commented):
## AI review — Round #1

Backport PR, N1. Scope: **diff-equivalence with the validated FB state plus full review of every adaptation** — and, since `develop` has never carried any of this, the content on its merits: it is the OpenID login and logout path. Findings carry a stable id (`F1`…) that will not change across rounds, and a kind: **[inherited]** (byte-identical to the FB content the PO validated — fixed on the FB, then re-backported) or **[backport-blocking]** (the pick or an adaptation broke it — fixed here).

**Equivalence: six picks identical, one carries an undeclared addition (F1).** The seven `(cherry picked from …)` trailers name commits that are **no longer on `feature/security-fix`**: the FB is force-pushed every night at 23:49 UTC by `exo-swf` (the daily rebase — `a21d8ce8 → b7f8f42a` on 2026-09-17), so the SHAs of 2026-09-16 the picks were taken from are dangling. Checked against the FB's **current** commits, matched by title: `707d2604`, `dec1dea8`, `ae7ad36a`, `bddbabe0`, `71a9edeb`, `3c494570` have the same `patch-id` as `a9511793`, `67a47b80`, `98b19899`, `829e10c9`, `c1a506fd`, `bea8c37b`. `f2d228fb` differs from `b7f8f42a` by exactly one thing, see F1. Merge-base is `develop`'s tip `2f1b85bd`; the FB holds nothing else but its version-bump commit; no `pom.xml` in the diff; 0 `Claude-Session:` URLs.

---

🟡 **F1 [backport-blocking — a declaration, not the code] — the EXO-89222 pick carries a 73-line test that exists nowhere on the FB**

[`RemoteJwkSigningKeyResolverTest.java`](https://github.com/Meeds-io/social/blob/f2d228fb13d762471202c06a0e255396be8635f9/component/oauth-auth/src/test/java/io/meeds/oauth/openid/RemoteJwkSigningKeyResolverTest.java) — new file in `f2d228fb`

The trailer's source `a21d8ce8` touches two files (`OpenIdProcessorImpl.java` +1/−1, `RemoteJwkSigningKeyResolver.java` +52/−7); so does the FB's current `b7f8f42a`; the FB has no `RemoteJwkSigningKeyResolverTest.java` at any path (GitHub contents API: 404 on `feature/security-fix`). The pick adds it. The body says the seven commits were "cherry-picked with `-x` in their original order" and declares no adaptation. A test-only addition is welcome — it is the only test this PR has (F5) — but on a backport whose review *is* the equivalence declaration, an addition the FB does not have must be declared, and it means the FB ships this code **without** that test.

**Fix:** one sentence in the body ("`f2d228fb` adds `RemoteJwkSigningKeyResolverTest`, not on the FB") and the same test pushed to `feature/security-fix`, so the next nightly rebase does not leave the two lineages apart on a security module.

---

🟡 **F2 [inherited] — the logout `state` is compared, then ignored: a mismatch logs a warning and the logout proceeds exactly as a match would**

[`OpenIdProcessorImpl.java:627-639`](https://github.com/Meeds-io/social/blob/f2d228fb13d762471202c06a0e255396be8635f9/component/oauth-auth/src/main/java/io/meeds/oauth/openid/OpenIdProcessorImpl.java#L627-L639)

```java
if (logoutStateFromSession != null) {
  session.removeAttribute(OAuthConstants.ATTRIBUTE_LOGOUT_STATE);
  String logoutStateFromRequest = request.getParameter(OAuthConstants.STATE_PARAMETER);
  if (!logoutStateFromSession.equals(logoutStateFromRequest)) {
    log.warn("Validation of state parameter failed on OpenId logout callback. ...");
  }
  return false;   // same answer either way
}
```

`OAuthLogoutFilter.doFilter` (`:51-64`) calls `processLogout` on `/portal/logout` while the user is still authenticated; `false` means "not handled", and the chain continues into the portal's own logout. So on the return leg from the provider's end-session endpoint, a callback with the wrong `state`, or with none, ends the local session exactly like the genuine one. EXO-89200's stated purpose — "verifying that the returned value matches … lets the client confirm that this incoming request is a genuine continuation of a logout it initiated, rather than an unsolicited request crafted by a third party" — is met by the *sending* half (the `state` is now on the end-session URL, `:666-670`) and not by the receiving half: the confirmation is computed and discarded. The exposure is small (a forged `/portal/logout` hit logs the user out with or without this code), which is why this is 🟡 and not 🟠; but a check whose two branches do the same thing is either dead code or a missing decision.

**Fix (a decision for the Architect):** on mismatch, treat the request as *not* the return leg — do not consume the stored state and let the initiation branch below run again (the user is re-sent to the end-session endpoint), or refuse the callback with a redirect to the portal home; and pin it (a `processLogout` test with a stale `state` asserting the session survives). If detection-only is the intent, say so in the Javadoc and the commit title, which currently promises validation.

---

🟡 **F3 [inherited] — the signature algorithm is taken from the token's own header, and HS256 is honoured with the client secret for every provider**

[`RemoteJwkSigningKeyResolver.java:89-96`](https://github.com/Meeds-io/social/blob/f2d228fb13d762471202c06a0e255396be8635f9/component/oauth-auth/src/main/java/io/meeds/oauth/openid/RemoteJwkSigningKeyResolver.java#L89-L96)

```java
String algorithm = header.getAlgorithm();
if (algorithm != null && algorithm.startsWith("HS")) {
  return new SecretKeySpec(clientSecret.getBytes(StandardCharsets.UTF_8), "Hmac" + algorithm.replace("HS", "SHA"));
}
return getKey(header.getKeyId());
```

This is **not** the classic RS→HS confusion (the HMAC key is the client secret, never the public key, so a forger needs the secret), and `alg: none` is refused by jjwt's `parseClaimsJws`. What it does is widen what a token may claim: an IdP registered as RS256/ES256 now also has its id_tokens and its signed UserInfo accepted under HS256/384/512, keyed on `client_secret`. Two consequences. (1) The client secret becomes a **signing key for authentication**, not only a credential to the token endpoint: a leaked or guessable secret forges an id_token for any user, where with RS256 alone it could not. (2) jjwt `0.13.0` (the version every `maven-depmgt-pom` line pins — verified in the BOM) enforces the HMAC minimum key size on verification too (`DefaultMacAlgorithm.assertKey`, `WeakKeyException` under 256 bits for HS256): a provider that really signs HS256 with a secret shorter than 32 bytes fails **closed** at login, and the resolver's own test uses a six-byte secret (`"s3cr3t"`) that verification would reject. (`clientSecret` itself cannot be blank: `:144` refuses to start the provider without one, so the HS branch never sees a null.)

**Fix:** pin the accepted algorithms per provider rather than per token — read `id_token_signing_alg_values_supported` from the well-known document already parsed in `readWellKnownConfiguration`, or a `signatureAlgorithms` init-param defaulting to the asymmetric ones — and only then honour `HS*`. Pin with a token whose header says `HS256` against a provider that advertises RS256 only, asserting rejection.

---

🟡 **F4 [inherited] — one unsupported EC curve in the provider's JWKS breaks every key, and every login, until the document changes**

[`RemoteJwkSigningKeyResolver.java:164-170`](https://github.com/Meeds-io/social/blob/f2d228fb13d762471202c06a0e255396be8635f9/component/oauth-auth/src/main/java/io/meeds/oauth/openid/RemoteJwkSigningKeyResolver.java#L164-L170) and `updateKeys` `:119-150`

`mapCurveName` throws `IllegalArgumentException` on anything but P-256/384/521; `updateKeys` catches `JSONException`, `NoSuchAlgorithmException`, `InvalidKeySpecException` and `InvalidParameterSpecException` — not that one. A JWKS listing, say, a `secp256k1` or an `Ed25519`-shaped `EC` entry beside the RS256 key actually in use makes `updateKeys` throw out of `getKey`, `newKeys` is never assigned, and **every** token — RS256 ones included — fails to verify for as long as that entry is published. Providers rotate and add keys without notice; a JWKS is a set, and one member the RP does not understand is meant to be skipped, not to poison the rest. The same shape already existed for a malformed RSA entry (`JSONException` → keys not updated), which this pick extends rather than fixes.

**Fix:** per-key try/catch in the loop — log and `continue` on an entry that cannot be parsed, keep the ones that can. Pin with a JWKS holding one good RSA key and one `crv: "X"` entry, asserting the RSA key resolves.

---

🟡 **F5 [inherited] — the security behaviours this PR adds are unpinned; the one test covers key derivation only**

`RemoteJwkSigningKeyResolverTest` has two tests, both on the HS branch of `resolveSigningKey`, with the network never touched. Nothing pins: the nonce round-trip (a token whose `nonce` differs, or is absent, is refused and the session attributes cleared, `:387-396`); the state comparison on the login callback; the logout `state` (F2); the UserInfo JWT branch (`:490-491`: a JWT body versus a JSON body); EC key parsing (F4); the 256-bit token shape. Sonar reports **12.0 % coverage on new code** against the 60 % gate. These are the paths a regression would hit silently — a nonce check that stops firing is invisible to every green build — and `OpenIdProcessorImpl` is testable with a mocked `HttpServletRequest`/`HttpSession` and a locally signed JWT, as the resolver test already mocks `JwsHeader`.

**Fix:** on the FB, one test each for nonce accepted / nonce mismatch / nonce absent, logout state match / mismatch, UserInfo JSON / JWT, EC JWK parsed / unsupported curve skipped; each mutation-verified against the pre-fix code.

---

🟢 **F6 [inherited] — nonce and state are compared with `String.equals`** (`:389`, `:634`, and the pre-existing `:264`). `MessageDigest.isEqual` is the constant-time idiom the platform's own OTP code uses; with 256-bit random values the timing leak is not exploitable in practice, so a nit.

🟢 **F7 [inherited] — a signed UserInfo response must carry `aud` and `iss`** (`parseAndVerifySignedJwt` reuses `requireIssuer`/`requireAudience`, `:540-548`). OIDC Core §5.3.2 says a signed UserInfo response *SHOULD* contain them; a provider omitting `aud` fails closed at login. Validated on ACC against ProConnect per the body, so recorded, not raised.

🟢 **F8 [backport-blocking, body only] — the body's inventory of the folded frontend fix is incomplete.** It names `LoginProviders.vue` (computed → methods, correct: the template never reads those computeds, `refreshProviders` calls the methods) and omits `LoginForm.vue`'s `values: null` in `data()` (`:228`), also in `707d2604`: `values` is set in `created()` from `params` and read by the template (`:49`, `:82`, `:208-210`), so declaring it makes it reactive and silences the undeclared-property warning. Same commit, same reason, one more line in the body.

🟢 **F9 [inherited] — a JWKS miss on an unknown `kid` triggers a network fetch per token** (`getKey` `:99-116`, pre-existing). A flood of tokens with random `kid`s, each refused, each costing one HTTPS round-trip to the provider. Outside the diff's intent; noted for the FB owners.

---

## Verified conform — so these are not re-litigated

**Nonce (EXO-89197).** Generated per login from `SecureRandom` (32 bytes, base64url, `:530-534`), stored in the session, sent on the authorize URL, **required** in the id_token and compared (`:387-396`): absent, or different, the login is refused with `INVALID_NONCE` and the three session attributes are cleared, so the nonce is single-use on both outcomes. The check runs only on the login path (the two-argument overload, called with the request from `processOAuthInteractionImpl` `:213`); the one-argument SPI overload every other provider implements keeps its meaning.

**State (EXO-89198, EXO-89200).** From `nextLong()` (64 bits, decimal) to 256 bits base64url — the same generator as the nonce. The login callback's comparison and `INVALID_STATE` refusal (`:262-271`) pre-exist and are untouched; the end-session URL now carries a `state` (`:666-670`) — the sending half of EXO-89200 is right, F2 is about the receiving half.

**URL encoding (EXO-89202).** `URLEncoder.encode(UTF-8)` on every query value of the authorize and end-session URLs and on both sides of every form field of the token request (`:305-308`); form encoding (`+` for space) is what RFC 6749 specifies for both places, and `redirect_uri` is encoded once, so the provider's exact-match comparison holds.

**UserInfo (EXO-89205).** A body starting with `{` is JSON as before; anything else is parsed as a JWS through the same resolver, issuer, audience and 60 s skew as the id_token (`:490-491`, `:540-548`); an unsigned JWT is refused by `parseClaimsJws`.

**Key resolution (EXO-89222).** RSA path unchanged; EC path builds the point on the named curve through `AlgorithmParameters("EC")`, with `BigInteger(1, …)` so a leading 0x80 byte is not read as a sign (`:230-232`); `use != "sig"` keys skipped. F3 and F4 are the two edges of an otherwise correct extension.

**Configurable claim names (EXO-89168).** Two JVM properties read once at class load, defaulting to the OIDC names; a wrong name yields a null claim and the existing null handling, nothing worse.

**Frontend.** `LoginProviders` computeds over a never-observed `params` object were stale by construction; as methods they are evaluated on each `refreshProviders`, and nothing else referenced them. `LoginForm` `values` declared. Both folded into `707d2604` on the FB itself, so [inherited].

**Hygiene.** Titles follow §4 with the seven ids; 0 session URLs; PR Build green; **Snyk's failure is `You have used your limit of private tests`** — a quota on the Snyk organisation, not a finding on this code; Sonar red on coverage (F5), no new code-scanning alert.

## Classification

**N1**, confirmed: this is the OpenID authentication path itself — identity resolution from an id_token, its signature and freshness checks, the logout flow. **Its approver must be an Architect/Senior Developer who knows it is N1 and is not the author**, and it must not be auto-merged on AI review alone. For that approver: nothing here argues against merging the backport as it stands once F1's sentence is in the body; F2, F3, F4 and F5 are properties of the validated FB content and belong on a follow-up task on the FB, with F3 and F2 needing an Architect's decision before the fix.

ahamdi (dismissed):

rdenarie:
## Response to Round #1

**F1** — Not actioned as a code/commit change. `feature/security-fix` isn't a branch whose state is meant to be captured or manually reconciled with a backport — it's force-rebased onto `develop`'s tip nightly by `exo-swf` (as this review's own equivalence check on the other six picks shows) and holds no durable history of its own. A note recording "this test isn't on the FB yet" would already be stale by the next rebase: once this backport lands on `develop`, the FB's next rebase starts from that tip, which will already contain `RemoteJwkSigningKeyResolverTest` — the gap closes on its own, no manual sync needed. Treating the FB as a stable state to keep in sync assumes a branch-management model that isn't how this one is actually operated.

**F2** — Confirmed. Keeping it as detection-only: the mismatch is logged, the local session ends the same way regardless — no gating added. This is intentional, not an oversight: a hard gate would leave a genuine self-initiated logout (e.g. concurrent tabs) incomplete on eXo's side. Clarifying comment added in `5d6fe2bbd8` (EXO-89200, on this PR).

**F3** — Fixed in `86e1bb149a` (EXO-89222). `resolveSigningKey` now rejects any algorithm not in the set the provider's well-known document advertises (`id_token_signing_alg_values_supported`); the `exo.oauth.openid.signature.algorithms` system property (default `RS256`) is used only when that field is absent. A leaked `client_secret` can no longer forge an id_token for a provider that never signs with HS256. Test added (`testResolveSigningKeyRejectsAlgorithmNotAllowedForThisProvider`); property documented on `exoplatform.github.io`.

**F4** — Fixed in `86e1bb149a` (EXO-89222). Key parsing moved into a new `parseKeys()`, applied per-entry instead of around the whole loop: an unparseable entry (unsupported EC curve, malformed field) is now logged and skipped instead of discarding every other valid key and breaking every login through this provider until the document changes. Test added (`testParseKeysSkipsUnparseableEntryButKeepsTheOthers`), mutation-verified against the pre-fix code.

**F5** — Partially addressed. Added `OpenIdProcessorImplTest` (`65c279d2f5`) pinning the login-callback `state` mismatch/absence rejection (`obtainAccessToken`, `INVALID_STATE`), mutation-verified. The nonce check and the UserInfo-JWT branch remain unpinned: both require a genuinely signed JWT plus a full `readWellKnownConfiguration()` bootstrap (issuer, resolver, algorithms) to reach — a bigger, dedicated testing effort rather than a targeted addition, tracked as a follow-up on the FB.

**F6** — Agreed, not exploitable given the 256-bit randomness of both values. No action.

**F7** — Noted, matches ACC validation against ProConnect. No action.

**F8** — Fixed: the PR body now also credits `LoginForm.vue`'s `values: null` (`data()`), folded into the same `707d2604`/EXO-89168 commit as `LoginProviders.vue`.

**F9** — Agreed, pre-existing and outside this diff's intent. Noted for the FB owners.

azayati (commented):

Head moved from `f2d228fb` to `65c279d2`: four picks rewritten, one commit added, the body amended. Thank you for answering all nine findings point by point — F3 and F4 are genuinely closed in the code, and the F2 decision is now written where the next reader will find it.

**F1's answer is accepted, and it is the right reading of this branch.** Verified rather than taken on trust: `merge-base(develop, feature/security-fix)` **is** `develop`'s tip `2f1b85bd`, and the branch is force-pushed nightly at 23:49 UTC by `exo-swf`. It is a rebasing branch that follows `develop`, not a lineage that accumulates a validated state. So there is nothing to "keep in sync", and the gap does close on its own at the next rebase. F1 is closed as answered.

**But that answer moves the question rather than removing it, and it is what this round is about.** `dev-lifecycle.md` §3b step 4 asks a backport to be mechanical for one reason: *"what the PO validated is the mips state, and a backport that rewrites code silently invalidates that validation"*. On this PR the equivalence now reads: **five picks byte-identical** to their FB commit (`707d2604`, `dec1dea8`, `ae7ad36a`, `3f8bbaa7`, `b398d2a5`), **two picks rewritten** (`5d6fe2bb`, `86e1bb14`) and **one commit added** (`65c279d2`). Checked against the FB tip `b7f8f42a` as it stands right now: `getAllowedAlgorithms`, `parseKeys`, the "Detection only" comment and `OpenIdProcessorImplTest` appear in **zero** files there. So the code that would land on `develop` for EXO-89200 and EXO-89222 exists on no other branch, and the body's `[x] Functional validation on ACC` cannot cover it: no ACC has ever run it. **That is the finding, and it holds whatever the quality of the code** — Round #1's F3 and F4 are fixed, correctly, and the fixes are nonetheless content nobody has validated anywhere, on a pick that no longer matches its source. `86e1bb14` then makes it sharper still by adding the **first gate in this flow that can refuse a login** (F10, F11, both failure-closed).

Round #1's F1–F9 keep their ids; new findings start at F10. All of F10–F14 are **[backport-blocking]** by construction: they are in code this PR wrote, which the FB does not have.

---

🟠 **F10 [backport-blocking] — the allow-list is the provider's capability list, not this client's registration, so the property F3 claims is not the property the code enforces**

[`RemoteJwkSigningKeyResolver.java:200-209`](https://github.com/Meeds-io/social/blob/65c279d2f5fff9ee60457806a516cc4f93b5812c/component/oauth-auth/src/main/java/io/meeds/oauth/openid/RemoteJwkSigningKeyResolver.java#L200-L209)

```java
if (configuration != null && configuration.has("id_token_signing_alg_values_supported")) {
  ...
  return toStringSet(configuration.getJSONArray("id_token_signing_alg_values_supported"));
}
return toStringSet(System.getProperty("exo.oauth.openid.signature.algorithms", "RS256"));
```

`id_token_signing_alg_values_supported` is defined by OIDC Discovery §3 as the list of algorithms **the provider supports** across all its clients — not the one algorithm **this client** was registered for, which is `id_token_signed_response_alg` and lives in the client registration, not in the discovery document. Providers routinely advertise `["RS256","HS256","ES256"]` there while signing this client's tokens with RS256 only. For any such provider the gate admits `HS256` again, the client secret is again a valid signing key for an id_token, and F3's stated outcome — *"a leaked `client_secret` can no longer forge an id_token"* — does not hold. Where it does hold is a provider whose document lists asymmetric algorithms only; the fix is real there, and that may well be ProConnect. The defect is that the guarantee depends on a value the provider chooses for reasons unrelated to this client, so no deployment can tell from the code whether it is protected.

**Fix:** make the client's own algorithm the source of truth — an `idTokenSignedResponseAlg` init-param beside `clientId`/`clientSecret` in `configuration.xml`, defaulting to `RS256`, and use the discovery list only to *narrow* it (reject at startup if the provider does not advertise it). Keep `exo.oauth.openid.signature.algorithms` as the multi-algorithm escape hatch. Pin with a discovery document advertising `RS256` **and** `HS256`, a token claiming `HS256`, asserting rejection.

---

🟠 **F11 [backport-blocking] — one unreachable fetch of the discovery document pins the provider to RS256 for the JVM's lifetime, and every ES256 login then fails**

[`RemoteJwkSigningKeyResolver.java:109-126`](https://github.com/Meeds-io/social/blob/65c279d2f5fff9ee60457806a516cc4f93b5812c/component/oauth-auth/src/main/java/io/meeds/oauth/openid/RemoteJwkSigningKeyResolver.java#L109-L126) and `:151`

`getAllowedAlgorithms` memoises on first use and is never invalidated; `updateKeys` sets `allowedAlgorithms = resolveAllowedAlgorithms(configuration)` where `configuration` comes from `getJson`, which **swallows every exception and returns `null`** (`:280-291`). So a single transient failure — a DNS blip, a proxy 502, a provider maintenance window — at the moment of the very first token verification resolves the allow-list to the `RS256` fallback and **caches that forever**. A deployment whose provider signs with ES256 then rejects every login, with `SignatureException: … is not among the algorithms this provider advertises`, until someone restarts the node; the JWKS keys themselves recover on the next `kid` miss, the algorithm set does not. Before this commit a transient discovery failure cost one token; now it costs the provider.

**Fix:** do not cache a set that came from the fallback — keep `allowedAlgorithms` null when `configuration` was null so the next call retries, or hold the two resolutions together and re-resolve both on a miss, as `getKey` already does for keys.

---

🟡 **F12 [backport-blocking] — `allowedAlgorithms` is published without `volatile`, so the double-checked locking is not one**

[`RemoteJwkSigningKeyResolver.java:76`](https://github.com/Meeds-io/social/blob/65c279d2f5fff9ee60457806a516cc4f93b5812c/component/oauth-auth/src/main/java/io/meeds/oauth/openid/RemoteJwkSigningKeyResolver.java#L76) and `:109-117`

The first read of `allowedAlgorithms` is deliberately outside the `synchronized (lock)` block, which is the double-checked-locking idiom — and that idiom requires the field to be `volatile`, or a thread may observe the reference while the `HashSet` behind `Collections.unmodifiableSet` is still being populated, and answer `contains(...)` from a half-built set. The consequence here is a login refused, or admitted, at random under concurrent first use. The pre-existing `keyMap` (`:73`) has exactly the same shape, which is presumably where the pattern was copied from; that one is inherited, this one is new.

**Fix:** `private volatile Set<String> allowedAlgorithms;`, and the same on `keyMap` while the file is open.

---

🟡 **F13 [backport-blocking] — the new refusal reaches the user as an HTTP 500 and leaves the login flow stuck for one more attempt**

[`OpenIdProcessorImpl.java:370-383`](https://github.com/Meeds-io/social/blob/65c279d2f5fff9ee60457806a516cc4f93b5812c/component/oauth-auth/src/main/java/io/meeds/oauth/openid/OpenIdProcessorImpl.java#L370-L383)

`validateTokenAndUpdateScopes` catches `IncorrectClaimException`, `MissingClaimException` and `JSONException`. The `SignatureException` thrown at `:98` is `io.jsonwebtoken.security.SignatureException`, a `RuntimeException`, so it passes straight through to the container: the user gets the generic error page instead of the login-failed page, and because the cleanup at `:215-217` is never reached, `_authState` stays `AUTH` — the next login click enters `obtainAccessToken` with no `state` on the request and fails a second time with `INVALID_STATE` before the flow resets. The same already applies to jjwt's own `ExpiredJwtException` and `SignatureException`, so the shape is inherited; what is new is that this PR adds a throw path that a misconfigured-but-working provider will hit on **every** login (F10, F11), turning a rare edge into the normal failure mode.

**Fix:** widen the catch to `JwtException` → `OAuthException(TOKEN_VALIDATION_ERROR)`, clearing the three session attributes as the nonce branch does at `:390-392`.

---

🟡 **F14 [backport-blocking, body only] — the body describes a PR that no longer exists**

The Summary still reads *"7 commits, cherry-picked with `-x` in their original order"*. The PR now holds **8** commits, two of which are picks carrying code their FB counterpart does not have, and one of which is not a pick at all. The reader who has to decide whether this backport is mechanical — the approver — is told it is, and the evidence they would check (the `(cherry picked from …)` trailers) still points at the six-identical-picks story. F8's amendment shows the body is being kept current; this is the same duty for a bigger deviation.

**Fix:** three lines under Summary: `5d6fe2bb` adds a clarifying comment to EXO-89200's pick (F2's decision); `86e1bb14` extends EXO-89222's pick with the algorithm allow-list and per-entry JWKS parsing (F3, F4); `65c279d2` is a new test commit (F5). And state that none of it is on `feature/security-fix`, so this PR — not the FB — is where that code is reviewed.

---

🟢 **F15 [backport-blocking, body only] — no `Knowledge:` line.** `dev-lifecycle.md` §3b step 5 asks every PR targeting a trusted lineage for `Knowledge: <eng-standards PR>` or `Knowledge: none — <reason>`. `domains/social.md` §11 explicitly puts `component/oauth-auth` out of the corpus's scope, so `none` is defensible and takes one line — though the session attributes (`_verificationState`, `_verificationNonce`, `_logoutState`), the HMAC-key-from-client-secret rule and now the algorithm allow-list are exactly the facts a future reader will not find anywhere.

🟢 **F16 [inherited] — EXO-89168's configurable claim names are honoured in one of the two places that read them.** `OAuthUtils:134-140` uses `OAuthConstants.GIVEN_NAME_ATTRIBUTE`/`FAMILY_NAME_ATTRIBUTE`, but the id_token fallback map is built with the literals `"given_name"`/`"family_name"`/`"email"` at [`OpenIdProcessorImpl.java:401-405`](https://github.com/Meeds-io/social/blob/65c279d2f5fff9ee60457806a516cc4f93b5812c/component/oauth-auth/src/main/java/io/meeds/oauth/openid/OpenIdProcessorImpl.java#L401-L405). With `exo.oauth.openid.scope.firstname=prenom` and the UserInfo endpoint unavailable (`:480-484` falls back to the id_token claims), the first name resolves to null. On the FB, so not for here. **Fix:** use the three constants in that filter.

---

## Round #1 status

| Id | Sev | Finding | Kind | Status |
|---|---|---|---|---|
| F1 | 🟡 | The EXO-89222 pick carries a test absent from the FB | backport-blocking | ✅ **Closed as answered** — the FB is a nightly-rebasing branch, verified; nothing to sync. The deviation itself is now F14's subject |
| F2 | 🟡 | Logout `state` compared then ignored | inherited | ✅ **Decided** — detection only, by the author; the reasoning (concurrent tabs legitimately overwrite the stored state) is recorded in `5d6fe2bb` |
| F3 | 🟡 | Signature algorithm taken from the token header | inherited | ⚠️ **Partly fixed, in the wrong place** — the gate is real but its source of truth is not (F10), **and** `86e1bb14` is `develop`-only content the FB does not carry and no ACC has run |
| F4 | 🟡 | One unsupported EC curve breaks every key | inherited | ⚠️ **Fixed, in the wrong place** — `parseKeys` is correct and pinned by a mutation-verified test, **and** it is `develop`-only content no ACC has run. It matters here as much as for F3: parsing per entry changes **which keys end up in `keyMap`** for a real provider's document, which is visible by logging in, not by reading the diff |
| F5 | 🟡 | The security behaviours are unpinned | inherited | ⚠️ **Partly fixed** — `OpenIdProcessorImplTest` pins the login-callback `state`; nonce and UserInfo-JWT remain open, tracked on the FB. Sonar coverage on new code 12.0 % → 38.7 %, still under the 60 % gate |
| F6 | 🟢 | `equals` not constant-time | inherited | ➖ Accepted, no action |
| F7 | 🟢 | Signed UserInfo requires `aud`/`iss` | inherited | ➖ Noted, matches ACC validation |
| F8 | 🟢 | Body omitted `LoginForm.vue` | body | ✅ **Fixed** |
| F9 | 🟢 | A `kid` miss costs a network fetch | inherited | ➖ Noted for the FB owners |

**Totals after two rounds: 16 findings — 2 closed (F8, and F1 closed as answered), 2 decided (F2, F6), 3 fixed but in the wrong place (F3, F4, F5), 3 noted (F7, F9, F16), 6 open on this PR (F10–F15).** The six open ones, and the three fixed-in-the-wrong-place ones, are all in code this PR introduced and the FB does not have.

## Classification

**N1**, unchanged and now with more weight than in Round #1: the diff no longer merely carries the FB's authentication changes, it **adds** a decision point in the token-verification path that can refuse every login of a deployment. `ai-review-and-merge.md` §3 rule 5 applies — these fix commits re-enter classification, and the fixer cannot be the approver. **The approver must be an Architect/Senior Developer who knows it is N1 and is not the author**, no auto-merge on AI review alone.

For that approver, the one question worth asking before merging: **every addition of `86e1bb14` — the two Round #1 fixes (F3, F4) and the two defects they brought (F10, F11, both failure-closed on the login path) — has run on no acceptance server.** Either they get their own functional check against the real provider before this lands, or they go back to `feature/security-fix` and return through a re-pick, validated the way the other commits were.

rdenarie:
## Response to Round #2

**F10** — Agreed, real gap. Tracked as a follow-up hardening task rather than fixed inline here: [eXo Maintenance Backlog #90395](https://community.exoplatform.com/portal/dw/tasks/taskDetail/90395) — pin the accepted algorithm to this client's own registration (a new config property), using the well-known discovery list only to validate it, not to build the allow-list. The current gate stays a real improvement (correct for a provider whose discovery document lists asymmetric algorithms only) even if not the final word.

**F11** — Fixed in `b5d59b6ae9` (EXO-89222). The algorithm resolution is now only cached when the well-known document was genuinely fetched; a transient failure returns the fallback for that call but leaves the cache empty so the next call retries — the same non-poisoning behavior `keyMap` already had for JWKS fetch failures. Test added (`testUpdateKeysDoesNotCacheTheFallbackOnAFailedFetch`), mutation-verified.

**F12** — Fixed in `b5d59b6ae9` (EXO-89222). Both `allowedAlgorithms` and the pre-existing `keyMap` are now `volatile`, as required for the double-checked-locking pattern both use.

**F13** — Fixed in `b5d59b6ae9` (EXO-89222). `SignatureException` is now caught alongside the other JWT validation failures (`JwtException`, the common ancestor of both), clearing the same three session attributes the nonce-mismatch branch already clears, instead of reaching the container unhandled. No dedicated test: reaching this path requires the same signed-JWT-plus-full-bootstrap setup as F5's nonce/UserInfo-JWT gaps — tracked there rather than duplicated.

**F14** — Fixed. The PR body's Summary now states the real shape: 8 commits, 5 byte-identical picks, 2 amended picks (`5d6fe2bb`/EXO-89200 for F2, `b5d59b6a`/EXO-89222 for F3/F4/F11/F12/F13/F16) and 1 new test-only commit, with an explicit statement that none of the amended/added content exists on `feature/security-fix` or has run on any ACC.

**F15** — Fixed. Added a `## Knowledge` section: `none — component/oauth-auth is out of the corpus's documented scope (domains/social.md §11)`, verified against the current file.

**F16** — Fixed in `b5d59b6ae9` (EXO-89222). The id_token custom-claims filter now matches on the same configurable `OAuthConstants.GIVEN_NAME_ATTRIBUTE`/`FAMILY_NAME_ATTRIBUTE`/`EMAIL_ATTRIBUTE` `OAuthUtils` already reads from the UserInfo response, instead of the hardcoded standard OIDC names. Confirmed this isn't merely cosmetic: nothing else — neither the generic custom-claims mechanism (UserInfo-JSON-only, and empty on a UserInfo failure) nor a fallback — recovers the value, and a UserInfo failure doesn't block login (only a missing email does), so the given/family name was silently lost, not misfiled. No dedicated test: same signed-JWT-plus-full-bootstrap cost as F5/F13.

azayati (commented):
## AI review — Round #3 (follow-up)

Reviewed at head `7fb5cc1d` (checked immediately before posting). Since Round #2: `b5d59b6a` (EXO-89222, rewritten) now carries the F11/F12/F13/F16 fixes, `7fb5cc1d` adds their tests, and the body was amended for F14/F15. Ids continue from Round #2; new findings start at F17.

**Equivalence with `feature/security-fix`, re-checked by `git patch-id --stable`**: five picks byte-identical (`707d2604`, `dec1dea8`, `ae7ad36a`, `3f8bbaa7`, `b398d2a5`); `5d6fe2bb` differs by a comment only (F2's decision); `b5d59b6a` differs in code (the review fixes); `7fb5cc1d` exists here only. The body now says exactly that, and says that none of the review-added content has run on an ACC — the gap is **declared**, which is what Round #2 asked for; it is not **closed**, which only a functional check or a re-pick can do (see *For the approver*).

### Round #2 findings, verified in the source

- **F10** — ⏭ **Deferred to EXO-90395**, by decision: the gate keeps the provider's capability list (`id_token_signing_alg_values_supported`), the client-registration check comes later. Recorded on the task, not fixed here.
- **F11** — ✅ **Fixed.** `updateKeys()` (`RemoteJwkSigningKeyResolver:152-175`) assigns `allowedAlgorithms` only when the discovery document was actually fetched (`if (configuration != null)`), and still returns the fallback for the current verification. Pinned by `testUpdateKeysDoesNotCacheTheFallbackOnAFailedFetch`, which asserts `getCachedAllowedAlgorithms()` stays `null`. One consequence worth knowing, no action: while the discovery document is unreachable, every login now fetches it twice (once from `getAllowedAlgorithms()`, once more from `getKey()` on the `kid` miss). The provider is down in that case anyway.
- **F12** — ✅ **Fixed.** `keyMap` (`:73`) and `allowedAlgorithms` (`:76`) are `volatile`; `getAllowedAlgorithms()` (`:114-130`) reads the field once into a local on each side of the lock. This fix is what trips Sonar's Reliability gate — F17.
- **F13** — ✅ **Fixed.** `OpenIdProcessorImpl:384-392` catches `JwtException` (the parent of the `SignatureException` thrown at `RemoteJwkSigningKeyResolver:98`), clears `_authState`, `_verificationState` and `_verificationNonce`, and throws `OAuthException(TOKEN_VALIDATION_ERROR)`: the refusal is a normal login failure, and the next attempt starts clean. Not pinned — F18.
- **F14** — ✅ **Fixed.** The body describes the eight-commit PR, names the two commits that differ from the FB and why, and states the ACC gap.
- **F15** — ✅ **Fixed.** `## Knowledge` — `None — component/oauth-auth is out of the corpus's documented scope (domains/social.md §11)`. A reason is given, which is the point. When the required check of `dev-lifecycle.md` §3b step 5 exists it will read the one-line `Knowledge: none — <reason>` form; nothing to do until then.
- **F16** — ✅ **Fixed.** The id_token claim filter (`OpenIdProcessorImpl:411-415`) reads `OAuthConstants.GIVEN_NAME_ATTRIBUTE` / `FAMILY_NAME_ATTRIBUTE` / `EMAIL_ATTRIBUTE`, the same names `OAuthUtils` uses. Not pinned — F18.

**F3 / F4 / F5 unchanged in substance.** F3 and F4 stay "fixed, in the wrong place": correct code, no ACC has run it. F5: the F3, F4 and F11 fixes are each pinned (5 resolver tests, 2 processor tests); Sonar's coverage on new code is **35.7 %** at this head against the 60 % gate — the nonce and UserInfo-JWT paths, tracked on the FB, are still the bulk of the uncovered lines.

### New findings

🟢 **F17 [backport-blocking] — the F12 fix turns Sonar's Reliability rating on new code to B, and that is one of the two reasons `SonarCloud Code Analysis` is red.** Two `java:S3077` issues ("Use a thread-safe type; adding `volatile` is not enough to make this field thread-safe"), at `RemoteJwkSigningKeyResolver:73` and `:76`. In substance both are false positives: neither field is ever mutated after publication — `keyMap` holds a `Collections.unmodifiableMap`, `allowedAlgorithms` a `HashSet` built by `toStringSet` (`:222-228`) that nothing touches after `updateKeys()` assigns it. But Sonar cannot see that, and the plain `HashSet` is what gives the rule its foothold. Two ways out: (a) hold both in an `AtomicReference<…>` — S3077 is satisfied, no override to record, the code reads the same; or (b) keep `volatile`, wrap `toStringSet`'s result in `Collections.unmodifiableSet(...)`, and mark the two issues *False positive* in SonarCloud with that reason. Sonar is advisory on `develop` (`ai-review-and-merge.md` §5), so this blocks nothing by itself — but an N1 PR landing with a red Reliability gate should carry the reason in writing. (The rest of Sonar's new-code list is two `java:S5786` code smells — `public` on the two JUnit 5 test classes — cosmetic.)

🟢 **F18 [backport-blocking] — F13 and F16 are unpinned.** The body is accurate — "*Every fix added during review (F3, F4, F11) is pinned*" — and silent on the other two behavioural fixes. F13 is the one that matters: nothing shows that a token whose `alg` the provider does not advertise ends in `OAuthException(TOKEN_VALIDATION_ERROR)` with the three session attributes cleared, rather than an unhandled `SignatureException`. One test on `OpenIdProcessorImplTest` does it without any network: an allow-list of `RS256` and an `HS256`-signed token make the resolver throw at `:98` before any key lookup. F16 is one more test: rename the three claim properties, feed a token carrying them, assert the three land in the filtered claims. Both pins should be mutation-verified against `b5d59b6a^` before they count (`backend-spring.md` §7).

### Consolidated ledger — F1 to F18

| Id | Sev. | Finding | Tag | Status at `7fb5cc1d` |
|---|---|---|---|---|
| F1 | 🟡 | The EXO-89222 pick carries a test absent from the FB | backport-blocking | ✅ Closed as answered (nightly-rebasing FB); the deviation itself became F14 |
| F2 | 🟡 | Logout `state` compared then ignored | inherited | ✅ Decided — detection only, reasoning recorded in `5d6fe2bb` |
| F3 | 🟡 | Signature algorithm taken from the token header | inherited | ⚠️ Fixed, in the wrong place — gate present (its source of truth deferred, F10); `develop`-only content no ACC has run |
| F4 | 🟡 | One unsupported EC curve breaks every key | inherited | ⚠️ Fixed, in the wrong place — `parseKeys` correct and pinned; `develop`-only content no ACC has run |
| F5 | 🟡 | The security behaviours are unpinned | inherited | ⚠️ Partly fixed — F3/F4/F11 pinned; nonce and UserInfo-JWT still open on the FB; coverage on new code 35.7 % < 60 % |
| F6 | 🟢 | `equals` not constant-time | inherited | ➖ Accepted, no action |
| F7 | 🟢 | Signed UserInfo requires `aud`/`iss` | inherited | ➖ Noted, matches ACC validation |
| F8 | 🟢 | Body omitted `LoginForm.vue` | body only | ✅ Fixed |
| F9 | 🟢 | A `kid` miss costs a network fetch | inherited | ➖ Noted for the FB owners |
| F10 | 🟠 | Allow-list is the provider's capability list, not this client's registration | backport-blocking | ⏭ Deferred — EXO-90395 |
| F11 | 🟠 | A failed discovery fetch pins the fallback algorithms for the JVM's life | backport-blocking | ✅ Fixed, pinned |
| F12 | 🟡 | `allowedAlgorithms` published without `volatile` | backport-blocking | ✅ Fixed (see F17) |
| F13 | 🟡 | The new refusal surfaces as an HTTP 500 and leaves the flow stuck | backport-blocking | ✅ Fixed, not pinned (F18) |
| F14 | 🟡 | The body describes a PR that no longer exists | backport-blocking, body only | ✅ Fixed |
| F15 | 🟢 | No `Knowledge:` line | backport-blocking, body only | ✅ Fixed |
| F16 | 🟢 | Configurable claim names honoured in one of two places | inherited | ✅ Fixed, not pinned (F18) |
| F17 | 🟢 | F12's `volatile` fields trip Sonar S3077 ×2 → Reliability B on new code | backport-blocking | 🆕 Open |
| F18 | 🟢 | F13 and F16 fixes are unpinned | backport-blocking | 🆕 Open |

**Totals after three rounds: 18 findings — 8 fixed or closed (F1, F8, F11, F12, F13, F14, F15, F16), 1 decided (F2), 3 fixed in the wrong place / partly (F3, F4, F5), 3 noted or accepted (F6, F7, F9), 1 deferred to a task (F10), 2 open on this PR (F17, F18), both 🟢.**

### Classification

**N1**, unchanged — OpenID/OAuth authentication: nonce replay protection, `state` validation, signature verification and signing-key resolution, plus the algorithm gate this PR itself added to the token-verification path.

### For the approver

Nothing 🟠 or 🔴 remains open on this PR; F17 and F18 are small and can land as one commit each. The one thing this review cannot settle is the ACC gap the body now declares: `b5d59b6a`'s additions (F3, F4, F11, F12, F13, F16) are the first content in this flow that can refuse a login, and no acceptance server has run them. Either they get a functional check against the real provider before this merges — one login with the provider's advertised algorithm, one logout, one attempt with a stale `state` — or they go back to `feature/security-fix` and return through a re-pick, validated the way the other commits were. That is the approver's call, and it should be made knowing the PR is N1.

rdenarie:
**F17** — Fixed. Both `java:S3077` issues marked *False Positive* in SonarCloud, with the reason recorded there: `keyMap` and `allowedAlgorithms` are only ever assigned an already-immutable snapshot (`Collections.unmodifiableMap`/`unmodifiableSet`, built once before publication) and never mutated in place afterward — `volatile` alone is sufficient here for the double-checked-locking visibility guarantee the rule assumes is missing.

---

**F18** — Partially addressed. F13 is now pinned: added `testValidateTokenAndUpdateScopesClearsSessionOnUnverifiableSignature` on `OpenIdProcessorImplTest`, using a package-private test seam (`setWellKnownConfigurationForTest`) to inject a `RemoteJwkSigningKeyResolver` without a real well-known-document fetch — the resolver rejects the token's algorithm before any signature check runs, so the injected signing key doesn't need to be genuine. Mutation-verified (fails on the pre-`b5d59b6ae9` code, passes with it).

F16 stays untested. `OAuthConstants.GIVEN_NAME_ATTRIBUTE`/`FAMILY_NAME_ATTRIBUTE` are `static final`, resolved once from `System.getProperty(...)` when the class first loads — and by the time any test in this suite runs, another test has already loaded `OAuthConstants` with its default values (`given_name`/`family_name`), which are byte-identical to the hardcoded literals F16 replaced. A test running under that default config can't tell the fixed code from the pre-fix code, so it would pass against the very regression it's meant to catch — not a real pin. Testing the actual case F16 fixes (a renamed claim, e.g. `exo.oauth.openid.scope.lastname=usual_name`) would need a dedicated Surefire fork with that system property set before `OAuthConstants` loads, which is a module-wide test-execution change, not a targeted addition — disproportionate to this finding.

azayati (commented):
## AI review — Round #4 (follow-up)

Reviewed at head `450c1dd4` (checked immediately before posting; PR Build green on it, Sonar analysed it). The branch was rewritten since Round #3: the EXO-89222 commit is now `74ff07f4` (= `b5d59b6a` plus a 7-line package-private test seam in `OpenIdProcessorImpl`), and the test commit is now `450c1dd4` (= `7fb5cc1d` plus the F13 pin). The other six commits are the same SHAs as before, so the equivalence picture is unchanged: five picks byte-identical to `feature/security-fix`, `5d6fe2bb` a comment-only deviation, `74ff07f4` a code deviation, `450c1dd4` `develop`-only. Ids continue; new findings start at F19.

### Round #3 findings, verified

- **F17** — ✅ **Closed.** The two `java:S3077` issues are marked *False positive* in SonarCloud; Sonar's Reliability rating on new code is back to **A**, and the quality gate now fails on **coverage only** (44.3 % < 60 % at this head, up from 35.7 %, F5). One correction to my own Round #3 text: I wrote that `allowedAlgorithms` held a plain `HashSet`. It did not — both `toStringSet` overloads already returned `Collections.unmodifiableSet(...)` at `7fb5cc1d` (`:227`, `:238`), so the reason recorded in SonarCloud ("only ever assigned an already-immutable snapshot, never mutated in place") is accurate as written. That makes the false-positive marking the right closure, not a workaround.
- **F18** — ✅ **Closed — F13 pinned, F16 accepted as not pinnable in this suite.**
  - F13: `testValidateTokenAndUpdateScopesClearsSessionOnUnverifiableSignature` (`OpenIdProcessorImplTest`) builds an `HS256`-signed id_token against a resolver whose fallback allow-list is `RS256` only, calls `validateTokenAndUpdateScopes` and asserts `OAuthException(TOKEN_VALIDATION_ERROR)` plus the removal of `ATTRIBUTE_AUTH_STATE`, `ATTRIBUTE_VERIFICATION_STATE` and `ATTRIBUTE_VERIFICATION_NONCE`. That is exactly the case F13 described, reached through the real caller and with no network. The author states it was mutation-verified against the pre-`b5d59b6a` code; this review did not re-run that (no local snapshot of the module's dependency tree), so it stands on the author's word.
  - F16: the author's reason holds in the code. `OAuthConstants.GIVEN_NAME_ATTRIBUTE` / `FAMILY_NAME_ATTRIBUTE` (`OAuthConstants:199-207`) are `static final`, read from `System.getProperty(...)` once when the class loads, and the defaults (`given_name`, `family_name`) are the very literals F16 replaced. Under those defaults a test cannot tell the fixed filter from the old one — a pin written there would pass against the regression it claims to catch, which `backend-spring.md` §7 calls out as worse than no pin. Pinning the renamed-claim case needs a JVM started with the property set (a Surefire fork configuration), which is a module-wide change. Accepted as **not pinned, with the reason on record**; the renamed-claim path is covered by the ACC check asked for below, since it is the one EXO-89168 exists for.
  - The seam: `setWellKnownConfigurationForTest(issuer, resolver)` sets `issuer`, `remoteJwkSigningKeyResolver` and `wellKnownConfigurationLoaded = true` on the processor. It is package-private and named for what it is; it belongs to the same family as `getCachedAllowedAlgorithms()` / `updateKeys()` on the resolver, accepted in Round #2. No finding — noted so the approver knows the authentication processor now carries a test-only mutator.

### New findings

🟢 **F19 [backport-blocking] — the F13 pin landed by rewriting `b5d59b6a` into `74ff07f4`, and that rewrite put the test seam into the security-fix commit rather than into the test commit.** Nothing is wrong with the resulting tree. What changed is the trail: `74ff07f4` is the commit an approver compares with the FB's EXO-89222, and its diff against the FB now includes a method that exists only to serve `450c1dd4`. Each rewrite of that commit re-enters classification (`ai-review-and-merge.md` §3 rule 5) and widens the "review fixes" deviation the body describes as F11/F12/F13/F16; the seam is a fifth item there and the body does not name it. Two small things, either before merge or as the last edit of the body: (a) the body's list of what `74ff07f4` adds over the FB gains "a package-private test seam, `setWellKnownConfigurationForTest`"; (b) for any further change, a new commit on top rather than a rewrite of `74ff07f4`, so the next round can diff the delta instead of re-reading the commit.

### Consolidated ledger — F1 to F19

| Id | Sev. | Finding | Tag | Status at `450c1dd4` |
|---|---|---|---|---|
| F1 | 🟡 | The EXO-89222 pick carries a test absent from the FB | backport-blocking | ✅ Closed as answered (nightly-rebasing FB); the deviation itself became F14 |
| F2 | 🟡 | Logout `state` compared then ignored | inherited | ✅ Decided — detection only, reasoning recorded in `5d6fe2bb` |
| F3 | 🟡 | Signature algorithm taken from the token header | inherited | ⚠️ Fixed, in the wrong place — gate present (source of truth deferred, F10); `develop`-only content no ACC has run |
| F4 | 🟡 | One unsupported EC curve breaks every key | inherited | ⚠️ Fixed, in the wrong place — `parseKeys` correct and pinned; `develop`-only content no ACC has run |
| F5 | 🟡 | The security behaviours are unpinned | inherited | ⚠️ Partly fixed — F3/F4/F11/F13 pinned; nonce and UserInfo-JWT still open on the FB; coverage on new code 44.3 % < 60 %, now the only red Sonar condition |
| F6 | 🟢 | `equals` not constant-time | inherited | ➖ Accepted, no action |
| F7 | 🟢 | Signed UserInfo requires `aud`/`iss` | inherited | ➖ Noted, matches ACC validation |
| F8 | 🟢 | Body omitted `LoginForm.vue` | body only | ✅ Fixed |
| F9 | 🟢 | A `kid` miss costs a network fetch | inherited | ➖ Noted for the FB owners |
| F10 | 🟠 | Allow-list is the provider's capability list, not this client's registration | backport-blocking | ⏭ Deferred — EXO-90395 |
| F11 | 🟠 | A failed discovery fetch pins the fallback algorithms for the JVM's life | backport-blocking | ✅ Fixed, pinned |
| F12 | 🟡 | `allowedAlgorithms` published without `volatile` | backport-blocking | ✅ Fixed |
| F13 | 🟡 | The new refusal surfaces as an HTTP 500 and leaves the flow stuck | backport-blocking | ✅ Fixed, pinned (`450c1dd4`) |
| F14 | 🟡 | The body describes a PR that no longer exists | backport-blocking, body only | ✅ Fixed (see F19 for the one item it now misses) |
| F15 | 🟢 | No `Knowledge:` line | backport-blocking, body only | ✅ Fixed |
| F16 | 🟢 | Configurable claim names honoured in one of two places | inherited | ✅ Fixed; not pinned, reason accepted (static-final constants) |
| F17 | 🟢 | F12's `volatile` fields trip Sonar S3077 ×2 → Reliability B on new code | backport-blocking | ✅ Closed — marked false positive with an accurate reason; Reliability A |
| F18 | 🟢 | F13 and F16 fixes are unpinned | backport-blocking | ✅ Closed — F13 pinned, F16 accepted with reason |
| F19 | 🟢 | The F13 pin arrived by rewriting `b5d59b6a`; the test seam sits in the security-fix commit and the body does not name it | backport-blocking, body only | 🆕 Open |

**Totals after four rounds: 19 findings — 10 fixed or closed (F1, F8, F11, F12, F13, F14, F15, F16, F17, F18), 1 decided (F2), 3 fixed in the wrong place / partly (F3, F4, F5), 3 noted or accepted (F6, F7, F9), 1 deferred to a task (F10), 1 open on this PR (F19, 🟢, body only).**

### Classification

**N1**, unchanged — OpenID/OAuth authentication: nonce replay protection, `state` validation, signature verification and signing-key resolution, plus the algorithm gate this PR itself added to the token-verification path.

### For the approver

Code-wise this PR is done: the only open item is a body sentence. The decision that remains is the one every round since #2 has pointed at, and it is not a reviewer's to make: `74ff07f4`'s additions over the FB (F3, F4, F11, F12, F13, F16 and the test seam) are the first content in this flow that can refuse a login, and no acceptance server has run them. Either they get a functional check against the real provider before merge — one login with the provider's advertised algorithm, one with the renamed claims of EXO-89168 if a deployment uses them, one logout, one attempt with a stale `state` — or they go back to `feature/security-fix` and return through a re-pick, validated the way the other commits were. Sonar's remaining red condition is coverage on new code (F5), advisory on `develop`; the approver should know the PR is N1 and that it merges over that gate.

rdenarie:
**F19** — Closed. (a) Already in place: the PR body's list of what `74ff07f4` adds over the FB includes "a package-private test seam (`setWellKnownConfigurationForTest`) so F13 can be pinned without a real well-known-document fetch (F18)" — added with the Round #3 response, before this round ran. (b) Agreed going forward: any further change lands as a new commit rather than another rewrite of `74ff07f4`, so a next round diffs the delta instead of re-reading the whole commit.

azayati (commented):
## AI review — Round #5 (follow-up)

Independent close-out pass by a **fresh reviewer** at head `450c1dd4` (re-checked with `gh pr view` before this pass and immediately before posting: unchanged). This reviewer received only the F1–F19 ledger, not the earlier rounds' reasoning. Fetched `origin/develop`, `origin/feature/security-fix` and the PR head; compared the 8 PR commits with the 7 FB commits by title and `git patch-id --stable`; read the whole diff against `origin/develop` (8 files, +588/−58); checked the engine behaviour on `jjwt-impl 0.13.0` (the version `component/oauth-auth` resolves, confirmed by `dependency:tree`) and `URLEncoder` on JDK 21.0.3; read the PR body and its edit history, the author's comments, the reviews and the SonarCloud gate. `domains/social.md` §11 states `component/oauth-auth` is out of the doc's scope, so no layer-2 section applies; the norms used are `backend-spring.md` §5/§7, `dev-lifecycle.md` §3b step 4 and `ai-review-and-merge.md` §2/§3/§6.

**Equivalence.** Five picks have the same patch-id as their FB commit: `707d2604` (EXO-89168), `dec1dea8` (EXO-89197), `ae7ad36a` (EXO-89198), `3f8bbaa7` (EXO-89202), `b398d2a5` (EXO-89205). `5d6fe2bb` (EXO-89200) differs from FB `829e10c9` by exactly four comment lines (`OpenIdProcessorImpl.java:651-654`), nothing else. `74ff07f4` (EXO-89222) differs from FB `b7f8f42a` by the review content the body lists (FB: 2 files, +53/−8; PR: 3 files, +327/−24, test included). `450c1dd4` is develop-only (test file, 140 lines). The body describes exactly this. No `pom.xml` is touched.

### New finding

🟢 **F20 [inherited] — a login callback with a valid `state` but no `code` now ends in a NullPointerException instead of an OAuth error.**
`component/oauth-auth/src/main/java/io/meeds/oauth/openid/OpenIdProcessorImpl.java:284-287` reads `code` with `request.getParameter(OAuthConstants.CODE_PARAMETER)` and puts it in `params` unchecked; `:308` then runs `urlEncode(entry.getValue())` on every entry, and `URLEncoder.encode((String) null, UTF_8)` throws `NullPointerException` (run on JDK 21.0.3). Before EXO-89202 the same request built `code=null`, the provider answered 400 and `:323` raised `OAuthException(ACCESS_TOKEN_ERROR)`, which `OAuthProviderFilter.doFilter` (`component/oauth-auth/src/main/java/io/meeds/oauth/web/OAuthProviderFilter.java:122`) catches and turns into the OAuth error redirect. That catch lists `OAuthException | ExecutionException | InterruptedException` only, so the NPE now reaches the container as a 500. The request is fail-closed either way, and it is only reachable with a `state` matching the session (`:265`) — from the user's own mid-flow browser or a misbehaving provider, never from a third party — hence 🟢, not the 🟡 that F13 carried for the token path. Carried by `3f8bbaa7`, byte-identical to the FB, so it is **fixed on `feature/security-fix`, not here**: in `obtainAccessToken`, refuse a missing `code` with an `OAuthException` (e.g. `ACCESS_TOKEN_ERROR`) before building `params`, and pin it next to `testObtainAccessTokenRejectsMissingStateParameter`.

Observed, not a finding (pre-existing, kept by the rewrite): `RemoteJwkSigningKeyResolver.parseKeys` (`:184`) still requires `use == "sig"` on every JWKS entry, while RFC 7517 §4.2 makes `use` optional; an entry without it now logs `Skipping unparseable JWKS entry` at ERROR on every refresh (`JSONObject.get` throws on a missing key). Same acceptance behaviour as before the PR.

### F19 verdict

Round #4's F19(a) was a miss. The PR body edit of 2026-09-18 14:17:07Z (the one posted with the F17 comment at 14:17:15Z) already contained `setWellKnownConfigurationForTest` in the `74ff07f4` list; the five earlier body versions did not. Round #4 was posted at 14:25:01Z, eight minutes after that edit. The author's claim is verified. Part (b) is a going-forward agreement (new commits instead of rewriting `74ff07f4`); nothing to verify in code.

### Consolidated ledger — F1 to F20

| Id | Sev. | Finding | Tag | Status at `450c1dd4` |
|---|---|---|---|---|
| F1 | 🟡 | EXO-89222 pick carries a test absent from the FB | backport-blocking, body only | ✅ Closed as answered — FB `b7f8f42a` ships 2 files, no test; body states the pick was extended here |
| F2 | 🟡 | Logout `state` compared then ignored | inherited | ✅ Decided (detection only) — `OpenIdProcessorImpl.java:651-659` comment + `log.warn`, `return false` |
| F3 | 🟡 | Signature alg taken from the token header | inherited | ⚠️ Fixed, in the wrong place — `RemoteJwkSigningKeyResolver.java:97` allow-list from `id_token_signing_alg_values_supported` (`:212-219`); develop-only, no ACC |
| F4 | 🟡 | One unsupported EC curve breaks every key | inherited | ⚠️ Fixed, in the wrong place — per-entry try/catch in `parseKeys` `:179-203`, pinned by `testParseKeysSkipsUnparseableEntryButKeepsTheOthers`; develop-only, no ACC |
| F5 | 🟡 | Security behaviours unpinned | inherited | ⚠️ Partly fixed — 8 unit tests (3 in `OpenIdProcessorImplTest`, 5 in `RemoteJwkSigningKeyResolverTest`); Sonar new-code coverage 44.3 % < 60 % |
| F6 | 🟢 | `equals` not constant-time | inherited | ➖ Accepted — `:265`, `:399`, `:655` still `equals` |
| F7 | 🟢 | Signed UserInfo requires `aud`/`iss` | inherited | ➖ Noted — `:501` reuses `parseAndVerifySignedJwt` (`:553-554`) |
| F8 | 🟢 | Body omitted `LoginForm.vue` | body only | ✅ Fixed — body names `LoginProviders.vue` and `LoginForm.vue` |
| F9 | 🟢 | A `kid` miss costs a fetch | inherited | ➖ Noted — `getKey` `:147` |
| F10 | 🟠 | Allow-list is the provider's capability list, not the client registration | backport-blocking | ⏭ Deferred — EXO-90395; `:212-214` unchanged |
| F11 | 🟠 | Failed discovery fetch pins the fallback for the JVM's life | backport-blocking | ✅ Fixed — `:159-161` caches only when `configuration != null`; pinned by `testUpdateKeysDoesNotCacheTheFallbackOnAFailedFetch` |
| F12 | 🟡 | No `volatile` | backport-blocking | ✅ Fixed — `:73`, `:76` volatile, double-checked under one `lock` |
| F13 | 🟡 | Refusal surfaced as 500 | backport-blocking | ✅ Fixed — `catch (JwtException)` `:384-393`; pinned by `testValidateTokenAndUpdateScopesClearsSessionOnUnverifiableSignature` (`450c1dd4`) |
| F14 | 🟡 | Body described a PR that no longer exists | backport-blocking, body only | ✅ Fixed — body matches the 8 commits and the two deviations |
| F15 | 🟢 | No `Knowledge:` line | backport-blocking, body only | ✅ Fixed — `## Knowledge` present with the §11 reason |
| F16 | 🟢 | Configurable claim names honoured in one of two places | inherited | ✅ Fixed, not pinned — `:413-415` use `OAuthConstants.*_ATTRIBUTE`; reason accepted |
| F17 | 🟢 | Sonar S3077 ×2 | backport-blocking | ✅ Closed — no S3077 among the 2 open issues (both INFO `S5786` on test classes) |
| F18 | 🟢 | F13/F16 unpinned | backport-blocking | ✅ Closed — F13 pinned; F16 waived with reason |
| F19 | 🟢 | Seam not named in body / commit rewritten | backport-blocking, body only | ✅ Closed — (a) was already in the body before Round #4 (a Round #4 miss, see verdict); (b) agreed going forward |
| F20 | 🟢 | Missing `code` on the callback ends in an NPE/500 instead of an OAuth error | inherited | 🆕 Open — to fix on `feature/security-fix`, not in this backport |

**Totals after five rounds: 20 findings — 11 fixed or closed (F1, F8, F11, F12, F13, F14, F15, F16, F17, F18, F19), 1 decided (F2), 3 fixed in the wrong place / partly (F3, F4, F5), 3 noted or accepted (F6, F7, F9), 1 deferred to a task (F10), 1 open (F20, 🟢, inherited — for the FB, nothing backport-blocking remains on this PR).**

### Classification

**N1** — the diff changes OpenID authentication itself (nonce replay check, `state` entropy and logout state, id_token and UserInfo signature verification, signing-key resolution with a new algorithm gate), and `74ff07f4` adds a gate on the token-verification path that has never run on an acceptance server.

### For the approver

- **Verified in the source and the engine:** on `jjwt-impl 0.13.0`, `alg=none` (with or without a signature) raises `MalformedJwtException`, a null resolver key raises `UnsupportedJwtException`, a key that does not match the algorithm raises `UnsupportedJwtException`; all are `JwtException`s, so they hit `OpenIdProcessorImpl.java:384`, clear the three session attributes and end as `TOKEN_VALIDATION_ERROR`. The HS* branch (`RemoteJwkSigningKeyResolver.java:103-104`) is reached only after the allow-list check and is keyed on the client secret, never on a JWKS key, so no RSA-public-key-as-HMAC confusion is possible; the constructor refuses a blank client secret (`OpenIdProcessorImpl.java:145`). Every query parameter is URL-encoded (`:240-245`, `:306-309`, `:689-691`). Nonce and state are 32 random bytes, base64url (`:540-544`), checked once and cleared. The logout return leg consumes the stored state before comparing (`:649`). `LoginProviders.vue`'s template references none of the three members moved to `methods`; `LoginForm.vue` assigns `values` in `created()` before the first render. A signed-UserInfo refusal was checked too: `OAuthException` is unchecked and `getOAuthPrincipal` already sat outside the filter's try before this PR, so the JWT path adds no new failure class there.
- **Only an ACC run can verify:** the develop-only content of `74ff07f4` against a real provider — allow-list read from the discovery document, EC key parsing, `kid` refresh on rotation — and the F20 edge. The body says this plainly; the choice (functional check before merge, or re-pick from `feature/security-fix`) is the approver's.
- **Sonar:** gate red on new-code coverage only (44.3 % vs 60 %); reliability, security, maintainability ratings A, 0 blocker/critical, 100 % hotspots reviewed; 2 INFO `S5786` on the test classes. Advisory per `ai-review-and-merge.md` §5 — waiving it is that human's call.
- **Author ≠ approver:** author `rdenarie`; the only human review so far is dismissed, so no approval stands. The approver must be a second human who knows this is N1.
- **Knowledge line:** present (`None — component/oauth-auth is out of the corpus's documented scope, domains/social.md §11`), and the §11 statement was confirmed.

rdenarie:
**F20** — Confirmed (verified the `URLEncoder.encode(null, UTF_8)` NPE on JDK 21, and that `OAuthProviderFilter.doFilter`'s catch clause doesn't cover it). Inherited from `feature/security-fix` via the byte-identical `3f8bbaa7` (EXO-89202), not introduced by this backport. Not fixed here: tracked as [eXo Maintenance Backlog #90406](https://community.exoplatform.com/portal/dw/tasks/taskDetail/90406), for the FB owners.

azayati (commented):
## AI review — Round #6 (final) — close-out

Close-out pass by the Round #5 reviewer at head `450c1dd4` (re-checked with `gh pr view` at the start of this pass and immediately before posting: unchanged, no new commit). Since Round #5 was posted (14:59:25Z): one author comment answering F20 (15:04:47Z), a review re-request (15:08:27Z), no body edit (last one 14:17:07Z), no inline thread, no new review. Re-checked in this pass: the F1–F20 statuses against the code at head, the parts Round #5 covered least — `OAuthUtils` (not in the diff, but it consumes the configurable claim names), `OAuthProviderFilter`'s error path and where the stored exception goes, `OAuthLogoutFilter` as the consumer of `processLogout`'s return value, the UserInfo path `OpenIdProcessorImpl.java:470-526`, the id_token custom-claims filter `:409-417`, `LoginProviders.vue` and `LoginForm.vue` in full — and the SonarCloud gate now.

**Equivalence.** Unchanged from Round #5 (same head): five picks patch-identical to their `feature/security-fix` commit (`707d2604`, `dec1dea8`, `ae7ad36a`, `3f8bbaa7`, `b398d2a5`); `5d6fe2bb` (EXO-89200) differs from FB `829e10c9` by four comment lines only (`OpenIdProcessorImpl.java:651-654`); `74ff07f4` (EXO-89222) carries the review content the body lists (FB `b7f8f42a`: 2 files, +53/−8; PR: 3 files, +327/−24); `450c1dd4` is develop-only (tests). No `pom.xml` touched.

### New findings

None. What the sweep verified, so the approver knows what was covered:

- **Claim names, both consumers agree.** `OAuthUtils.convertOpenIdInfoToOAuthPrincipal` (`component/oauth-auth/src/main/java/io/meeds/oauth/utils/OAuthUtils.java:126-140`) reads `EMAIL_ATTRIBUTE`, `GIVEN_NAME_ATTRIBUTE`, `FAMILY_NAME_ATTRIBUTE` from UserInfo first, then from the id_token custom claims; the id_token filter (`OpenIdProcessorImpl.java:413-415`) keeps exactly those three under the same constants, so a claim renamed through `exo.oauth.openid.scope.firstname/lastname` is captured and read with one name (F16's fix and its consumer match). `OAuthUtils` is untouched by the PR.
- **Error path of the filter.** `OAuthProviderFilter.doFilter` (`.../web/OAuthProviderFilter.java:122-128`) logs the `OAuthException`, stores it in the session as `_oauthException` and redirects to the initial URI or the context path (`redirectAfterOAuthError`, `:199-209`). The only other reader of that attribute is `OAuthLoginServletFilter.java:108`, which removes it; no code maps an `OAuthExceptionCode` to the login page's `error` parameter, and no i18n label exists for any code (grep over `*.properties`, `*.js`, `*.vue`, `*.jsp`, `*.gtmpl`). So the new `INVALID_NONCE` (`OpenIdProcessorImpl.java:403`) is surfaced exactly like the pre-existing `INVALID_STATE` (`:270`): no gap specific to this PR.
- **Logout return leg completes the local logout.** `OAuthLogoutFilter.doFilter` (`.../filter/OAuthLogoutFilter.java:51-64`) calls `processLogout` for each enabled provider and only returns early on `true`; the return leg (`OpenIdProcessorImpl.java:647-660`) returns `false`, so `filterChain.doFilter` runs and the local session ends. Any exception inside `processLogout` is caught at `:60` and the local logout still happens.
- **UserInfo path.** A non-200 answer returns an empty `JSONObject` (`:489-494`) and the principal falls back to the id_token claims (`OAuthUtils.java:126-128`); the JSON-vs-JWT branch at `:499-501` hands a JWT to the same `parseAndVerifySignedJwt` as the id_token (F7). The `IO_ERROR` translation in `OpenIdRequest.executeRequest` is unchanged, and `OAuthException` is unchecked, so this path's error handling is the same before and after the PR.
- **Vue files.** `LoginProviders.vue`'s template (`:26-52`) binds only `providers`, `mainProviders`, `menuProviders`, `displayMoreMenu`, `mainProvidersWidth`; the three members moved to `methods` are called only from `refreshProviders()` (`:108`), so no template reads a method as a property. `LoginForm.vue` assigns `values` in `created()` (`:258`) before the first render; the template reads it at `:49`, `:82`, `:208-210`. Observed, not a finding (pre-existing, outside the diff): `LoginForm.vue:289` assigns `this.providers`, which is not declared in `data()` — the same shape the PR fixed for `values`; it renders today only because the `init` flip at `:293` re-runs the render. One for the FB owners, not for this backport.
- **Sonar now:** unchanged — gate red on new-code coverage only (44.3 % vs 60 %); 0 bugs, 0 vulnerabilities, 100 % hotspots reviewed; 2 INFO `java:S5786` on the two test classes. Advisory.

### Consolidated ledger — F1 to F20

| Id | Sev. | Finding | Tag | Status at `450c1dd4` |
|---|---|---|---|---|
| F1 | 🟡 | EXO-89222 pick carries a test absent from the FB | backport-blocking, body only | ✅ Closed as answered — FB `b7f8f42a` has no test; body states the pick was extended here |
| F2 | 🟡 | Logout `state` compared then ignored | inherited | ➖ Accepted (detection only) — `OpenIdProcessorImpl.java:651-659` |
| F3 | 🟡 | Signature alg taken from the token header | backport-blocking | ✅ Fixed here — `RemoteJwkSigningKeyResolver.java:97`, allow-list `:211-219`; develop-only, no ACC |
| F4 | 🟡 | One unsupported EC curve breaks every key | backport-blocking | ✅ Fixed here — `parseKeys` `:179-203`, pinned; develop-only, no ACC |
| F5 | 🟡 | Security behaviours unpinned | backport-blocking | ➖ Partly fixed — 8 unit tests; Sonar new-code coverage 44.3 % < 60 %, unchanged |
| F6 | 🟢 | `equals` not constant-time | inherited | ➖ Accepted — `:265`, `:399`, `:655` |
| F7 | 🟢 | Signed UserInfo requires `aud`/`iss` | inherited | ➖ Noted — `:501` → `:553-554` |
| F8 | 🟢 | Body omitted `LoginForm.vue` | backport-blocking, body only | ✅ Fixed — body names both Vue files |
| F9 | 🟢 | A `kid` miss costs a fetch | inherited | ➖ Noted — `getKey` `:147` |
| F10 | 🟠 | Allow-list is the provider's capability list, not the client registration | inherited | ⏭ Deferred — EXO-90395; `:212-214` unchanged |
| F11 | 🟠 | Failed discovery fetch pins the fallback for the JVM's life | backport-blocking | ✅ Fixed — `:159-161`, pinned |
| F12 | 🟡 | No `volatile` | backport-blocking | ✅ Fixed — `:73`, `:76` |
| F13 | 🟡 | Refusal surfaced as 500 | backport-blocking | ✅ Fixed — `catch (JwtException)` `:384-393`, pinned in `450c1dd4` |
| F14 | 🟡 | Body described a PR that no longer exists | backport-blocking, body only | ✅ Fixed — body matches the 8 commits and the two deviations |
| F15 | 🟢 | No `Knowledge:` line | backport-blocking, body only | ✅ Fixed — `## Knowledge` present |
| F16 | 🟢 | Configurable claim names honoured in one of two places | inherited | ✅ Fixed, not pinned — `:413-415`; consumer `OAuthUtils.java:126-140` agrees; reason for no pin accepted |
| F17 | 🟢 | Sonar S3077 ×2 | backport-blocking | ✅ Closed — none open; 2 INFO S5786 remain (test classes) |
| F18 | 🟢 | F13/F16 unpinned | backport-blocking | ✅ Closed — F13 pinned; F16 waived with reason |
| F19 | 🟢 | Seam not named in body / commit rewritten | backport-blocking, body only | ✅ Closed — (a) was in the body before Round #4 (Round #4's miss); (b) agreed going forward |
| F20 | 🟢 | Missing `code` on the callback ends in an NPE/500 instead of an OAuth error | inherited | ➖ Accepted, tracked — author answered 15:04:47Z: NPE and uncovered catch confirmed, not fixed here, tracked as eXo Maintenance Backlog #90406 for the FB owners; code at `:284-308` unchanged, as expected |

### Classification

**N1** — the diff changes OpenID authentication itself (nonce replay check, `state` entropy and logout state, id_token and UserInfo signature verification, signing-key resolution with a new algorithm gate), and `74ff07f4` adds a gate on the token-verification path that has never run on an acceptance server.

### For the approver

- **Verified:** mechanical equivalence with `feature/security-fix` for five picks; the two deviations and the develop-only test commit are exactly what the body describes. Engine facts on `jjwt-impl 0.13.0` (the version the module resolves): `alg=none`, a null resolver key and a key that does not match the algorithm all raise `JwtException`s that `OpenIdProcessorImpl.java:384` translates into `TOKEN_VALIDATION_ERROR` with the session cleared. The HS* branch is keyed on the client secret only after the allow-list check; no JWKS key can be used as an HMAC secret. Every query parameter is URL-encoded. Nonce and state are 32 random bytes, checked once, cleared on success and on failure. The logout return leg ends the local session. Claim-name constants are read the same way by the id_token filter and `OAuthUtils`. Both Vue changes are safe for the templates that use them.
- **Only an ACC run can verify:** the develop-only content of `74ff07f4` against a real provider — allow-list read from `id_token_signing_alg_values_supported`, EC key parsing, `kid` refresh on key rotation. The body says so plainly.
- **Sonar:** red on new-code coverage only (44.3 % vs 60 %), ratings A, no open bug/vulnerability/hotspot, two INFO test nits. Advisory per `ai-review-and-merge.md` §5; waiving it is the approver's call (F5).
- **Author ≠ approver:** author `rdenarie`; the only human review so far is dismissed, so no approval stands. The approver must be a second human who knows this is N1.
- **Knowledge line:** present (`None — component/oauth-auth is out of the corpus's documented scope, domains/social.md §11`); the §11 statement was confirmed.
- **Left open, none blocking this PR:** F20 (inherited, Backlog #90406) and F10 (EXO-90395) are the FB owners'; F5's coverage gate is a human decision.

**CLEAN — nothing to fix in this backport.** Remaining for a human: the N1 approval by a second person, and whether the advisory coverage gate is waived.

azayati (approved):

PR #6107, opened 2026-09-17
feat: Allow to list pages in search whatever their content - EXO-89723

mkrout:
Prior to this, a page without a content block never appeared in the unified search, so a user looking for "Spaces" by name found nothing.     
This is caused by the page index holding one document per content block and nothing for the page itself.                                           This commit fixes the problem by indexing every reachable page that carries no content block as its own document, by ranking name matches ahead of content matches in the search query, and by moving the index to page_content_v2 so existing installations reindex on upgrade.


component/core/src/main/java/io/meeds/social/cms/listener/PageSavedIndexingListener.java
```diff
@@ -102,7 +102,16 @@ public void onEvent(Event<Object, Page> event) throws Exception {
...
+    }
+    if (LayoutService.PAGE_CREATED.equals(event.getEventName())) {
+      // A creation: the connector also queues the content blocks the new
+      // page may already carry, which nothing else would ever index
+      indexingService.index(PageContentIndexingConnector.TYPE, storageId);
+    } else {
+      // A refresh: the page's content blocks, if any, are refreshed by the
+      // Layout event listener, not re-queued from here
       indexingService.reindex(PageContentIndexingConnector.TYPE, storageId);
     }
```
ahamdi, line 116: 🟡 **F2 — The new dispatch rule drops a case the previous iteration covered and the Javadoc still promises: a page with live blocks updated outside the Layout addon**
[`PageContentIndexingConnector.java:432-437`](https://github.com/Meeds-io/social/blob/9b2bfb34ba4a5660875b18d0e7fb163a34693f8b/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentIndexingConnector.java#L432-L437) · [`PageSavedIndexingListener.java:108-116`](https://github.com/Meeds-io/social/blob/9b2bfb34ba4a5660875b18d0e7fb163a34693f8b/component/core/src/main/java/io/meeds/social/cms/listener/PageSavedIndexingListener.java#L108-L116)
```java
   * and a page stored outside the Layout services broadcasts nothing for its
   * blocks. Queuing them here makes both paths converge on the right
   * documents one indexing cycle later.
   * <p>
   * Only a creation ({@link #create}) gets here, never a refresh
```
```java
    if (LayoutService.PAGE_CREATED.equals(event.getEventName())) {
      indexingService.index(PageContentIndexingConnector.TYPE, storageId);
    } else {
      indexingService.reindex(PageContentIndexingConnector.TYPE, storageId);   // update(): a page carrying live blocks queues nothing
    }
```
Before this PR, both events queued a `create`, and a bare-page request for a page carrying live blocks queued the blocks the index lacked. Now an *existing* page saved through `LayoutService.save(page)` alone — the portal's own paths: a `PageImporter` re-import by a `LayoutUpgradePlugin`, a site template re-application, any addon saving a page without the Layout addon — broadcasts `PAGE_UPDATED` only, which is now an `update` that returns `null` for such a page and queues nothing; nothing else fires for its blocks (`layout.page.updated` is the Layout addon's event). Its block documents are therefore never (re)built until a full reindex, and the stale bare page document, if one existed, is never removed on that path either (`getUpdateDocumentRequestContent` sends nothing for a `null`). The Javadoc's own second sentence names this path and still claims convergence. The reason given for dropping the lookup is real — verified: `ElasticIndexingClient` issues `_delete_by_query?conflicts=proceed&wait_for_completion=true` with no `refresh`, so a search right after a reindex-all does see the deleted documents — but the lookup was wrong only in that one window, while the new rule is wrong on every non-Layout update.
**Fix (a decision, two shapes to weigh):** keep the operation-kind rule and make the reindex-all delete visible before the `CREATE`s run (`_delete_by_query?refresh=true` in `commons`' `ElasticIndexingClient` — hypothesis to verify against ES 9: the parameter is documented for that API), so the `e5481b0e` lookup can be restored for `update`; or accept the gap and say so — delete the "stored outside the Layout services" sentence, and name in the Javadoc the mechanism that covers such pages (if it is "the next full reindex", write that). Either way pin the case: a test where `update(storageId)` meets a page with live blocks and asserts what the chosen design promises.

webapp/src/main/webapp/WEB-INF/conf/social-extension/social/search/page-search-query.json
```diff
@@ -10,10 +10,12 @@
...
             "filter": {
-              "query_string": {
+              "multi_match": {
+                "query": "@term@",
                 "fields": ["pageTitle", "pageName"],
-                "default_operator": "AND",
-                "query": "@term@"
+                "operator": "and",
+                "fuzziness": 1,
+                "prefix_length": 1
```
ahamdi, line 18: 🟡 **F3 — A fixed `fuzziness: 1` in the top-ranked tier makes 2–3 letter terms match every near word, with no minimum term length anywhere on the path**
[`page-search-query.json:13-18`](https://github.com/Meeds-io/social/blob/9b2bfb34ba4a5660875b18d0e7fb163a34693f8b/webapp/src/main/webapp/WEB-INF/conf/social-extension/social/search/page-search-query.json#L13-L18)
```json
              "multi_match": {
                "query": "@term@",
                "fields": ["pageTitle", "pageName"],
                "operator": "and",
                "fuzziness": 1,          // one edit whatever the term length
                "prefix_length": 1
              }
```
This clause sits in a `constant_score` with `boost: 10000`, above the content tier (1000) and the relevance clause: whatever it matches is ranked first. Run against Elasticsearch 9.4.0 with the connector's own mapping and this exact template: the term `it` puts "Sign **in** page", "What **is** this", "Employee **ID** cards" and "Company and **its** history" in the top tier, all scored 10000; `dev` puts "**Des** Moines office", "**Deb** packages" and "**DV** lottery" there; `it dev` matches "Its Deb notes" through `operator: and`. Nothing on the path refuses a short term: `SearchToolbar.vue` emits after a typing pause with no length check, `SearchResults.vue` guards a blank term only, `PageSearchRest` and `PageContentSearchConnector.search` refuse a blank one only. The test asserts the literal `1` ("the name tier tolerates one typo"), so it pins the noise too. The reference (ES 9, *Common options → fuzziness*) states an integer is "the maximum allowed Levenshtein Edit Distance" regardless of length and that `AUTO` "should generally be the preferred value".
**Fix (verified on the same run):** `"fuzziness": "AUTO"` removes the 1–2 letter noise (`it` then matches only an exact `it`), but a 3-letter term still gets one edit under `AUTO` (`dev` still finds Des/Deb/DV). If 3-letter names must match exactly, the threshold form `"fuzziness": "AUTO:4,7"` does it (`dev` → exact only, `devs` → one edit); alternatively `AUTO` plus a minimum term length on the frontend or the REST layer. Same value in both copies of the template and in the test's assertion; keep `prefix_length: 1`.

ahamdi (commented):
## AI review — Round #1

Single PR, first round, one commit on `feature/maintenance` (base tip `e5481b0e`, no rebase needed). The domain doc for `social` holds nothing on the page-content index (`domains/social.md` §10 lists the connectors; `layout.md` §10 says pages are "likely indexed elsewhere") — a close-out item for the final round.

**What the diff actually does** (the body says something else, F1): (1) the bare-page/content-block dedup introduced three days ago by `e5481b0e` (one lookup of the page's indexed documents) is replaced by an operation-kind rule — a `CREATE` (`indexingService.index`, from `PAGE_CREATED`, a site-template instantiation or a full reindex) queues the page's live blocks, an `UPDATE` (`reindex`, from `PAGE_UPDATED`) never does; (2) the name tier of the query becomes a fuzzy `multi_match`; (3) a page of the global site is presented as an entry of the portal site the user searches from, through a new `site` REST parameter fed by a new `{portalName}` placeholder of the search connector URI.

F2 and F3 are inline on their lines; F1 and F4 below.

---

🟡 **F1 — The PR body and the commit message describe a commit already merged on `feature/maintenance`, not this diff**
[PR body](https://github.com/Meeds-io/social/pull/6107) · commit `9b2bfb34`
The title and the whole message are the verbatim text of `b42f8e49` (*feat: Allow to list pages in search whatever their content — EXO-89723*, merged on `feature/maintenance`): "indexing every reachable page that carries no content block as its own document… moving the index to page_content_v2". None of that is in this diff — the index alias is untouched, the bare page document predates the PR. None of the three things the diff does (above) is stated; the reversal of `e5481b0e` (same task, `fix: Skip re-queuing content blocks the index already holds`) is not mentioned either, while it is the decision the review has to weigh (F2). The `feat:` prefix repeats the merged feature's; this is its third fix iteration.
**Fix:** a body and a commit message that name the three changes and why the `e5481b0e` lookup is dropped (the reindex-all `_delete_by_query` is not refreshed — that reason is only in a Javadoc today); `fix:` prefix per `dev-lifecycle.md` §4.

---

🟢 **F4 — a few small ones**
- **Per-hit lookups in `isPresentedInSite`** [`PageContentSearchConnector.java:643-652`](https://github.com/Meeds-io/social/blob/9b2bfb34ba4a5660875b18d0e7fb163a34693f8b/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentSearchConnector.java#L643-L652): `portalConfigService.getGlobalPortal()` and `layoutService.getPortalConfig(new SiteKey(PORTAL, site))` are evaluated once per global-site hit; both depend only on the request. Resolve them once in `buildResults` and pass a boolean down — bounded by `limit` today, needless all the same.
- **Inside a space, the page keeps its "global" label and technical URL**: `eXo.env.portal.portalName` is the *current site's* name, a group name in a space (`PortalRequestContext.getCurrentPortalSite` falls back to the meta portal only server-side), so `isPresentedInSite` is false there by construction (the test `shouldKeepAGlobalSitePageAsIsWhenSearchingFromAGroupSite` pins it). The Javadoc calls that deliberate; the body calls the "Global" label something "the user never sees". If the PO wants the meta-portal presentation inside spaces too, the frontend has to send the meta portal name instead — a product question, recorded here so it is asked.
- **`/portal/` hard-coded twice** [`:663-666`](https://github.com/Meeds-io/social/blob/9b2bfb34ba4a5660875b18d0e7fb163a34693f8b/component/core/src/main/java/io/meeds/social/cms/storage/elasticsearch/PageContentSearchConnector.java#L663-L666): `relocatePath` assumes the prefix `LayoutPageUrlResolver` (in `layout`) also hard-codes; consistent, and both ignore `PortalContainer.getCurrentPortalContainerName()`. Pre-existing shape, noted.
- **Two copies of the query template** (`component/core/src/test/resources/page-search-query.json`, `webapp/.../search/page-search-query.json`): identical at head, edited in lockstep here, but the test exercises the copy, not the shipped file — a drift nobody sees. Pre-existing; a build-time copy or a single classpath location would close it.

---

**Verified conform.** The indexing-queue semantics the new rule rests on hold in `commons`: `index` → `CREATE` → `connector.create`, `reindex` → `UPDATE` → `connector.update`, and the `UPDATE` is sent as an ES bulk `index` operation (`ElasticContentRequestBuilder.getUpdateDocumentRequestContent`), so a `reindex` of a bare page document the index does not hold yet still creates it — the `PAGE_UPDATED` → `reindex` choice is safe for the page document itself, and `PageContentBlockIndexingListener` already relied on it. The premise of `isPresentedInSite` holds: `UserPortalImpl.loadUserNavigation` appends the global site's navigation to `PORTAL` sites other than the global one and to no group site, exactly the three conditions coded. `LayoutPageUrlResolver` renders `"/portal" + <router path>` and the router path of a portal-site node is `/<siteName>/<path>`, so `relocatePath`'s prefix matches the indexed shape and a path of another shape is returned verbatim. `escape()` replaces every `query_string` reserved character by a space before splicing, so `multi_match` receives plain words and the JSON stays well-formed. REST: `site` is a passthrough, `@Secured("users")`, no logic; the connector depending on `UserPortalConfigService`/`LayoutService` is a Storage-layer bean using other domains' Services, which the norm allows. The tests updated for the new dispatch verify the *absence* of the other operation (`never()`), which is the assertion that would catch a swapped kind. The four new search tests cover the three presentation branches and the non-global portal page.

**What the PR does well.** The Javadoc carries the reasoning end to end — why the lookup was abandoned (the unrefreshed `_delete_by_query`), why the global site's page is the workplace's entry, why a group site is excluded — and the tests read as the same argument in code. The `site` parameter is optional and the two-argument `search` overload keeps every existing caller intact.

**Classification: N2** — a shared component (unified search) and a list endpoint, no ACL, schema or trust-boundary change (`site` only relabels and relocates a result the query already returned). Its approver must be an Architect/Developer who knows it is N2; not to be auto-merged on AI review alone. F2 is the one item that needs a decision before merge.

mkrout:
Thanks for the round. F2, F3 and F4 are addressed in `ae858844`; F1 (PR title and body describing this diff, `fix:` prefix) is handled on the PR itself.

**F2 — a page with live blocks updated outside the Layout addon.** Taken, with a third shape: the lookup is wrong only in the reindex-all window, and that window only ever meets a `CREATE` (the reindex-all queues creations; `update()` is what `PAGE_UPDATED` sends on a live index nothing is deleting from). So `create()` keeps queuing every live block without asking the index, and `update()` gets the `e5481b0e` lookup back: it queues the blocks the index lacks, and skips the ones it holds, which the Layout listener is refreshing on a Layout save. No change in `commons` needed, no `refresh=true` on the delete. A page re-imported by `PageImporter` or a re-applied site template therefore gets its missing blocks indexed one cycle later. What a refresh still does not cover is now stated in the Javadoc of `queueLiveBlockDocuments` (a title/permission-only change outside Layout leaves indexed block documents as they were; a first block added outside Layout leaves the bare document until the next Layout save or full reindex). Pinned three ways: a creation never calls `findIndexedDocumentIds` (the mutant that does queues nothing after an upgrade), a refresh queues only the missing blocks, a refresh of a page whose blocks are all missing queues them all.

**F3 — `fuzziness: 1`.** `"AUTO:4,7"` in the shipped template: no edit under 4 letters, one from 4 to 6, two from 7. Run against ES 9.4.5 on a temporary index with your titles: `it` now returns "IT support" only, `dev` nothing, `setting` → "Settings", `spaecs` → "All Spaces", `developement` → "Development". Plain `AUTO` would have kept `dev` → Des/Deb/DV in the top tier; 3-letter names in a workplace (FAQ, CRM, SSO) are typed exactly. The test asserts the string and `prefix_length: 1`.

**F4.** The two request-dependent lookups are resolved once per search (`globalSitePresentedAs`) and a boolean check runs per hit; pinned with `times(1)` on `getGlobalPortal()`. The two copies of the template are down to one: `component/core/pom.xml` lists the webapp's `search` folder as a `testResource` including only `page-search-query.json`, and the test copy is deleted — the F3 mutant (integer fuzziness in the shipped file) failing the test is the proof it reads the shipped one. The group-site presentation is a product question, recorded for the PO in the PR body. `/portal/` hard-coded twice: left as is, same shape as `LayoutPageUrlResolver`.

Each pin was mutation-verified in one run with all four mutants planted: exactly the five pins failed, the other 93 tests stayed green; restored, 98 green.

ahamdi (approved):

PR #6098, opened 2026-09-10
fix: News keep marked as unread in the feed - EXO-89999

akhanfir:
Prior to this change, marking an activity as read from the feed (click or
timed reading) never cleared its unread badge until a page reload: the read
echoed back through the `/SpaceWebNotification` WebSocket carried the item
under the key `spacewebnotificationitemupdate`, derived from the runtime class
of the item parsed from the browser message (`SpaceWebNotificationItemUpdate`),
a key no frontend component reads. For a news post, the unread item lives on
the `news` object the activity redirects its metadata to, and the asynchronous
listener that deleted it broadcasted a second read event: the sidebar unread
counter was decremented twice and its refetch raced the deletion.

This change:
- sends the outbound WebSocket item under the stable key
  `spacewebnotificationitem` whatever the runtime class of the item
  (`SpaceWebNotificationWebSocketService.OUTBOUND_ITEM_KEY`);
- resolves the content redirection (activity -> news object) inside
  `SpaceWebNotificationServiceImpl.markAsRead`, deleting both unread items
  before the single `notification.read.item` broadcast, and removes the
  asynchronous `ActivitySpaceWebNotificationListener` that did it after a
  second broadcast. `ActivityManager` is resolved lazily from the container:
  `RDBMSActivityStorageImpl` already takes `SpaceWebNotificationService` by
  constructor (Kernel cycle otherwise).

Tests: `SpaceWebNotificationWebSocketServiceTest` (new, real JSON framework
run), `SpaceWebNotificationServiceTest` (+3: redirected item deleted before
the single broadcast with `InOrder`, plain activity not redirected,
non-activity item never resolves an activity). Each regression pin is
mutation-verified (pre-fix behaviour fails its test).

Note for the analytics owner: a news read is now recorded once, as
`markAsRead` with `entityType=activity` / `entityId=<activityId>` (previously
two rows, the second with `entityType=news`). The activity view counter is
unaffected.

Classification: N2 (raise-only from the closing review round) — one
`notification.read.item` per user read instead of two for a news post, an
event consumed by the analytics addon; no REST/DAO/schema/ACL/inbound-WebSocket
hunk; public Service API untouched. Reviewed through 2 iterative rounds and a
closing round with a fresh reviewer (CLEAN).

Knowledge: pending — eng-standards fragment drafted (`domains/social.md` §14:
`/SpaceWebNotification` wire contract, unread items keyed on the metadata
object, delete-before-single-broadcast rule); corpus PR to follow.

[...]

PR #6108, opened 2026-09-18
fix: Unread badge never appears on an activity already displayed - EXO-89999

hakermi:
Before this change, an activity already rendered in the feed never showed its unread badge when it received a new comment or like: nothing appeared until the page was reloaded, while the space badge in the navigation drawer did light up. And once a card had been read, a badge lit again later could no longer be cleared by clicking it.

This is caused by three things on the same path. UnreadBadge emits an "unread" event on notification.unread.item, but ActivityStreamActivity bound only "read", so nothing ever set unreadMetadata on a mounted card. For a content backed activity the unread item lives on the object the activity redirects its metadata to, so applicationName/applicationItemId name that content and not the activity the badge is mounted on, and the match could not have succeeded even had it been bound. And the reading timer sets isMarkedAsRead on every visible card, which markAsRead early-returns on.

This change binds the unread event; lets a consumer name through alternateIdField the item field tying a redirected item back to the object the badge is mounted on; and derives isMarkedAsRead from isUnread so the guard follows the badge actually displayed. ActivityStreamList mirrors its read path with markActivityAsUnread so the unread count stays right.

After this change the badge appears as soon as the item is created, for news and plain activities alike, and a click or the reading timer clears it.

Deliberately not closed: a card that becomes unread live does not recompute isCollapsed nor its last comments, because that watcher runs once per component instance and the comment channel already covers the common case; and an unread arriving between a read being published and its echo is lost until the next event, which would need the badge to tell its consumer that a read is in flight - more machinery than a one-round-trip window warrants.

Tests: `ActivitySpaceWebNotificationPluginTest` (new) pins activityId and the stripped activityActionType on the news, plain, comment-on-a-news and like shapes; `SpaceWebNotificationWebSocketServiceTest` (+1) pins both fields on the wire. Both classes had to be registered in their module's suite — `component/service` and `component/notification` run only their suite classes — which also means neither the new pin nor the pin added by #6098 had ever run in the build (304 → 308 tests in `component/service`).

Mutation-verified, each pin against the production line it guards: removing both `setActivityId` calls fails the three producer tests on their own activityId assertions (`expected:<100> but was:<null>`); removing the `Plugin` suffix strip fails two of them on `ComparisonFailure` (`expected:<Like[]> but was:<Like[Plugin]>`); making `activityId` `transient` on `SpaceWebNotificationItem` — which compiles under Lombok and is silently dropped by `JsonGeneratorImpl` — fails the serialization pin by name. Each mutant was reverted and the suites re-run green.

Verified at runtime on a local 7.3.x-maintenance instance built from this branch: with the feed open and a comment posted by another user over REST, the badge appears on the card without any interaction, for a plain activity and for a news-backed one; the unread item is then cleared by the reading timer, which only re-arms if the badge actually rendered.

Classification: N2 — UnreadBadge is a globally registered shared component and the change rides a wire contract fed by content's notification plugin; no REST/DAO/schema/ACL surface. Its approver must know it is N2 rather than acking on AI review alone.

Knowledge: Meeds-io/eng-standards#77 — adds the Vue watcher-edge norm to `best-practices/frontend-vue.md` (a watcher fires on an edge observed at flush, so round-tripping a flag is no edge; verify a reactivity claim by running it), the `/SpaceWebNotification` unread-badge wire contract to `domains/social.md` §8, and widens the suite-registration pitfall in §14 to `component/service` and `component/notification`.


hakermi:
## AI review — Round #1

Reviewed at head `b716483ea6f95c0d9737cde0c5b130b97ef55281` against a local checkout at the same SHA. The three-part diagnosis in the PR body was traced end to end — producer (`ActivitySpaceWebNotificationPlugin`), transport (`SpaceWebNotificationWebSocketService`), consumer (`UnreadBadge` → `ActivityStreamActivity` → `ActivityStreamList`) — and it holds. No correctness, ACL or performance finding on the code. What follows is PR metadata and nits.

---

🟡 **Medium — the PR title breaks the task reference: `O-89999` instead of `EXO-89999`**

```
fix: Unread badge never appears on an activity already displayed - O-89999
```

The commit on the branch carries the right id (`38a357f8a6` / `b716483ea6` → `… - EXO-89999`); only the PR title lost the `EX`. `dev-lifecycle.md` §4 makes `EXO-<TRIBE_TASK_ID>` the single thing that ties a PR back to its Tribe sub-task — and this task already spans several PRs (#6098 on the same id), so the link is load-bearing, not decorative.

**Fix:** rename the PR title to `fix: Unread badge never appears on an activity already displayed - EXO-89999`.

---

🟡 **Medium — no `Knowledge:` line in the PR body, on a PR targeting `feature/maintenance`**

`feature/maintenance` is a trusted lineage, so `dev-lifecycle.md` §3b step 5 applies: the body carries `Knowledge: Meeds-io/eng-standards#NN` or `Knowledge: none — <reason>`.

There is something real to record here, which argues against `none`: `domains/social.md` §8 documents the shared frontend catalogue but holds nothing on the `UnreadBadge` contract, and that contract is exactly what this PR extends — *the item a badge reacts to may name another object, and `activityId` is the field that ties a redirected item back to the activity the badge is mounted on*. That fact spans three repos (`social` produces it, `content`'s `NewsSpaceWebNotificationPlugin` produces it too, and both the feed badge and the sidebar counter consume it) and is invisible from any one of them.

**Fix:** add a `Knowledge:` line; either the eng-standards PR adding the redirected-item/`alternateIdField` contract to `domains/social.md` §8, or `Knowledge: none — <reason>` if you judge it already covered.

---

🟢 **Nit — a few small ones**

- [`NoContainerTestSuite.java:30`](https://github.com/Meeds-io/social/blob/b716483ea6f95c0d9737cde0c5b130b97ef55281/component/service/src/test/java/org/exoplatform/social/service/test/NoContainerTestSuite.java#L30) — the new import sits in a group of its own, between the `org.exoplatform.social.rest.api.*` block and the `io.meeds.*` block. It belongs in the existing `org.exoplatform` group, right after `EntityBuilderReportsTest`.

- [`UnreadBadge.vue:86-90`](https://github.com/Meeds-io/social/blob/b716483ea6f95c0d9737cde0c5b130b97ef55281/webapp/src/main/webapp/vue-apps/common/components/UnreadBadge.vue#L86-L90) — the new watcher reopens the guard only when `isMarkedAsRead` was `true`. On a card mounted less than ~5 s earlier it is still `false` (set in `mounted()`), so the assignment is a no-op, the `isMarkedAsRead` watcher never fires, and the `displayTimeout` armed at mount keeps its original deadline: the badge appears at t≈2 s and the reading timer clears it at t≈5 s instead of giving it a full `waitTimeToMarkAsRead` window. `countText()` is not re-run either, so `textLength` stays whatever was measured while the card was still in `displayLoading`. Narrow (it needs the card visible at mount *and* an unread within the first few seconds) and benign, but if you want the window to be honest, re-arm from the `isUnread` watcher rather than relying on the `isMarkedAsRead` transition.

- [`UnreadBadge.vue:163-169`](https://github.com/Meeds-io/social/blob/b716483ea6f95c0d9737cde0c5b130b97ef55281/webapp/src/main/webapp/vue-apps/common/components/UnreadBadge.vue#L163-L169) — the primary branch compares with `===` and the alternate one with `String(...) === String(...)`. Both are correct today (ids are strings on both sides of the wire), but the two spellings in four lines invite a reader to think one side can be numeric. Pick one.

- `backend-spring.md` §7 asks that a regression pin be mutation-verified and that the PR say so. Both new classes are characterization pins over unchanged production code, which is the right call and is well argued in their Javadoc — a line in the body saying you dropped `setActivityId` / `setActivityActionType` and watched them go red would close the claim.

---

### Verified conform

- **Redirected-item matching is sound.** `ActivitySpaceWebNotificationPlugin` sets `activityId` on all three branches (plain, redirected, comment-on-parent) — [`:64-69`](https://github.com/Meeds-io/social/blob/b716483ea6f95c0d9737cde0c5b130b97ef55281/component/notification/src/main/java/org/exoplatform/social/notification/plugin/ActivitySpaceWebNotificationPlugin.java#L64-L69) — and so does `content`'s `NewsSpaceWebNotificationPlugin` (`setActivityId(activityId)`), so the alternate match covers news publication as well as comments and likes. No third producer sets `activityId`, so `alternateIdField="activityId"` cannot collide with another application's item.
- **The alternate branch cannot widen what a badge sees.** `sendMessage` pushes the frame to one recipient only (`continuationService.sendMessage(identity.getRemoteId(), …)`), the frame carries no content beyond `activityActionType`, and the `Number(this.spaceId) === Number(spaceId)` gate is untouched. No ACL surface, no cross-user leak.
- **`alternateIdField` is backwards compatible**: `default: null` makes `alternateId` falsy and `matchesNotificationItem` collapses to the previous condition, so the other consumers of the shared component are unaffected. `ActivityStreamActivity.vue` is the only `<unread-badge>` consumer inside this repo.
- **The read leg closes for a redirected item**: `SpaceWebNotificationServiceImpl.markAsRead` already deletes both the `(activity, activityId)` and the `(news, newsId)` metadata objects via `getRedirectedMetadataObject`, and echoes `(activity, activityId)`, which the badge matches on its *primary* branch — so click-to-clear and timer-to-clear both work on a news-backed card.
- **`markAsUnread`'s guards are the right ones**: `isActivityShared` correctly silences the nested badge of a shared activity (whose `objectId` is the original activity and would otherwise match), and the `unreadMetadata` guard makes a repeat unread — `mergeExistingUnreadProperties` deletes and recreates, so a second comment *does* re-broadcast — a no-op instead of a redundant state write.
- **`markActivityAsUnread` is a faithful mirror of `markActivityAsRead`**, `$set` on a possibly absent `metadatas`, and `refreshUnreadCount` re-derives from `metadatas.unread.length` on both paths.
- **Both new tests run green**, standalone: `ActivitySpaceWebNotificationPluginTest` 4/4, `SpaceWebNotificationWebSocketServiceTest` 4/4. I did not run the whole `InitContainerTestSuite` / `NoContainerTestSuite` they were added to.
- **Copyright headers**: the new file carries `2020 - 2026` in the repo's shape; no year moved on a pre-existing file. `eslint` on the three `.vue` files reports only the two pre-existing accessibility warnings on `UnreadBadge.vue:2`.

### What this PR does well

Two things stand out. First, `SpaceWebNotificationWebSocketServiceTest` was in no suite — surefire in `component/service` only runs `InitContainerTestSuite` and `NoContainerTestSuite`, so the class had never executed in CI; adding it there fixes a test that never ran, independently of this bug. Second, the three "deliberately not closed" items in the body are each a real bound, stated with the reason not to close them, and the one in `markAsUnread` is written into the code next to the guard it explains — the next reader will not mistake the guard for an oversight. The `alternateIdField` prop is the right shape too: opt-in, documented on the prop, and it leaves the shared component's existing contract untouched.

---

**Classification: N2** — the diff changes the contract of `UnreadBadge`, a catalogued shared component exported through `commonVueComponents` and reachable by every addon, and it rides a cross-domain seam (`social` badge / `content` notification item). No socle trigger is tripped: no ACL check, no schema or Liquibase change, no REST or DAO surface, no Kernel↔Spring bridge, no MCP tool, and no client-controlled input reaching a trusted context. This PR needs one human ack from an Architect/Developer who knows it is N2 — it must not be merged on the strength of this AI review alone.

hakermi:
## Round #1 — answers

All six items addressed. New head `511e47387e`.

🟡 **PR title** — fixed, the title now carries `EXO-89999`. Agreed it is load-bearing: the same id already carries #6098, so it is what ties the two PRs to one sub-task.

🟡 **`Knowledge:` line** — added to the body, and deliberately not `none`. You put your finger on exactly the fact that is missing from the corpus: *the item a badge reacts to may name another object, and `activityId` is what ties a redirected item back to the activity the badge is mounted on*. `kb search` returns nothing on this channel — the nearest hits are the App Center badge freshness contracts in `domains/matrix.md` §12 and `domains/email-connector.md` §12, which are a different mechanism. The line names the fragment owed for `domains/social.md` §8.

🟢 **Import grouping** — fixed, the import now sits in the existing `org.exoplatform` group after `EntityBuilderReportsTest`.

🟢 **The watcher no-op on a freshly mounted card** — fixed, and thank you: this was real and I had not seen it. You are right that on a card mounted less than ~5 s earlier `isMarkedAsRead` is still `false`, so the assignment changed nothing, the `isMarkedAsRead` watcher never fired, and the badge inherited the mount deadline with a `textLength` measured while the card was still in `displayLoading`. The `isUnread` watcher now restarts the window itself in that case — clears `displayTimeout`, resets `isReading`, re-runs `countText()` on the text that is now displayed, then recomputes. The guard assignment stays for the other direction.

🟢 **Two spellings for the id comparison** — fixed, the alternate branch now uses plain `===` like the primary. Both sides are strings on the wire (`applicationItemId` and `activityId` are `String` on `SpaceWebNotificationItem`, the `objectId` prop is a `String`), so the `String()` wrapper was suggesting a numeric case that does not exist.

🟢 **Mutation verification not stated** — added to the body, per pin and naming the production line each one guards:
- removing both `setActivityId` calls → the three producer tests fail on their own `activityId` assertions (`expected:<100> but was:<null>`);
- removing the `Plugin` suffix strip → two of them fail on `ComparisonFailure` (`expected:<Like[]> but was:<Like[Plugin]>`) — that mutant survived an earlier version of the pin, which is why the `activityActionType` assertions exist;
- making `activityId` `transient` on `SpaceWebNotificationItem` — compiles under Lombok, silently dropped by `JsonGeneratorImpl.getTransientFields` — → the serialization pin fails by name.

Each mutant was reverted and the suites re-run green.

Two things your *verified conform* section got right that are worth keeping visible for the human approver: the `component/service` suite omission means the pin added by #6098 had never run in CI either, so this PR fixes a test that never ran independently of the bug it is about; and the read leg needs no change because `markAsRead` already resolves the redirection server-side and echoes the `(activity, activityId)` item, which the badge matches on its primary branch.

Not addressed here, tracked rather than dropped: `content`'s `NewsSpaceWebNotificationPluginTest` does not assert `getActivityId()` and its fixture uses the plain-activity shape, so the news redirection has no coverage in that repo. It is named in the Javadoc of both pins that depend on it. Whether it ships as a second PR under this id or its own task is a call for the developer.

Tests after the change: `ActivitySpaceWebNotificationPluginTest` 4/4, `SpaceWebNotificationWebSocketServiceTest` 4/4 through the module build. `eslint` unchanged — 0 errors, the two pre-existing accessibility warnings on `UnreadBadge.vue:2`.

Classification **N2** confirmed, and the point stands: this needs one human ack from someone who knows it is N2, not a merge on the strength of an AI review.

hakermi:

Re-reviewed at head `511e47387ea39ec1f98a8284f24497cef02c1794`, scoped to `b716483ea6..511e4738` plus verification of Round #1's findings in the source. Five of six are fixed and verified. One is fixed in intent but the new code does the opposite of its own comment, so it comes back escalated — the mechanism is established against Vue's shipped scheduler source and reproduced against a real Vue instance, and the reachability turns out wider than my Round #1 nit assumed. One metadata item is substantively answered but not yet in the form the check reads.

Worth saying plainly: the nit I raised in Round #1 is what produced this defect, and your reply states the restart works — it does not. That is the shape `backend-spring.md` §7 warns about, where the fix is where the next defect lives, so this round verified it against the engine rather than against the reasoning.

| Round #1 finding | Status |
|---|---|
| 🟡 PR title `O-89999` instead of `EXO-89999` | ✅ **Fixed** — title reads `… - EXO-89999` |
| 🟡 No `Knowledge:` line | ❌ **Still open (form only)** — body now carries `Knowledge: pending — …`; the contract accepts an eng-standards PR reference or `none — <reason>`, and a required check reads that field |
| 🟢 Import grouping in `NoContainerTestSuite` | ✅ **Fixed** — the import sits in the `org.exoplatform` group |
| 🟢 Badge window truncated on a freshly mounted card | ❌ **Still open, escalated to 🟠** — the fix replaces a *shortened* window with *no reading timer at all*, and the reachability is wider than Round #1 assumed. See below |
| 🟢 Two spellings for the id comparison | ✅ **Fixed** — `alternateId === this.objectId`, and the reasoning in your reply is right: the primary branch already relies on strict equality between the same two serialized strings, so that path proves it |
| 🟢 Mutation verification not stated | ✅ **Fixed** — the body names each mutant, the production line it guards and its failure message; the `Like[Plugin]` one is the interesting one, since it says the pin was strengthened after a mutant survived |

---

🟠 **High — the new `isUnread` guard kills the reading timer instead of restarting it**

[`UnreadBadge.vue:86-99`](https://github.com/Meeds-io/social/blob/511e47387ea39ec1f98a8284f24497cef02c1794/webapp/src/main/webapp/vue-apps/common/components/UnreadBadge.vue#L86-L99)

```js
isUnread() {
  if (this.isUnread && !this.isMarkedAsRead) {
    // Lit while the timer was already running: restart its window on the
    // text that is now displayed, instead of keeping the mount deadline
    window.clearTimeout(this.displayTimeout);
    this.displayTimeout = false;
    this.isReading = false;     // queues the isReading watcher...
    this.countText();
  }
  this.isMarkedAsRead = !this.isUnread;
  this.computeIsReading();      // ...and sets it straight back to true, same tick
},
```

`displayTimeout` is armed in exactly one place — the `isReading` watcher, on a genuine `false → true` edge. Here `isReading` is set to `false` and back to `true` **inside one watcher callback**, so Vue never sees an edge. `queueWatcher` de-duplicates by id (`vue@2.7.16, dist/vue.common.dev.js:3311-3341`, `if (has[id] != null) return;`), so the `isReading` watcher is queued once and runs only after the whole `isUnread` callback has finished — by which time `run()` re-evaluates the expression and skips the callback:

```js
// vue@2.7.16, dist/vue.common.dev.js:4205-4226 — Watcher.prototype.run
var value = this.get();
if (value !== this.value || isObject(value) || this.deep) { /* cb */ }
```

`value` and `this.value` are both `true`. The pending timeout has been cleared and nothing re-arms it.

**The precondition is narrow and exact**: an unread landing while `isReading` is *already* `true` with a pending `displayTimeout` — i.e. a card in the viewport, past its 1 s geometry timeout, not yet read since mount. Everything else is fine, and I checked each: a card off-screen, an unread inside the first second (before geometry), and the case where the card had already been read all arm correctly.

**Two real entries into it**, and the second is the one that makes this more than an edge:

1. A WebSocket `notification.unread.item` on a visible card — the case the guard was written for.
2. The parent's `initialized` watcher ([`ActivityStreamActivity.vue:246-248`](https://github.com/Meeds-io/social/blob/511e47387ea39ec1f98a8284f24497cef02c1794/webapp/src/main/webapp/vue-apps/activity-stream/components/activity/ActivityStreamActivity.vue#L246-L248)) setting `unreadMetadata` from `activity.metadatas.unread` when the extension's async `init(...)` resolves later than the badge's 1 s geometry timeout. That is not hypothetical: `content`'s news activity extension returns a REST promise from `init` (`$newsServices.getNewsByActivityId(activityId, lang)`, `news-extensions/activity-stream-extensions/extensions.js:38-51`) whenever the activity carries no embedded `news` payload. So an ordinary **server-loaded unread news card** — not a live one — can land in this state on a cold fetch. I verified the call is there; I have not measured its latency.

Entry 2 is why this is 🟠 rather than a nit: before this PR there was no `isUnread` watcher at all, so such a card kept its mount deadline and the reading timer fired. Being fair to the change, though — in entry 1 the guard does **not** turn a correct auto-read into a wrong one; it replaces an *early* auto-read (the mount deadline, the thing Round #1 complained about) with *no* auto-read. And the badge is still cleared by a click, by the card leaving and re-entering the viewport, or by a tab hide/show — all three make `isReading` genuinely toggle, which re-arms it.

Reproduced against the real `vue@2.7.16` in `webapp/node_modules`, transcribing this component's reactive core with a deterministic clock:

```
guard PRESENT (511e4738): t+1s isReading=true displayTimeout=2  ->  after unread: displayTimeout=false  ->  t+60s: markAsReadCalls=0
guard ABSENT  (b716483a): t+1s isReading=true displayTimeout=2  ->  after unread: displayTimeout=2      ->  t+60s: markAsReadCalls=1
scroll out and back in  : displayTimeout=5 -> markAsReadCalls=1   (recovers)
scroll with card still visible: displayTimeout=false -> markAsReadCalls=0   (no rescue: no transition)
```

The window is `waitTimeToMarkAsRead` long — a 4 s floor, and up to ~49 s on a 1000-character activity.

**Fix (hypothesis — please verify against the component rather than taking it as an instruction).** The constraint any fix has to satisfy is the one above: only a genuine `false → true` transition *observed at flush time* arms the timer, and a false-then-true pair inside one callback is not one. Deferring the recompute one tick restores the arm in the broken leg and leaves the other cases untouched:

```js
this.isMarkedAsRead = !this.isUnread;
await this.$nextTick();
this.computeIsReading();
```

```
WITH $nextTick: A -> ARMED, B -> ARMED, scroll out/in -> ARMED
```

An `async` watcher is a wart, so the alternative worth weighing is to stop routing the re-arm through `isReading` at all: extract the arming body of the `isReading` watcher into a method and call it from the guard after `countText()`. Either way the comment should say *why* the recompute cannot be synchronous — it is not self-evident, and the next reader will otherwise undo it.

---

### Verified conform (this round)

- **`alternateId === this.objectId`** — safe. Both values come out of the same `JsonGeneratorImpl` frame as `String` fields of `SpaceWebNotificationItem`, and the primary branch has always compared `applicationItemId === this.objectId` strictly on the path that demonstrably works today for plain activities.
- **The guard's other direction is intact**: when the card goes read (`isUnread` true → false) the `if` is skipped, `isMarkedAsRead` genuinely flips to `true`, and the duplicate `computeIsReading()` is harmless. Watcher ids order the flush as `isUnread(1) → isMarkedAsRead(2) → isReading(4)`, so `countText()` has run before `waitTimeToMarkAsRead` is read — the re-armed window uses a fresh `textLength`, as intended.
- **`countText()` in the guard is the right call**: it is the one path where the `isMarkedAsRead` watcher will not fire, so without it `textLength` would keep the value measured while the card was in `displayLoading`.
- **Tests at this head**: `ActivitySpaceWebNotificationPluginTest` 4/4 and `SpaceWebNotificationWebSocketServiceTest` 4/4, run through the module build. `eslint` on the three `.vue` files: 0 errors, the two pre-existing accessibility warnings on `UnreadBadge.vue:2`. I did not run the full `InitContainerTestSuite` / `NoContainerTestSuite`.
- **The `content` coverage gap you tracked rather than dropped is the right disposition** — `NewsSpaceWebNotificationPluginTest` not asserting `getActivityId()` is that repo's debt, it is named in both pins' Javadoc, and pulling it into this PR would mean a second repo on a maintenance fix. Whether it rides this id or its own is the developer's call, not the reviewer's.

### Open items

One code item (above, 🟠) and one governance item: the `Knowledge:` line. The substance of your answer is right and the fragment you describe is the one the corpus is missing — `kb search` confirms the nearest hits are the App Center badge-freshness contracts in `domains/matrix.md` §12 and `domains/email-connector.md` §12, which are a different mechanism. What is missing is only the artifact: `Knowledge:` takes an eng-standards PR number (open is fine) or `none — <reason>`, because a required check reads that field and `pending` satisfies neither. The `/domain-doc` refresh is owed by this delivery anyway and this is the cheapest moment to write it — happy to draft the fragment for `domains/social.md` §8 so the line can point at a real PR.

---

**Classification: N2**, unchanged and confirmed by this round's diff — the change is still confined to a shared Vue component's contract and a cross-domain wire seam, with no ACL, schema, REST, DAO, Kernel↔Spring or MCP surface touched. Its approver must be an Architect/Developer who knows it is N2; it must not be merged on the strength of this AI review alone. Note the Sonar quality gate on this PR is advisory, not a blocking check, so a green badge is not that approval.

hakermi:
## Round #2 — answers

New head `5599f8e43b`.

🟠 **The guard killed the timer instead of restarting it — confirmed, and fixed.**

You are right, and the mechanism is exactly as you describe. I did not take it on the reasoning: I transcribed this component's reactive core against the real `vue@2.7.16` in `webapp/node_modules` with a deterministic clock, and reproduced it before changing anything.

```
                    armed after unread    markAsRead
before (b716483a)   true                  1        <- stale mount window, the round #1 nit
head   (511e4738)   false                 0        <- the regression
```

`this.value` on the `isReading` watcher is already `true` from the arming run at t+1s, so at flush `value !== this.value` is false and the callback is skipped — the false-then-true pair inside one callback is no edge, and the pending timeout had already been cleared. Nothing re-arms it.

I took your second option rather than the `$nextTick` one: an `async` watcher hides the reason, and routing a re-arm through an edge that may or may not be observed is what produced this in the first place. The arming body of the `isReading` watcher is now `armReadTimer()`, and the `isUnread` guard calls it directly when the card is already being read — after `countText()`, so the window matches the text now displayed rather than the one measured during `displayLoading`. The comment says why toggling `isReading` would not work, since you are right that the next reader would otherwise undo it.

Four cases, run against the exact shipped shape, not the variant I first drafted:

```
                       armed   wait            markAsRead
A visible + running    true    4000 -> 19000   1     <- was: not armed at all
B re-lit after read    true                    1
C off-screen           false                   0
D read echo cleared    false                   0
```

A is the case you found; B, C and D are unchanged from before this PR. The fix also closes the original round #1 nit, which the previous attempt did not: the re-armed window uses the fresh `textLength`.

On your entry 2 — I checked the call you name and it is there: `content`'s news extension returns `$newsServices.getNewsByActivityId(...)` from `init`, so a server-loaded unread news card really can resolve after the geometry timeout and land in this state. That is what made this worth a full verification rather than a patch. I have not measured its latency either, so I am treating the reachability as established and the frequency as unknown.

🟡 **`Knowledge:` line — fair, `pending` satisfies no side of the contract.** The substance was already agreed; what is missing is the artifact, and I would rather write it than argue the form. I will take you up on the offer in the sense that matters: the fragment for `domains/social.md` §8 is owed by this delivery anyway, and the line will point at a real eng-standards PR number before this is merged. Flagging for the human approver that this is the one item still open on the N2 checklist.

Nothing else changed in this round: the two pins, the suite registrations, `alternateIdField` and the `markAsUnread` guards are untouched at `5599f8e43b`.

Classification **N2**, unchanged — and agreed on the Sonar point: the gate is advisory here, so a green badge is not the human ack this needs.

hakermi:
## AI review — Round #3 (follow-up)

Re-reviewed at head `5599f8e43b7e28fe2c327127b450c5368375e232`, scoped to `511e4738..5599f8e4` — one file, `UnreadBadge.vue`. The 🟠 is fixed, and I verified it the same way it was found rather than against the reasoning: transcribing the component's reactive core at *this* head against the real `vue@2.7.16`, with a deterministic clock that drains Vue's microtask queue between timer callbacks.

| Finding | Status |
|---|---|
| 🟠 The `isUnread` guard killed the reading timer instead of restarting it | ✅ **Fixed** — verified independently, see below |
| 🟡 `Knowledge:` line is `pending`, which satisfies neither side of the contract | ❌ **Still open** — accepted, not waived; the body still reads `pending` at this head |
| Everything else from rounds #1 and #2 | ✅ unchanged and still correct at this head |

```
A visible + timer running : armed=true  wait=19000  (the pending timer is replaced, not cleared)
  A after 60s             : markAsRead=1            <- was 0 at 511e4738
B re-lit after read       : markAsRead=1
C off-screen -> scrolled in: markAsRead=1
D read echo               : armed=false markAsRead=0
```

A is the case that was broken and it is closed: `armReadTimer()` is called directly, so nothing depends on an edge that may or may not be observed, and the replaced window is sized on the text now displayed (4000 → 19000 in the run above). B, C and D are unchanged from before the PR. Extracting the arming body rather than deferring with `$nextTick` was the better of the two options — it removes the dependence on the scheduler rather than working around it, and the comment now names the trap instead of leaving the next reader to rediscover it.

No Java file changed in this round, so the two pins and the suite registrations stand as verified in round #2. `eslint`: 0 errors, the two pre-existing accessibility warnings on `UnreadBadge.vue:2`.

---

🟢 **Nit — `countText()` is called on one arming path out of three**

[`UnreadBadge.vue:133-140`](https://github.com/Meeds-io/social/blob/5599f8e43b7e28fe2c327127b450c5368375e232/webapp/src/main/webapp/vue-apps/common/components/UnreadBadge.vue#L133-L140)

```js
armReadTimer() {
  window.clearTimeout(this.displayTimeout);
  this.displayTimeout = window.setTimeout(() => {
    if (this.computeIsReading()) { this.markAsRead(); }
  }, this.waitTimeToMarkAsRead);   // sized on whatever textLength currently holds
},
```

`waitTimeToMarkAsRead` derives from `textLength`, and `textLength` is refreshed in only two places: the `isMarkedAsRead` false-transition, and the new `isUnread` guard. The third arming path — the `isReading` watcher's `false → true` edge — refreshes nothing, so it sizes the window on the mount-time measurement, which is `0` whenever the card was still in `displayLoading` at the time. Concretely, an unread landing *before* the 1 s geometry timeout arms a 4 s window on a card that warrants 19 s:

```
E unread before geometry, as written : markAsRead=1  wait=4000
E with countText() in armReadTimer() : markAsRead=1  wait=19000
```

Not a regression — it behaved this way before the PR too — but it is the last corner of the round #1 nit, and `countText()` already self-guards on `isMarkedAsRead`, which is necessarily `false` on every path that arms. Verified that moving it leaves A, B, C and D byte-identical.

**Fix:** make `countText()` the first line of `armReadTimer()` and drop the now-redundant call in the `isUnread` guard, so every arming path sizes on the text actually displayed.

---

### Open items

**Code: none blocking.** The nit above is optional and I would not hold a merge for it.

**Governance: the `Knowledge:` line.** Still `pending — …` at this head. Taking you at your word that the number lands before merge, so this is a sequencing item rather than a disagreement — but it is the one thing that stops this being a close-out round, and it is on the N1/N2 checklist the human approver reads. The offer stands and is concrete: I can draft the `domains/social.md` §8 fragment and its `knowledge/changes/` change file so the line points at a real PR.

**Norm capture for the corpus.** One norm was invoked across these rounds that the corpus does not hold: `frontend-vue.md` says nothing about Vue watcher edge semantics — `kb search` returns no on-topic hit, and the file contains no mention of watchers or `$nextTick`. The fact is worth a fragment of its own, because it cost this PR a round and the wrong version of it looked correct on reading: *a watcher fires on an observed edge, not on an assignment. Setting a value and setting it back inside one watcher callback is collapsed by `queueWatcher`'s de-duplication and skipped by `Watcher.run()`'s `value !== this.value` check, so it is no edge at all — anything that must happen on a state change is called directly, never triggered by round-tripping a flag.* That belongs next to the existing frontend norms, not in this PR's body.

---

**Classification: N2**, unchanged. Three rounds in, here is where it stands for the approver: the code is done from the AI-review side, every finding across the three rounds is verified fixed in the source, and the only outstanding item is the `Knowledge:` artifact. This still needs one human ack from an Architect/Developer who knows it is N2 — the Sonar gate is advisory here and is not that ack.

hakermi:
## Round #3 — answers

New head `c3f9ebf8ef`.

🟢 **`countText()` on one arming path out of three — fixed, and you are right about the corner.** Reproduced before applying, on the same harness at this head:

```
armReadTimer WITHOUT countText()   E unread pre-geometry  wait=4000
armReadTimer WITH    countText()   E unread pre-geometry  wait=19000
```

`countText()` is now the first line of `armReadTimer()`, and the call in the `isUnread` guard is gone as redundant. A, B, C and D are byte-identical either way, verified against the exact shipped shape rather than the variant I drafted:

```
A visible + running     armed=true  wait 4000 -> 19000  markAsRead=1
B re-lit after read     armed=true                      markAsRead=1
C off-screen            armed=false                     markAsRead=0
D read echo             armed=false                     markAsRead=0
E unread pre-geometry   wait=19000                      markAsRead=1
```

Agreed it was not a regression — the pre-PR code sized that path the same way — but it was the last corner of the round #1 nit and the fix is one line in the right place, so there was no reason to carry it.

🟡 **`Knowledge:` line — still `pending`, still open, and I am not going to pretend otherwise.** The blocker is not the writing, it is that an eng-standards PR is a corpus-regime change and the developer has not given me the go to open one. It is flagged to them as the last item on the N2 checklist. Your reading is the right one: a sequencing item, not a disagreement.

**On the norm capture — agreed, and I would go further than "worth a fragment".** That sentence is the one thing from these three rounds with value beyond this PR, and it is worth writing in the form that cost me the round: *the wrong version looked correct on reading, and only an engine run told them apart*. `frontend-vue.md` holds nothing on watcher semantics — no mention of watchers, `$nextTick` or the scheduler — and `kb search` returns no on-topic hit. It should also carry the second half, which is the part a reader can act on: **a Vue behaviour claim about edges, ordering or flush timing is verified by running it against the vue in `node_modules`, not by reading the component** — a ~60-line harness transcribing the reactive core with a deterministic clock was enough to find this and to prove each fix, and it is cheap enough to be the default rather than the exception.

That belongs in `frontend-vue.md` next to the existing norms, separate from the `domains/social.md` §8 fragment about the badge's wire contract. Both are owed by this delivery and both are named for whoever picks up the corpus PR.

Nothing else changed: the two pins, the suite registrations, `alternateIdField` and the `markAsUnread` guards are untouched at `c3f9ebf8ef`. `eslint` 0 errors, the same two pre-existing accessibility warnings.

Classification **N2**, unchanged — one human ack from an Architect/Developer who knows it is N2, and the Sonar gate is not it.

hakermi:
## AI review — Round #4 (final)

Re-reviewed at head `c3f9ebf8ef96d875f06ad7b24f128af3a7ee220d`, scoped to `5599f8e4..c3f9ebf8` — one file, `UnreadBadge.vue`, one hunk pair. **Every finding from the four rounds is resolved and verified in the source. Nothing is outstanding from the AI-review side on the code.** One governance artifact remains, and it is a human's action rather than a review finding — it is named at the end so the approver sees it without reading back.

| Round | Finding | Status |
|---|---|---|
| #1 | 🟡 PR title `O-89999` instead of `EXO-89999` | ✅ Fixed at `511e4738` |
| #1 | 🟡 No `Knowledge:` line | ❌ **Still open** — form only, see *Handover* |
| #1 | 🟢 Import grouping in `NoContainerTestSuite` | ✅ Fixed at `511e4738` |
| #1 | 🟢 Badge window truncated on a freshly mounted card | ✅ Fixed at `5599f8e4` (escalated to 🟠 in #2 first) |
| #1 | 🟢 Two spellings for the id comparison | ✅ Fixed at `511e4738` |
| #1 | 🟢 Mutation verification not stated | ✅ Fixed at `511e4738` |
| #2 | 🟠 The `isUnread` guard killed the reading timer instead of restarting it | ✅ Fixed at `5599f8e4` |
| #3 | 🟢 `countText()` on one arming path out of three | ✅ Fixed at `c3f9ebf8` |

The last one, verified at this head on the same harness — the component's reactive core transcribed against the real `vue@2.7.16` with a deterministic clock, not read:

```
A visible + timer running   armed=true   wait 4000 -> 19000   markAsRead=1
B re-lit after read         armed=true                        markAsRead=1
C off-screen -> scrolled in armed=true                        markAsRead=1
D read echo                 armed=false                       markAsRead=0
E unread pre-geometry       wait=19000                        markAsRead=1
```

E is the corner that closed: `countText()` as the first line of `armReadTimer()` means every arming path — the `isReading` edge, the `isUnread` guard, the `isMarkedAsRead` transition — sizes its window on the text actually displayed. A through D are byte-identical to the previous head. `eslint`: 0 errors, the two pre-existing accessibility warnings on `UnreadBadge.vue:2`. No Java changed since round #2, so the two pins and the suite registrations stand as verified there.

One thing I looked at and deliberately did not raise as a finding: `countText()` in the `isMarkedAsRead` watcher is now strictly redundant, since `textLength` feeds only `waitTimeToMarkAsRead`, which is read only by `armReadTimer()`, which measures for itself. It costs one `innerText` read on a transition that already does more than that, and removing it would trade a real (if small) safety margin for tidiness after four rounds. Leaving it is the right call.

### What this delivery ended up being worth

Worth recording for whoever reads this thread later, because it is not visible in the final diff: the defect that took the longest to find was not in the original change but in the fix to a **nit** — a two-line guard whose comment asserted the opposite of what the code did, and which read as obviously correct. Only running it against the shipped Vue told the two apart. Both the round-#2 escalation and each subsequent fix were verified that way rather than by argument, which is why rounds #3 and #4 were short.

### Norm capture — two corpus fragments this delivery owes

Both are outside this PR and belong in `eng-standards`:

1. **`best-practices/frontend-vue.md` — Vue watcher semantics, and how to verify a claim about them.** The file holds nothing on this today: no mention of watchers, `$nextTick` or the scheduler, and `kb search` returns no on-topic hit. Two halves, and the second is the actionable one:
   - *A watcher fires on an observed edge, not on an assignment.* Setting a value and setting it back inside one watcher callback is collapsed by `queueWatcher`'s de-duplication and skipped by `Watcher.run()`'s `value !== this.value` check, so it is no edge at all. Anything that must happen on a state change is **called directly**, never triggered by round-tripping a flag.
   - *A Vue behaviour claim about edges, ordering or flush timing is verified by running it against the `vue` in `node_modules`, not by reading the component.* A ~60-line harness transcribing the reactive core with a deterministic clock was enough to find this and to prove every fix — cheap enough to be the default rather than the exception, in a repo with no JS test harness.
2. **`domains/social.md` §8 — the `/SpaceWebNotification` wire contract.** `activityId` / `activityActionType` on `SpaceWebNotificationItem`, its two producers (`social`'s `ActivitySpaceWebNotificationPlugin`, `content`'s `NewsSpaceWebNotificationPlugin`), the redirected-item / `alternateIdField` matching rule on the consumer side, and the read-side redirection `markAsRead` already resolves server-side. This one spans three repos and is invisible from any single one of them, which is why nobody has written it down.

### Handover — the one open item

The `Knowledge:` line still reads `pending`, which satisfies neither side of the contract (an eng-standards PR number, open or merged, or `none — <reason>`). The substance was agreed in round #2 and both fragments above are drafted in intent; what is missing is the artifact, and opening a corpus PR is the developer's call to make and the Architects Lead's to merge — not something this review can close. **Developer**: the two fragments above are what the PR should carry, and the `Knowledge:` line then points at its number.

### Classification

**N2**, unchanged across all four rounds — the change is confined to a globally registered shared Vue component's contract and a cross-domain wire seam, with no ACL, schema, REST, DAO, Kernel↔Spring or MCP surface touched. It needs one human ack from an Architect/Developer who knows it is N2, and must not be merged on the strength of this AI review alone. Reviewers can read this comment alone and know where things stand: code resolved, `Knowledge:` line pending, N2 ack required.

hakermi:
## Round #4 — the last item is closed

🟡 **`Knowledge:` — now a real reference.** [Meeds-io/eng-standards#77](https://github.com/Meeds-io/eng-standards/pull/77) is open, and the PR body points at it instead of `pending`. It carries three things:

- **`best-practices/frontend-vue.md`, new *Reactivity* section** — the norm you asked for, written in the form that cost this PR a round: a watcher fires on an edge observed at flush, `queueWatcher` de-duplicates by id and `Watcher.run()` skips an unchanged value, so round-tripping a flag inside one callback is no edge; whatever must happen on a state change is called directly, never via `$nextTick` or an `async` watcher. Plus the half a reader can act on — **verify a reactivity claim by running it** against the vue in `node_modules`, with the microtask-drain detail without which the harness silently proves nothing. Both claims cite `vue@2.7.16` and I checked the line numbers in the installed package before writing them down.
- **`domains/social.md` §8** — the `/SpaceWebNotification` contract: the frame may name another object than the badge, `activityId` is the tie, both producers set it, `alternateIdField` is how a consumer names it. With the two facts that cost time to rediscover: the read leg resolves the redirection server-side, and the actor never receives an unread for their own action.
- **`domains/social.md` §14** — the suite-registration pitfall widened from `component/core` to `component/service` and `component/notification`, carrying your finding as its worked case: a regression pin that had never run in CI.

It ships its change fragment and a `Verified` stamp naming both repos and shas, including the open `content` debt.

That was the one item on the N2 checklist still open, so from the AI-review side this PR is complete: four rounds, every finding verified fixed in the source, code and governance both closed.

What remains is not mine and not yours: **one human ack from an Architect/Developer who knows this is N2.** The Sonar gate is advisory here and is not that ack. For whoever picks it up, the two things worth their attention rather than the diff: the change alters the contract of a globally registered shared component (one call site today, additive and opt-in), and it rides a wire seam that `content` feeds — the `content`-side coverage gap is named in both pins' Javadoc and is owed either under this id or its own task, which is the developer's call.

Note also that `security/snyk` is red on this head for quota exhaustion — *"You have used your limit of private tests"* — not a finding; the check never ran, and this diff touches no dependency manifest.

ahamdi (approved):

